PR #11, opened 2015-11-15
Proposition of enhancement of $app-2panes-toolbar

doubleface:
I spent some time reading cozy-ui's code this week-end with the hope to use it for the databrowser. Here is a proposition of enhancement just to see if i understand cozy-ui a little bit.

### DIFF
--- a/demo/app/assets/index.html
+++ b/demo/app/assets/index.html
@@ -1,15 +1,29 @@
 <!DOCTYPE html>
 <html>
 <head>
-	<meta charset="utf-8">
-	<title>Cozy Ui - Demo app</title>
-	<meta name="viewport" content="width=device-width,initial-scale=1">
-	<link rel="stylesheet" href="styles/app.css">
+    <meta charset="utf-8">
+    <title>Cozy Ui - Demo app</title>
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <link rel="stylesheet" href="styles/app.css">
 </head>
 <body>
-	<div class="main-container">
-		<h1>Demo App</h1>
-	</div>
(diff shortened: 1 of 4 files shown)


frankrousseau:
@m4dz could you have a look at this?

m4dz:
Thanks for the PR!

I thinks some things in naming elements should be updated. I realize this when reading your PR. I'll tweak them myself.

Anyway, you've got it :wink:!

src/stylus/cozy-ui/app/layouts.styl
```diff
@@ -27,6 +27,15 @@ $app-2panes-toolbar
         display  flex
         flex-direction column
 
+    .drawer
+        display flex
+        flex-direction column
+        padded .5em
```
m4dz: I prefer don't add spacers in layouts mixins, because in most cases you'll want to deal with it app-by-app, depending of your design and features.

demo/app/assets/index.html
```diff
@@ -1,15 +1,29 @@
...
+        </div>
+    </main>
+    <script src="scripts/app.js"></script>
+    <script>
+        document.querySelector(".drawer-toggle").addEventListener("click", function(){
+            var $drawer = document.querySelector(".drawer");
+            var value = $drawer.getAttribute("aria-expanded") === "true" ? "false" : "true";
+            document.querySelector(".drawer").setAttribute("aria-expanded", value);
+        });
+    </script>
```
m4dz: Nice :). I'll export it later, but in fact, I think the drawer toggler will be exported later in a dedicated webcomponent, which will be distributed in Cozy-Ui :smiley:

PR #20, opened 2016-12-13
Migrate to vanillaJS

m4dz:
This PR remove the coffee-script dependency and clean our process for the cozy v3. It:

- removes the prepublish task
- moves the sources to root
- replaces coffee-script files with vanillaJS versions
- add linting for JS files
- removes the old demo app
- updates readme accordingly


package.json
```diff
@@ -1,10 +1,7 @@
 {
   "name": "cozy-ui",
-  "version": "0.1.7",
+  "version": "3.0.0",
   "description": "Cozy apps Ui SDK",
-  "scripts": {
```
GoOz: I don't get it. You removed the prepublish script but you didn't replaced it with a build and/or a watch. How is it supposed to build the CSS then? Or maybe i missed something else somewhere else. :open_mouth:

.gitignore
```diff
@@ -1,27 +1 @@
...
-# Coverage directory used by tools like istanbul
-coverage
-
-# Dependency directory
-# Commenting this out is preferred by some people, see
-# https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git-
 node_modules
-
-# Users Environment Variables
-.lock-wscript
```
nono: @GoOz FYI, you can create a global gitignore. Cf https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

package.json
nono: @GoOz It is intented to be used as a stylus plugin. It's the project that will include cozy-ui that will build the CSS from stylus. On npm and github, cozy-ui will be just a set of stylus files.

package.json
```diff
@@ -18,12 +15,15 @@
...
   },
   "homepage": "https://github.com/cozy/cozy-ui",
+  "scripts": {
+    "test": "standard"
+  },
   "devDependencies": {
-    "coffee-script": "^1.9.3",
-    "stylus": "^0.52.0"
+    "standard": "^8.6.0",
+    "stylus": "^0.54.0"
```
nono: We should move stylus from devDependencies to dependencies, no?

.gitignore
GoOz: @nono Yeah i know, i was just wondering if it wasn't more safe to add those in the gitignore of the project. But i was really just wondering, i'm not gonna fight for this :)

package.json
GoOz: @nono OK then :)

.gitignore
nono: I don't know. @m4dz, your call

m4dz: Yup, I think we can add environnement / OS / editors specifics files in the .gitignore, it can prevent issues essaebtially from external contributors 👍

m4dz:
> It is a bit sad that we won't have a demo anymore

Yep, I hope we can re-add it later to, but this version was too old (and still use our old Brunch stack). I prefer remove it currently and do it properly again later 😄 

> Maybe, we can avoid the directory stylus/cozy-ui and put app/, components/, etc. in the root directory

As cozy-ui will contains generic (p)react components too, sandbox stylus rulesets in a `stylus` directory seems ok to me, nope?

nono:
OK. Maybe `stylus/app`, `stylus/components` then?

PR #21, opened 2016-12-16
Update UI with button style and new colors

GoOz:


stylus/base/defaults.styl
```diff
@@ -9,3 +9,13 @@ $error
     border-radius   4px
     background-color red
     color            white
+.coz-hidden
```
m4dz: Maybe we should don't use dot-names, as they'll be exported in final CSS as classes, and stuck on dollar-names, that can be sfaely extended w/o infating the output. Except if we prefer to have a `.coz-hidden` class available, obviously (but I don't think it's a good idea, or we must document it very well)

m4dz: (also, maybe an empty line before can be useful for readability 👅 )

stylus/base/palette.styl
```diff
@@ -18,9 +18,13 @@ grey-05 = #92A0B2
 grey-06 = #748192
 grey-07 = #4F5B69
 grey-08 = #32363F
+grey-09 = #d6d8da
+grey-10 = #f5f6f7
+grey-11 = #95999d
 
 // more of cozy UI
 blue      = #33A6FF
+blue2     = #297ef2
```
m4dz: Naming is hard, isn't it 😆 ?

stylus/components/button.styl
```diff
@@ -0,0 +1,30 @@
+// Buttons of any kind
+.coz-btn
+    box-sizing border-box
+    display inline-block
+    vertical-align top
+    margin 0 em(8px)
+    border 1px solid grey-09
+    height: em(43px)
```
m4dz: Syntactic purism, but we should definitely choose between colons or no-colons, and if we pefer values aligned or not. Any preference on this? (previously we said no-colons and aligned https://github.com/cozy/cozy-guidelines#css--stylus, but we can updat this guide too).

/poke @gregorylegarec, @goldoraf, @CPatchane, @nono about this one

stylus/components/button.styl
```diff
@@ -0,0 +1,30 @@
...
+    text-transform uppercase
+
+    &--regular
+        border-color blue2
+        background-color blue2
+        color white
+
+    &--upload
+        left-padded(44px)
+        background-image: embedurl("../../node_modules/cozy-ui/assets/icons/icon-upload.svg")
```
m4dz: Ok at this time for the path, I'l update the module config to add the cozy-ui assets folder in the [lookup paths for the `embedurl` method](http://stylus-lang.com/docs/functions.url.html) so referencing external file will be more convenient 👍

stylus/components/button.styl
CPatchane: I completely agree about no-colons it's much cleaner using stylus, and same thing about aligned values (even if less important to my mind).

stylus/base/defaults.styl
GoOz: Yes, i was wondering about utility classes like this one. In my head, @extend modules are for functionnal modules (like $error, $dialog, etc) and this utility class isn't a module, it's more like a tool to easily and quickly enhance you view.
With this in mind i was thinking if maybe make it an @extend module wasn't in some way overkill. But i may be too cautious 👅

stylus/base/palette.styl
GoOz: Yeah… i almost overwrote the blue color with the new one but i thought that maybe it was still in use in another app. Then i thought that i could use [Name that color](http://chir.ag/projects/name-that-color/). Really i don't know for the time being what's the most clever choice.

stylus/base/defaults.styl
m4dz: Nope, I totall agree, fine for me 👍

stylus/components/button.styl
m4dz: I know a stylus-lint tool, I'll take a look at it :)

stylus/components/button.styl
m4dz: Fixed in #23 \o/

PR #27, opened 2017-01-09
Add new elements for Cozy apps

CPatchane:
This PR:

* [x] Use better naming and form elements
* [x] Add new form elements
* [x] Add notifications elements
* [x] Add navigation + sidebar elements
* [x] Add new buttons styles


stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+        -webkit-appearance  none
+        appearance          none
+        background-image    url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='16px' height='16px'><polyline points='14.764,4.966 8.053,11.677 1.342,4.966'/></svg>")
+        background-position right 8px center
+        background-repeat   no-repeat
+        &::-ms-expand
+            display none
+        &::placeholder
+            color     grey-07
+            font-size 1.05em
```
GoOz: There's a Stylus function to deal with `em` if you want (eg: em(24px)). It would be easier than manually calculate every size.

stylus/ui-app/nav.styl
```diff
@@ -0,0 +1,53 @@
+$navigation
+    .coz-nav
+      margin     em(24px) 0
+      padding    0
+      list-style none
+    .coz-nav-item
+        position    relative
+        z-index     10
```
GoOz: Why 10? Is there a lot of stuff that have z-index between 0 and 10 or do you increment by 10 ? Just asking, but it made me think that we absolutely need guidelines and a linter, the sooner the better.

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+            cursor      pointer
+
+    [role=button],
+    button
+        // mutualize behavior between button and a
+        box-sizing        border-box
+        border            0
+        display           block
+        padding           0.9em 0.75em
+        min-height        (40/basefont)em
```
m4dz: You can use the `em(40)` function to direclty express a size in EM, relative to its context

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+        text-decoration   none
+
+        &.cancel
+            background-color grey-01
+            color            grey-08
+
+        &.danger
+            background-color rgba(255, 93, 70, 0.1)
+            border           solid 1px rgba(255, 93, 71, 0.1)
+            color            #ff5d47
```
m4dz: Maybe those three colors should be in the _base/palette.styl_ to centralize them

stylus/ui-app/notifications.styl
```diff
@@ -0,0 +1,54 @@
...
+ * =============
+ *
+ * This file contains all needs for notifications
+ */
+
+
+$notifications
+    notification-width = (546/basefont)em
+    .coz-notifier
+        z-index 100
```
GoOz: Same question than before on your z-index incrementation style.

stylus/ui-app/notifications.styl
```diff
@@ -0,0 +1,54 @@
...
+        background-color green
+        box-shadow       0 (6/basefont)em (18/basefont)em 0 rgba(50, 54, 63, 0.23)
+        box-sizing       border-box
+        padding          1em
+        color            white
+
+        //                  When only a title is displayed
+        min-height          (96/basefont)em
+        padding-left        (96/basefont)em
+        background-image    url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='48' height='48' viewBox='0 0 48 48'><path fill='#FFFFFF' fill-rule='evenodd' d='M24,48 C37.254834,48 48,37.254834 48,24 C48,10.745166 37.254834,0 24,0 C10.745166,0 0,10.745166 0,24 C0,37.254834 10.745166,48 24,48 Z M13.4142136,24.5857864 C12.633165,23.8047379 11.366835,23.8047379 10.5857864,24.5857864 C9.80473785,25.366835 9.80473785,26.633165 10.5857864,27.4142136 L17.5857864,34.4142136 C18.366835,35.1952621 19.633165,35.1952621 20.4142136,34.4142136 L35.4142136,19.4142136 C36.1952621,18.633165 36.1952621,17.366835 35.4142136,16.5857864 C34.633165,15.8047379 33.366835,15.8047379 32.5857864,16.5857864 L19,30.1715729 L13.4142136,24.5857864 Z'/></svg>")
```
GoOz: You could use a shorthand here on both background and transition properties.

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+
+        &[aria-busy=true]
+            position relative
+
+            &::after
+                position     static
+                width        (16/basefont)em
+                height       (16/basefont)em
+                margin-left  (8/basefont)em
+                margin-right (-24/basefont)em
```
m4dz: Here also, we can use the `em()` function. For spacing, there's spacers helpers that can be useful, like `left-spaced` and `right-spaced` but I still ask myself about using them intensively. @GoOz, any though on them?

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+
+$form-text
+    input[type=text]
+    input[type=password]
+        display       inline-block
+        padding       0.9em 0.75em 0.75em 0.9em
+        border-radius 4px
+        box-sizing    border-box
+        border        1px solid grey-03
+        background    #f5f5f5
```
m4dz: We should externalize colors references :)

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+
+
+$form-select
+    select
+        display       inline-block
+        padding       0.75em 0.75em 0.9em
+        border-radius 4px
+        box-sizing    border-box
+        border        1px solid grey-03
+        background    #f5f5f5
```
m4dz: Poke colors 👅

stylus/ui-app/forms.styl
```diff
@@ -5,9 +5,158 @@
...
+        color         black
+        -moz-appearance     none
+        -webkit-appearance  none
+        appearance          none
+        background-image    url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg' width='16px' height='16px'><polyline points='14.764,4.966 8.053,11.677 1.342,4.966'/></svg>")
```
m4dz: Isn't better to put the SVG in a dedicated file in the _assets_ directory, and reference it here with the `embedurl` function?

stylus/ui-app/nav.styl
```diff
@@ -0,0 +1,53 @@
...
+                content         ''
+                position        absolute
+                background      grey-09
+                height          3em
+                width           90%
+                z-index         -1
+                border-radius   0 3px 3px 0px
+        a
+            display         block
+            left-padded(50px)
```
m4dz: If you want to keep aligned values, you can use it like

```stylus
display         block
left-padded     50px
```

stylus/ui-app/notifications.styl
```diff
@@ -0,0 +1,54 @@
+/*!
```
m4dz: Minus things in this file: externalize colors, `em()` function, `embedurl`… Like in those above :)

stylus/ui-app/forms.styl
GoOz: Well, I don't know either if it's that interesting. Or maybe it misses a shorthand feature but even so… padding/margin with em() functions seems pretty explicit and useful already, I guess.

stylus/ui-app/nav.styl
GoOz: And it still recognize left-padded as a function this way? I didn't know you could do that. Interesting. :)

stylus/cozy-ui/defaults.styl
```diff
@@ -64,7 +64,7 @@ input[type=password]
 
 
 [data-input=radio]
-    @extend $checkbox
+    @extend $form-checkbox
```
CPatchane: Actually, I don't use defaults.styl for cozy-settings, I just updated this files because that was throwign an error when I updated variable names. Maybe we doesn't it anymore? /poke @m4dz

stylus/ui-app/nav.styl
CPatchane: True. I can just use 1 here indeed.

stylus/ui-app/forms.styl
m4dz: Yep, I totally agree. My primary though about them is to use a shorthand that we can later extend if we need, w/o the need to refactor all code that only use margin/padding. But maybe it's totally overkill. Whatever we choose, we need to make all codebase compliant 😉

stylus/ui-app/nav.styl
m4dz: Yup 😃 .  In fact, CSS directives in stylus are just internal functions 😉

stylus/ui-app/notifications.styl
CPatchane: The notification should on the top of all elements in the page, so I think it's why we should define here to use z-index 100 by default. What do you think?

gregorylegarec:
@CPatchane As discussed this morning, what about renaming UI _notifications_ to _alerts_ and change the code subsequently ?

stylus/ui-app/notifications.styl
GoOz: Well I agree, we should secure the position of notifications on top of anything but by setting its z-index to 100 (on the premise that we used z-index incrementation by 1), it means that we allow 100 layers of positioned elements on a page. That is a lot. 
That's just my opinion but if we have a UI with 100 positioned elements, we really messed up the CSS and/or UI, big time. I'd say over 20 positioned elements it's already too much. 
Maybe we could at least go at 50 but I still think it would be too much.

But we could use the [postcss z-index plugin](https://github.com/ben-eb/postcss-zindex) that unify and simplify z-indexes all over CSS in a project ( poke @m4dz ). I used it before, it works fine. This way you can put the z-index at 50 or 100 if you like and the plugin would parse all the CSS and give the smallest number possible and still be the highest of the CSS file.

CPatchane:
@gregorylegarec I agree, I just replaced that to use `alert` instead. 👍

m4dz:
@CPatchane, @GoOz is this one mergeable or do you need to bring more changes before?

GoOz:
@m4dz Not yet. I think @CPatchane only needs to merge its buttons.styl to the already existent button.styl and that's it.

GoOz:
Nice job :) 👍

PR #29, opened 2017-01-12
[feat] Update alerts for Cozy Files v3

gregorylegarec:
* Adds styles for an overlay
 * Remove styles for icons in alerts
 * Update margin/padding, etc.
 * Adds styles for buttons in alerts
 * Deal with CSS transition on overlay and alerts

### DIFF
--- a/stylus/ui-app/alerts.styl
+++ b/stylus/ui-app/alerts.styl
@@ -39,8 +39,8 @@ $alerts
         cursor                     pointer
 
         p
-          margin 0
-          line-height 1.5
+            margin 0
+            line-height 1.5
 
         .coz-alert-title
             font-weight bold
@@ -49,13 +49,13 @@ $alerts
         background-color pomegranate
 
     &.coz-alerter--hidden
-      .coz-alert
-        top                        -25%
-        opacity                    0
-        transition-timing-function ease-in
+        .coz-alert
+            top                        -25%
+            opacity                    0
+            transition-timing-function ease-in
 
-      .coz-overlay
-        opacity 0
+        .coz-overlay
+            opacity 0
 
     @media all and (min-width alert-width)
         .coz-alert
--- a/stylus/ui-app/alerts.styl
+++ b/stylus/ui-app/alerts.styl
@@ -35,11 +35,11 @@ $alerts
         color            white
         opacity          1
 
-        transition                 top .5s ease-out, opacity .2s ease-out
-        cursor                     pointer
+        transition       top .5s ease-out, opacity .2s ease-out
+        cursor           pointer
 
         p
-            margin 0
+            margin      0
             line-height 1.5
 
         .coz-alert-title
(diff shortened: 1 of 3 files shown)


stylus/ui-app/alerts.styl
```diff
@@ -21,23 +31,46 @@ $alerts
...
+        top                        -25%
+        opacity                    0
+        transition-timing-function ease-in
+
+      .coz-overlay
+        opacity 0
+
+    // Deprecated update existing code and remove
+    // User .coz-alerter--hidden on .coz-alerter instead
     .coz-alert--closing
```
CPatchane: I see that you replaced `.coz-alert--closing` by `coz-alerter--hidden`. @GoOz How do we handle changes in cozy-ui that should be updated in other app (here just for cozy-settings)? A changelog?

GoOz: Well we should limit changes as much as possible, but if we have to a changelog and an update on the version should be done, obviously. That being said, we're still on the early stage of Cozy-UI. I think we can skip on this for the time being, until we have a stable Cozy-UI repo. Don't you think? /poke @m4dz

stylus/ui-app/alerts.styl
```diff
@@ -21,23 +31,46 @@ $alerts
...
         .coz-alert-title
             font-weight bold
 
     .coz-alert--error
         background-color pomegranate
-        background-image embedurl('../../assets/icons/icon-warn.svg')
 
+        .coz-btn
+          color         pomegranate
+          border        none
```
GoOz: If you have a new style for a button, you should put it in `/ui-components/button.styl`
And giving the fact that a style for .coz-btn already exists, if you can't find a matching style on the already existing modifiers, just add yours there.

gregorylegarec: How do you think I should proceed ?
```css
.coz-alert .coz-btn {}
// or
.coz-alert-btn {}
```

stylus/ui-app/alerts.styl
CPatchane: @gregorylegarec I suggest for now to remove the old way and I will update cozy-settings to use the new way. We will discuss about the cozy-ui changelog later.

stylus/ui-app/alerts.styl
CPatchane: I have a personal preference for the first one which uses the same global .coz-btn class

stylus/ui-app/alerts.styl
```diff
@@ -39,41 +39,23 @@ $alerts
         cursor                     pointer
 
         p
-          margin 0
-          line-height 1.5
+            margin 0
```
GoOz: it just needs aligned values and we're good :)

gregorylegarec: Oops, my bad :3

GoOz:
Thanks 😉

PR #31, opened 2017-01-13
Better alerts

CPatchane:
* __Class changes__: I didn't noticed but the previous class `.coz-alerter--hidden` was not compatible with a way to stack alerts (the hiding system worked by 'hiding' the alerter instead of just the alert). So I changed to use `.coz-alert--hidden` and `.coz-overlay--hidden` instead of `.coz-alerter--hidden`.
It's now compatible with a generic alerts system introduced in `cozy-settings` in this [PR](https://github.com/cozy/cozy-settings/pull/2/files#diff-6ff124af85549596cf89cf814b4de476)
* Improve overlay fadeout transition by using `visibility` property
* Add buttons classes for `info` (grey) and `success` (green) alert types.

### DIFF
--- a/stylus/ui-app/alerts.styl
+++ b/stylus/ui-app/alerts.styl
@@ -19,7 +19,15 @@ $alerts
         left       0
         background white
         opacity    .6
-        transition opacity ease-out .3s
+        visibility visible
+        transition opacity .3s, visibility 0s ease-out
+
+        &.coz-overlay--hidden
+            opacity          0
+            visibility       hidden
+            transition       opacity .3s, visibility 0s ease-in
+            transition-delay 0s, .3s
+
 
     .coz-alert
         display          block
@@ -45,17 +53,20 @@ $alerts
         .coz-alert-title
             font-weight bold
 
-    .coz-alert--error
-        background-color pomegranate
-
-    &.coz-alerter--hidden
-        .coz-alert
+        &.coz-alert--hidden
             top                        -25%
             opacity                    0
             transition-timing-function ease-in
 
-        .coz-overlay
-            opacity 0
+    .coz-alert--error
+        background-color pomegranate
+
+    .coz-alert--success
+        background-color emerald
+
+    .coz-alert--info
+        background-color grey-12
+
 
     @media all and (min-width alert-width)
         .coz-alert
--- a/stylus/ui-app/alerts.styl
+++ b/stylus/ui-app/alerts.styl
@@ -25,8 +25,7 @@ $alerts
         &.coz-overlay--hidden
             opacity          0
             visibility       hidden
-            transition       opacity .3s, visibility 0s ease-in
-            transition-delay 0s, .3s
+            transition       opacity .3s 0s, visibility 0s ease-in .3s
 
 
     .coz-alert
--- a/stylus/ui-components/button.styl
+++ b/stylus/ui-components/button.styl
@@ -42,6 +42,31 @@ $button--alert-error
     &:hover
         color     darken(pomegranate, 25%)
 
+$button--alert-info
+    color               white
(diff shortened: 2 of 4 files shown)


stylus/ui-app/alerts.styl
```diff
@@ -19,7 +19,15 @@ $alerts
...
         opacity    .6
-        transition opacity ease-out .3s
+        visibility visible
+        transition opacity .3s, visibility 0s ease-out
+
+        &.coz-overlay--hidden
+            opacity          0
+            visibility       hidden
+            transition       opacity .3s, visibility 0s ease-in
+            transition-delay 0s, .3s
```
GoOz: I don't get what you're trying to do here when redefining transition-delay.

stylus/ui-components/button.styl
```diff
@@ -42,6 +42,31 @@ $button--alert-error
...
+    background-color    grey-11
+    border              none
+    font-weight         bold
+    margin              .5rem 0 0
+    padding             .5rem 1rem
+    height              auto
+    border-radius       em(3px)
+
+    &:hover
+        color     darken(white, 10%)
```
GoOz: wrong alignement i think

stylus/ui-components/button.styl
```diff
@@ -42,6 +42,31 @@ $button--alert-error
...
+    color         emerald
+    border        none
+    font-weight   bold
+    margin        .5rem 0 0
+    padding       .5rem 1rem
+    height        auto
+    border-radius em(3px)
+
+    &:hover
+        color     darken(emerald, 25%)
```
GoOz: wrong alignement too, I think

stylus/ui-app/alerts.styl
GoOz: Ok I think I got it 😆 
First timed value is duration and then you set the delay.
You can merge them by writing `transition: name duration delay, name duration delay`

CPatchane: Yeah, that's it! Thanks! 👍

stylus/ui-app/alerts.styl
```diff
@@ -45,17 +52,20 @@ $alerts
         .coz-alert-title
             font-weight bold
 
-    .coz-alert--error
-        background-color pomegranate
-
-    &.coz-alerter--hidden
```
gregorylegarec: I don't get the point of removing this rule, for me it allows us to toggle visibility on both `.coz-alert` and `.coz-overlay`at the same time, while triggering transitions. We toggle the`*--hidden` class on one element instead of two (or maybe more in the future).

stylus/ui-app/alerts.styl
```diff
@@ -19,7 +19,14 @@ $alerts
         left       0
         background white
         opacity    .6
-        transition opacity ease-out .3s
+        visibility visible
```
gregorylegarec: What the role of visible here ?

stylus/ui-app/alerts.styl
gregorylegarec: I've just read the description of the PR, I got my response. Sorry :)

stylus/ui-app/alerts.styl
CPatchane: It was quite complicated to animate the overlay here because when you just put the opacity to 0, it's still 'clickable' et avoid the user to click on other parts of the app. As it's not possible to use transition with `display:none`, the `visibility` allows to disable the 'clickable behaviour' by being compatible with transition (by using a delay of 0.3s for the hiding transition)

gregorylegarec:
Actually it is a problem to merge this PR, as Cozy-Files-v3 still use the previous publicated version.

So thanks @CPatchane, great job as always 👍

PR #34, opened 2017-01-26
Set viewport to [role=application]

m4dz:
This PR assumes that cozy apps always uses a `[role=application]` root element, and enforces it to be set to all viewport, rather than relying on the `$app-2panes` ruleset for this.


GoOz:
Great idea 👍

PR #36, opened 2017-01-26
[chore] Enable linting on travis

m4dz:
⚠️ **DO NOT MERGE YET** ⚠️ 
Wait for config & refactoring before merging.

PR #37, opened 2017-01-27
Chore/lint

GoOz:
⚠️ **TO  BE MERGED AFTER** #36 ⚠️ 

- [x] Adds configuration to match guidelines
- [x] Cleanup CSS so it matches guidelines


m4dz:
Great! Shall we explicitely declares `error: false` statuses as it's the default value in config, or doing it only when we want to trigger an error seems to be sufficient?

GoOz:
@m4dz Absolutely not, you're totally right. I was playing with the config and it was easier to use the "verbose" version of each option to do so. Now it should be cleaned as it is indeed redundant. Will do asap.

m4dz:
@GoOz any news on it?

GoOz:
@m4dz just did it. I had trouble finding some spare time to do it.
FYI there's still some warning about not respected depth limit, but ignore them, there's some heavy refactoring on the way.

m4dz:
Woohooo \o/

PR #38, opened 2017-02-01
[refactor] Moved icons

y-lohse:
Moved cozy-ui specific icons to a subfolder so they don’t get accidentally overiden by the host application.


m4dz:
Perfect 🎉

PR #42, opened 2017-02-06
[fix] remove input css outline on focus

y-lohse:
At least Chrome applies a default blue outline when focusing an input; this outlines covers the border color we define on hover, focus and error, making them all look the same. 
This PR removes the outline on focus.


GoOz:
That's the only viable reason to remove outline. 😄  
👍

PR #47, opened 2017-02-07
[feat] Added alerter component!

y-lohse:
A basic component to display notifications!

It comes in 3 variations: error, success and info.
It auto-closes after a few seconds or you can provide a button text and callback to handle the closing elsewhere.

### DIFF
--- a/stylus/ui-components/alerts.styl
+++ b/stylus/ui-components/alerts.styl
@@ -5,7 +5,6 @@
  * This file contains all needs for alerts
  */
 
-
 $alerts
     alert-width = em(480px)
     .coz-alerter
@@ -30,6 +29,7 @@ $alerts
 
     .coz-alert
         display          block
+        z-index          100
         position         absolute
         top              em(16px)
         left             0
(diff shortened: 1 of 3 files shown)


stylus/ui-components/alerts.styl
```diff
@@ -30,6 +29,7 @@ $alerts
 
     .coz-alert
         display          block
+        z-index          100
```
GoOz: What changed since the last version of alerter that explains this new z-index? Unless it's a bugfix, maybe?

y-lohse: As far as I can tell, the other z-index in the file (`coz-alerter`) has no effect. To me it's a bugfix, but I'm not sure why the other z-index was there in the first place. I'm happy to change it back if it has implications I didn't see.

GoOz: Honestly I don't know, I'm not the one who wrote it. Maybe a cleaning would be a good thing but not right now. And I have some kind of a Z-Index manager in my todo list, so we'll see that later.
So, if it's a bugfix then I'm good :)

goldoraf:
:tada:

PR #49, opened 2017-02-08
Add i18n & withState React helpers

goldoraf:
...and some very little styling additions :guitar:


react/helpers/withState.jsx
```diff
@@ -0,0 +1,40 @@
+/** @jsx h */
+import { h, Component } from 'preact'
```
y-lohse: Why is it importing from `preact` and not `react`here?

goldoraf: Just an artefact from a not so distant past :), fixed!

goldoraf:
I use `withState` in cozy-settings and my-accounts, we'll probably find other uses of it.

stylus/ui-app/forms.styl
```diff
@@ -11,9 +11,11 @@ checkbox-size = 20px
 
 $form
     .coz-form
-        position relative
         margin   1em 0 .5em
 
+    .coz-form
+    .coz-form-group
```
GoOz: What is the use case for a new class? I can't really see the need without an example :(

goldoraf: Well, the use of flexbox on the form elements force us to stack labels and inputs, but sometimes an "input" may be composed of many elements, and I would like to group these elements into a div that flexes the same way than `coz-form`. Does it make sense to you?

GoOz: It does 😄 
I'll wait to see it in action to make my mind but so far that seems like a good reason to add a new class indeed. 😉

PR #50, opened 2017-02-08
[fix] Modal improvements

y-lohse:
- Adds the proper `pointer` cursor on the close button
- Use a natural width of 90% for small screens (gets bumped up via a media-query for bigger screens)


GoOz:
👍

PR #51, opened 2017-02-08
Alerter improvements

y-lohse:
- Added an optional close callback for the alerter - turns out updating the state after the notification auto-hides is quite useful!
- Made the duration configurable via an extra prop

### DIFF
--- a/react/Alerter/index.jsx
+++ b/react/Alerter/index.jsx
@@ -3,7 +3,7 @@ import classNames from 'classnames'
 
 import styles from './index.styl'
 
-const DEFAULT_AUTOCLOSE_DELAY = 3000
+const DEFAULT_AUTOCLOSE_DELAY = 5000
 
 class Alerter extends Component {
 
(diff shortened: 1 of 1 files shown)


react/Alerter/index.jsx
```diff
@@ -3,7 +3,7 @@ import classNames from 'classnames'
 
 import styles from './index.styl'
 
-const AUTOCLOSE_DELAY = 2000
+const DEFAULT_AUTOCLOSE_DELAY = 3000
```
gregorylegarec: 3 seconds seems pretty short to me. It may feel long during development session but when you got a real notification during navigation, I think it's too short. What think @kelukelu or @jsilvestre about that ?

jsilvestre: You are right, let's go for 5 seconds! Thank you :)

gregorylegarec:
Tanks @y-lohse ! 👍

PR #53, opened 2017-02-08
Add react Modal :speech_balloon:

kosssi:
It's a Modal from @y-lohse model

- [x] Create component
- [x] Refactoring @y-lohse [modal](https://github.com/cozy/cozy-files-v3/pull/75/commits/2c4ae202ec100623979c942e835dbd7a6ea1ad04) on https://github.com/cozy/cozy-files-v3/pull/75

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -1,0 +1,42 @@
+import React, { Component } from 'react'
+import classNames from 'classnames'
+
+import styles from './index.styl'
+
+class Modal extends Component {
+  render () {
+    const {
+      title, description,
+      cancelType, cancelText, cancelAction,
+      validateType, validateText, validateAction
+    } = this.props
+
+    const cancelTypeDefault = cancelType || 'secondary'
+    const validateTypeDefault = validateType || 'primary'
+
+    return (<div className={styles['coz-modal-container']}>
+      <div className={styles['coz-overlay']}>
+        <div className={styles['coz-modal']}>
+          <h2 className={styles['coz-modal-title']}>{title}</h2>
+          <button
+            className={classNames('coz-btn', 'coz-btn--close', styles['coz-modal-close'])}
+            onClick={cancelAction}
+            >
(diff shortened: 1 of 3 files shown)


react/Modal/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+import styles from './index.styl'
+
+class Modal extends Component {
+  render () {
+    const {
+      title, description,
+      cancelType, cancelText, cancelAction,
+      validateType, validateText, validateAction
+    } = this.props
```
enguerran: MAybe we can write that like our other components:

```jsx
export default ({ title, description,
                  cancelType = 'secondary', cancelText, cancelAction,
                  validateType = 'primary', validateText, validateAction }) => (
  // write here the bunch of div
)

File: react/Modal/index.jsx [hunk 2]
```diff
@@ -0,0 +1,42 @@
...
+              {validateText}
+            </button>
+          </div>
+        </div>
+      </div>
+    </div>)
+  }
+}
+
+export default Modal
```
enguerran: We may add `PropTypes` for component in `react-ui` (as well as for our own applications but it is less mandatory). cf. https://facebook.github.io/react/docs/typechecking-with-proptypes.html (use to be [reusable react component](http://reactjs.cn/react/docs/reusable-components.html).

react/Modal/index.jsx
GoOz: FYI, if "secondary" and "primary" refer to button classes, they won't exist for long. There are replacements, of course, but the class name changed during the refactoring that will be released soon. That's fine like this, just wanted to warn you 😄

react/Modal/index.jsx
y-lohse: I think that's a good idea! The Alerter component should probably be retro-fitted with those (we can do it in a separate PR)

enguerran:
I add a PR to help https://github.com/kosssi/cozy-ui/pull/1 (don't know how to commit onto this PR).

PR #59, opened 2017-02-13
[feat] Added mobile styles for alerts

y-lohse:
Update the alerter style to work on mobile.

The `coz-overlay` class is removed because it is never actually used.

Just a head's up, when this PR is merged I'm going to need a new published version of cozy-ui :-)

### DIFF
--- a/stylus/ui-components/alerts.styl
+++ b/stylus/ui-components/alerts.styl
@@ -20,7 +20,7 @@ $alerts
         background-color emerald
         box-shadow       0 em(6px) em(18px) 0 rgba(50, 54, 63, .23)
         box-sizing       border-box
-        padded(.8em, 1em)
+        padded           .8em, 1em
         color            white
         opacity          1
         transition       bottom .5s ease-out, top .5s ease-out, opacity .2s ease-out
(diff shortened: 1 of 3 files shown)


stylus/ui-components/alerts.styl
```diff
@@ -10,39 +10,20 @@ $alerts
...
+        bottom           50px
         left             0
-        margin           0 em(16px)
-        border-radius    em(10px)
+        right            0
         background-color emerald
         box-shadow       0 em(6px) em(18px) 0 rgba(50, 54, 63, .23)
         box-sizing       border-box
-        padding          1em 1.5em
+        padded(.8em, 1em)
```
GoOz: Just wondering, does it work if you write `padded .8em, 1em` or not?

stylus/ui-components/alerts.styl
```diff
@@ -69,7 +50,14 @@ $alerts
...
-            left        50%
-            max-width   alert-width
-            margin-left -(alert-width/2)
-            width       100%
+            top              em(16px)
+            bottom           auto
+            left             50%
+            margin-left      -(alert-width / 2)
+            max-width        alert-width
+            padded(1em, 1.5em)
```
GoOz: Same here.

stylus/ui-components/alerts.styl
y-lohse: Yes it does! I keep forgetting that, amending the PR.

y-lohse:
@GoOz  Ok, all good now; see the latest addition to z-indexes, just in case.

GoOz:
Man, Z-Index is like a plague 😆 
That'll do it for now. Thanks. 🎉

PR #64, opened 2017-02-28
[fix] Prevent alerter to be full height on mobile

y-lohse:


GoOz:
Oh I forgot, could you update the changelog according to your changes?
Like in the 'Unreleased' section and 'Fixed' sub-section, add very briefly what you did?

GoOz:
Thanks! 😄

PR #65, opened 2017-02-28
[feat] Add coz-mobile and coz-desktop util classes

gregorylegarec:


stylus/cozy-ui/defaults.styl
```diff
@@ -42,6 +23,34 @@ body
...
 
+// UTILS
+@import '../ui-base/mixins'
+:global
+    .coz-hidden
+        @extend $hidden
+
+    .coz-desktop
+        @media (max-width: (1023/basefont)rem)
+            display none !important
```
GoOz: Why the `!important`? Did you get any trouble without it? Or did you see some trouble in some use cases I didn't think of?

gregorylegarec: Because of `[role=button], button` selector.

GoOz: Hah I see. It's not really their fault but the fact that their entire selector must have one or more class selector in it. So in this case it totally makes sense, you're right, keep the `!important`. 
You won! 🤘

PR #66, opened 2017-03-01
Fix/readaptation

GoOz:
Cleaning, fixing and readaptation of the previous refactoring.


react/Alerter/index.jsx
```diff
@@ -18,7 +18,7 @@ class Alerter extends Component {
 
   componentDidMount () {
     if (this.shouldAutoClose) {
-      const closeDelay = this.props.duration ? parseInt(this.props.duration) : DEFAULT_AUTOCLOSE_DELAY;
+      const closeDelay = this.props.duration ? parseInt(this.props.duration) : DEFAULT_AUTOCLOSE_DELAY
```
m4dz: Prefer `parseInt(this.props.duration, 10)` to be sure we parse the duration in base 10 ;)

GoOz: I just removed the trailing `;` but sure ok, seems legit 😛

enguerran:
Needed for https://github.com/cozy/cozy-files-v3/pull/137

PR #67, opened 2017-03-01
Labels in pop overs

y-lohse:
Some elements in popover menus are not links, they are buttons, but they all look the same.

### DIFF
--- a/stylus/ui-components/popover.styl
+++ b/stylus/ui-components/popover.styl
@@ -27,6 +27,7 @@ $popover
         border-top  1px solid grey-09
 
     a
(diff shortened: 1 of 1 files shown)


stylus/ui-components/popover.styl
```diff
@@ -27,6 +27,7 @@ $popover
         border-top  1px solid grey-09
 
     a
+    [role=button]
```
GoOz: Could you just add the element `button` as a selector too so every case is covered, please?

GoOz:
Great, thanks 🎉

Issue #54, opened 2017-02-09
Font fighting on some elements


y-lohse:
normalize.css sets the `font-family`of certain elements, for example [buttons](https://github.com/necolas/normalize.css/blob/master/normalize.css#L257). In cozy-ui, we [rely on inheritance](https://github.com/cozy/cozy-ui/blob/v3/stylus/cozy-ui/defaults.styl#L42) to propagate the default font.

This means that buttons with no extra font-declaration get the default sans-serif instead of our Lato. There are actually a few places in our v3 UIs where the wrong font is used, most notably the selection bar in files.

Not sure of the best way to fix this. Should we just declare the `font-family` on a lot more elements in the default styles? Ping @GoOz and @m4dz for input.

GoOz:
Good catch. I guess we could just force the use of Lato on buttons, select, etc using the .coz-btn class and such. As we don't really care about what looks like an unstyled element.
Any thought @m4dz?

y-lohse:
Well, the default `coz-btn` carries quite a bit of extra style with it (backgroud, paddings, etc). Again, I'm mostly thinking about the buttons in the selection bar in files (which, to be fair, could be `a`'s instead of `button`'s), but I'm sure there are others out there.

GoOz:
Fixed in #72

PR #68, opened 2017-03-06
[fix] Font family issue

y-lohse:
This is a proposed fix for #54 (sorry, it bothers me every time I see it, and I see it a lot :D).
Normalize.css only defines `font-family` on 4 occasions, two of which are legitimate, one is the body that we already override, and the last one is the situation that I override in this PR.

### DIFF
--- a/stylus/cozy-ui/defaults.styl
+++ b/stylus/cozy-ui/defaults.styl
@@ -22,6 +22,12 @@ body
     font 1em/1.5
     @extend $font-labor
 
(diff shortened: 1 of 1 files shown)


GoOz:
@y-lohse Timing is a bitch. I actually did the fix in the branch i'm working on (not yet in any PR). Supposed to be merged this sprint.

y-lohse:
Oh cool, nevermind then!

PR #69, opened 2017-03-06
[feat] Adds style for upload button when it's in a menu

gregorylegarec:
I spent a little time "deconstructing" button style to adapt it for a menu item, and then I realized that the better way was to remove classes `btn`, `coz-btn` and `coz-btn--upload` when the upload button is in a bosonic menu ...

Then we just have to specify a background icon.


stylus/cozy-ui/defaults.styl
```diff
@@ -102,6 +102,9 @@ body
     &.coz-btn-alert--success
         @extend $button-alert--success
 
+:global .coz-btn--upload-menu-item
```
GoOz: Those classes are already in a `:global`, you actually don't need it but you do need the `&`

stylus/ui-components/button.styl
```diff
@@ -106,6 +106,11 @@ $button--upload
     background-position  em(17px, 16px) center
     background-repeat    no-repeat
 
+$button--upload-menu-item
```
GoOz: I'm wondering if a more generic name, like $button--upload-link or something, wouldn't be better. Just in case it is re-used out of any menu.

stylus/cozy-ui/defaults.styl
GoOz: I just realised it is an all new button, separated from `.coz-btn`, so forget my suggestions.

stylus/ui-components/button.styl
gregorylegarec: The point is precisely that only upload button in a menu should have those properties.

To mutualize this ruleset, another way should be to declare `coz-btn--upload-primary` for the actual `coz-btn--upload` and use `coz-btn--upload` for this one.

I also hesitated to not use a `coz-btn-*` class but some kind of `coz-menu-item--upload`. Maybe it's the right solution as in the case of Cozy-Photos-v3 this element does not have other button classes like `coz-btn`, `coz-btn--upload`.

stylus/cozy-ui/defaults.styl
gregorylegarec: Yes, the point is it does not fit with other button classes `coz-btn` and `coz-btn--upload`, as it is contained in a menu. See https://marvelapp.com/139hdh2/screen/17497373

stylus/cozy-ui/defaults.styl
```diff
@@ -102,6 +102,9 @@ body
     &.coz-btn-alert--success
         @extend $button-alert--success
 
+:global .coz-btn--upload-link
+        @extend $button--upload-link
```
GoOz: extra useless indentation here

gregorylegarec:
Yep it totally makes sense.

GoOz:
Perfect 🎉

gregorylegarec:
Travis build is failing because of Cozy UI import. https://travis-ci.org/cozy/cozy-photos-v3/builds/208414871, may you have a look @m4dz ?

I assume that it is a transpiling issue, I as faced the same kind of problems with tests tonight.

m4dz:
Travis is on green, let's merge it @GoOz !

gregorylegarec:
@m4dz It was late yesterday night, actually my comment was for another PR, this one https://github.com/cozy/cozy-photos-v3/pull/11 ><
I "mistabbed". 👎

PR #70, opened 2017-03-07
[fix] Avoid hover effect on disabled buttons

gregorylegarec:


gregorylegarec:
@GoOz Works well on my Chrome, Firefox, Safari. Tested on the create album form below.
capture d ecran 2017-03-07 a 12 26 13

Do you think we need a fallback, some kind of 
```stylus
    // [...]
    &:hover
    &:not([disabled]):not([aria-disabled=true]):hover
    // [...]
        border-color      science-blue
        background-color  science-blue
```

GoOz:
Nope, no need for a fallback.

PR #71, opened 2017-03-07
[fix] Arrange sidebar height and selectionbar bottom offset on mobile

gregorylegarec:
There was an issue on photo in mobile mode, a little gap between selection bar and side bar. This commit fixes the problem.


gregorylegarec:
I don't get it neither but ... it works :/

Ok for 48px instead of 50 👍

PR #72, opened 2017-03-07
Fix/cosmetic again

GoOz:
A lot of micro-fix to fit mockups and give our beloved designer some joy.

### DIFF
--- a/stylus/ui-app/layouts.styl
+++ b/stylus/ui-app/layouts.styl
@@ -54,13 +54,9 @@ $app-2panes-sticky
     @extend $app-2panes
 
     @media (max-width: (1023/basefont)rem)
-        height 'calc(100vh - %s)' % em(90px, 16px)
-        // Occupy all the available height minus the heights of the Cozy-bar
-        // (40px) on top & the navigation (50px) on bottom
+        flex-direction column-reverse
 
         > aside
-            position fixed
-            left 0
-            bottom 0
+            flex 0 0 em(50px, 16px)
             width 100%
             z-index $nav-index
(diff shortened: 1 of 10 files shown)


stylus/ui-app/layouts.styl
```diff
@@ -40,20 +40,23 @@ $app-2panes
...
 // When you want the sidebar to act like a sticky footer
 $app-2panes-sticky
     @extend $app-2panes
 
     @media (max-width: (1023/basefont)rem)
-        height 'calc(100vh - %s)' % em(102px)
+        height 'calc(100vh - %s)' % em(90px, 16px)
         // Occupy all the available height minus the heights of the Cozy-bar
-        // (52px) on top & the navigation (50px) on bottom
+        // (40px) on top & the navigation (50px) on bottom
```
m4dz: I ask myself about the flexibility of this one… do we really can rely on a fixed value for cozy-bar and bottom avout their height? Can't it be broken unexpectingly?

GoOz: I don't like it either. Between that and the z-index, I'm struggling to find something less hazardous. So far i can't find any good idea but I'm open to hear new ones. 
Anyway, until then we can use it as it is, but we clearly should find a better way.

y-lohse:
I know these are lots of micro-fixes, but shouldn't there be a mention of some in the changelog?

enguerran:
At least something like "a lot of minor UI fixes"

GoOz:
Damn right. Just do what i say, not what i do xD

PR #73, opened 2017-03-08
Added redux plumbing to the Alerter

goldoraf:
This PR is paired to https://github.com/cozy/cozy-files-v3/pull/152


react/Alerter/index.jsx
```diff
@@ -1,19 +1,94 @@
...
+})
 
+export const alertClear = () => ({
+  type: ALERT_CLEAR
+})
+
+// Dumb component that wraps the alerts
+const Wrapper = ({ t, alerts, dismiss }) => (
+  <div className={styles['coz-alerter']}>
+    {alerts.map(alert =>
```
y-lohse: So if there are several alerts at the same time, they are all going to appear at the same time, right? I don't think the css handles that situation?

react/Alerter/index.jsx
```diff
@@ -22,7 +97,11 @@ class Alerter extends Component {
 
       this.closeTimer = setTimeout(() => {
         this.setState({ hidden: true })
-        if (this.props.onClose && typeof this.props.onClose === 'function') this.props.onClose()
+        if (this.props.onClose && typeof this.props.onClose === 'function') {
```
y-lohse: If I'm not mistaken, `props.onClose` is now always `dismiss`, so it's always set and always a function and this check isn't needed anymore?

react/Alerter/index.jsx
```diff
@@ -22,7 +97,11 @@ class Alerter extends Component {
...
+          this.dismissTimer = setTimeout(() => {
```
y-lohse: Why is there an extra timer before the dismiss action is fired?

enguerran:
What is the purpose to use redux, a centralized state manager, for a single and autonomous component?
Does the client application have to dispatch some alerter's action to interact with the component?

react/Alerter/index.jsx
goldoraf: Yes, I think it'd be useful to be able to display multiple alerts at the same time, but you're right, the CSS can't handle it right now. That's a job for @GoOz :smile:

react/Alerter/index.jsx
goldoraf: Yeah, right

react/Alerter/index.jsx
goldoraf: Because of the fade-out animation: we need to wait after the transition ended before firing the dismiss. If not, the alert would disappear immediately

goldoraf:
@enguerran because we want to display alerts when an action contains an `alert` property: we need to connect the component to the redux store in order to do this. The component also exposes specific action creators, but you're not forced to use them.

react/Alerter/index.jsx
y-lohse: So maybe we should just display them one after the other at the moment? Or just one?

enguerran:
If I add this specific reducer to my app reducers and if I dispatch any action with:

```javascript
{
  type: 'DO_SOMETHING_AWESOME',
  payload: { data: "some data" },
  alert: {
    message: 'awesome alert',
    messageData: 'I don\'t know how it differs from message',
    buttonText: 'Click Me',
    buttonAction: console.log.bind(console),
    level: 'danger'
  }
}
```

it will display an alert?

PR #74, opened 2017-03-08
[fix] fix thead on Safari (requires updated Files & Bar)

GoOz:
That's only a part of the fix.


GoOz:
Related :
https://github.com/cozy/cozy-files-v3/pull/155
https://github.com/cozy/cozy-bar/pull/24

y-lohse:
Does this have any implications for exiting apps (aka. should it be in the changelog?)

GoOz:
Well I don't think so. It doesn't change anything visually on any browser but Safari.

PR #75, opened 2017-03-08
[fix] Layout height

gregorylegarec:
In every browser, Files was looking like this:
image

Removing `height: auto` solves the issue on Firefox, Chrome, and also Safari (except it show a vertical scrollbar).

### DIFF
--- a/stylus/ui-app/layouts.styl
+++ b/stylus/ui-app/layouts.styl
@@ -53,7 +53,6 @@ $app-2panes
 $app-2panes-sticky
     @extend     $app-2panes
     flex        0 0 100%
-    height      auto
     align-items stretch
 
     main,


GoOz:
Did you update Files & Cozy-bar ? because i can't reproduce that bug and I don't want to change anything until I'm sure it exists on the current version

GoOz:
I just did some tests and I'm pretty sure you just didn't update Cozy-bar. :)

gregorylegarec:
@GoOz You're right, after yarnlinking to cozy-bar's last version, it works well (except in mobile/responsive view on my Chrome but it is another concern.

Sorry for the inconvenience !

GoOz:
Don't worry, it was a hard fix on 3 different repos, i knew it wouldn't be smooth 😛

PR #77, opened 2017-03-10
Fix/microfix

GoOz:
- Add a position relative on [role=contentinfo] so we can position elements absolutely in it such as the loading spinner
- Change $icon-n placeholders' width & height with rem instead of em so it doesn't grow with the font-size of the parent. They should be fixed on their sizes.


CPatchane:
Thanks @GoOz 👍

PR #78, opened 2017-03-10
feat: s/validate/primary/ & s/cancel/secondary/

enguerran:
To accept retro-compatibility, I had a `console.warn` for future deprecated props.

I also change two things:
- a new props: `withCross` as a bool to control the cross display
- the default value for props

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -3,57 +3,98 @@ import classNames from 'classnames'
 
 import styles from './index.styl'
 
-const Modal = ({
-  title, description,
-  cancelType, cancelText, cancelAction,
-  validateType, validateText, validateAction, children
-}) => (
-  <div className={styles['coz-modal-container']}>
-    <div className={styles['coz-overlay']}>
-      <div className={styles['coz-modal']}>
-        <h2 className={styles['coz-modal-title']}>{title}</h2>
-        <button
-          className={classNames('coz-btn', 'coz-btn--close', styles['coz-btn-modal-close'])}
-          onClick={cancelAction}
-          >
-          <span className='coz-hidden'>{cancelText}</span>
-        </button>
-        { description && <div className={classNames(styles['coz-modal-content'], styles['coz-description'])}>
-            {description}
-          </div>
+const ModalTitle = ({ title }) =>
+  (
+    <h2 className={styles['coz-modal-title']}>{title}</h2>
+  )
+
+const ModalCross = ({ withCross, secondaryAction, secondaryText }) =>
+  withCross &&
+  (
+    <button
+      className={classNames('coz-btn', 'coz-btn--close', styles['coz-btn-modal-close'])}
+      onClick={secondaryAction}
+      >
+      <span className='coz-hidden'>{secondaryText}</span>
+    </button>
+)
+
+const ModalDescription = ({ description }) =>
+  description &&
+  (
+    <div className={classNames(styles['coz-modal-content'], styles['coz-description'])}>
+      {description}
+    </div>
+  )
+
+const ModalButtons = ({ secondaryText, secondaryAction, secondaryType, primaryText, primaryAction, primaryType }) => {
+  const displayPrimary = primaryText && primaryAction
+  const displaySecondary = secondaryText && secondaryAction
+  return (displaySecondary || displayPrimary) &&
+    (
+      <div className={classNames(styles['coz-modal-content'], styles['coz-modal-buttons'])}>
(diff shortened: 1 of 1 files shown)


enguerran:
Needed for https://github.com/cozy/cozy-files-v3/issues/172

react/Modal/index.jsx
```diff
@@ -3,57 +3,98 @@ import classNames from 'classnames'
...
+  (
+    <h2 className={styles['coz-modal-title']}>{title}</h2>
+  )
+
+const ModalCross = ({ withCross, secondaryAction, secondaryText }) =>
+  withCross &&
+  (
+    <button
+      className={classNames('coz-btn', 'coz-btn--close', styles['coz-btn-modal-close'])}
+      onClick={secondaryAction}
```
y-lohse: Let's hope we'll never have to bind different actions to the cross and the secondary button. But I think this is good as it is.

react/Modal/index.jsx
```diff
@@ -3,57 +3,98 @@ import classNames from 'classnames'
...
+          <button className={classNames('coz-btn', 'coz-btn--' + primaryType)} onClick={primaryAction}>
+            {primaryText}
+          </button>
         }
       </div>
+    )
+}
+
+const Modal = (props) => {
+  props.primaryType = props.primaryType || props.validateType
```
y-lohse: ❤️

lespacedunmatin: Why not use http://babeljs.io/docs/plugins/transform-es2015-parameters/ instead ?

react/Modal/index.jsx
```diff
@@ -3,57 +3,98 @@ import classNames from 'classnames'
...
-  validateType: React.PropTypes.string,
-  validateText: React.PropTypes.string,
-  validateAction: React.PropTypes.func
+  secondaryType: React.PropTypes.string,
+  secondaryText: React.PropTypes.string,
+  secondaryAction: React.PropTypes.func,
+  primaryType: React.PropTypes.string,
+  primaryText: React.PropTypes.string,
+  primaryAction: React.PropTypes.func,
+  cancelType: deprecated('cancelType') || React.PropTypes.string,
```
lespacedunmatin: Why is it not `React.ProTypes.string || deprecated('cancelType')` ?

react/Modal/index.jsx
enguerran: If we need it later... We could change it later.

react/Modal/index.jsx
enguerran: I would like to evaluate `deprecated()` (to display the deprecated warning in the console) but do not want to change the value of the `cancelType` parameter.

And as you can see in the two last examples in the [mdn (logical OR)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_Operators#Logical_OR) page:

> o8 = ''    || false      // returns false
> o9 = false || ''         // returns ""

lespacedunmatin: @enguerran : ok, deprecated is a function that first warn the developpeur that the result will be false soon, and later is really deprecated ?

react/Modal/index.jsx
enguerran: I do not use destructuring here because I think:

```javascript
const Modal = ({ title, description, primaryType, primaryText, primaryAction, secondaryType, secondaryText, secondaryAction, validateType, validateText, validateAction, cancelType, cancelText, cancelAction, withCross }) => { /* etc. */ }
```

was less clean than 

```javascript
const Modal = (props) => { /* etc. */ }
```

mostly when I decomposed the massive Modal component into smaller one by passing them `...props` instead of the props they need explicitly. Especially with a component that declares its propTypes.

But I may be wrong.

What do you think about?

lespacedunmatin: Oh yes, missed the fat arrow. Ok for me.

react/Modal/index.jsx
enguerran: Yes, that is right. they are temporary instructions that [warn the developer](https://github.com/cozy/cozy-ui/pull/78/files/3ec9a0e91a5199bbcf09dd0ee31d40f47db9e33e#diff-c3e57fd2150089dc0703cdf857587297R72). As soon as others apps stop using those props, we remove these declarations (and [the ones in the Modal function](https://github.com/cozy/cozy-ui/pull/78/files/3ec9a0e91a5199bbcf09dd0ee31d40f47db9e33e#diff-c3e57fd2150089dc0703cdf857587297R52)).

PR #84, opened 2017-03-16
Add Empty component style

GoOz:
As there was a similar empty view on Files, Photos and now Settings apps, I added it as a shared component. 
You then just need to specify the path to the icon you want in a `background-image` on the `:before`  when you extend it in your app.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - none yet
 
 ### Added
-- none yet
+- Empty component style
 
 ### Removed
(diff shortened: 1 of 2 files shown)


GoOz:
@CPatchane well, you can just add in Photos app something like that:
```
.pho-empty,
.pho-error
    @extend $empty
```
It seems cleaner to me than adding a new placeholder that would do exactly the same thing, don't you think? Or maybe i didn't get your question 😛

CPatchane:
@GoOz That should also work for now, seems good to me so 👍

m4dz:
Thanks to all :tada:

PR #85, opened 2017-03-17
[fix] Overlay position to absolute to fixed

GoOz:
Because of the `position: relative` on [role=contentinfo] the overlay wasn't taking all the viewport.


m4dz:
:tada:

PR #87, opened 2017-03-22
Add table component style…

GoOz:
…with benefits
- [x] New utils class `.coz-error` with warning option
- [x] Renamed form classes for consistency


stylus/cozy-ui/defaults.styl
```diff
@@ -32,7 +32,20 @@ textarea
 
 // UTILS
 @import '../ui-base/mixins'
+@import '../ui-base/palette'
 :global
+    .coz-error
+        color  pomegranate
+
+        &.coz-error--warning:before
```
m4dz: aren't we exposing too much common classes in global namespace (true question) ?

GoOz: @m4dz Funny you asked because I tend to think that there's not enough. I may be wrong but I think when you make it too flexible, you make it too complicated to use. 
It should be easy as pie for little elements or utility classes, like "you want to display an error text in red then use this class. But if you want a big component, you'll have to extend it mate, too big for the default css". You see what I mean? I'd be happy to discuss this though, despite my firstname initials I ain't omniscient, I may miss a point.

y-lohse: > you want to display an error text in red then [use this class](http://tachyons.io/docs/themes/skins/). But if you want a big component, [you'll have to extend it](http://tachyons.io/components/cards/album-centered/index.html) mate, too big for the default css

Have I ever mentioned how wonderful tachyons is? Yes? What do you mean "get the fuck out of here"?

GoOz: hahaha 😆 
There's 2 philosophy conflicting here. Tachyons has every classes globally extended, classes with very low purposes each, but you have to extend every classes to be effective which mean the app CSS weights more than it should… always.
Or a library of extendable classes that you use when you need them which means flexibility, just the weight of CSS you need, but a more complex use.
The goal i had in mind was kinda meet them in the middle.

stylus/ui-components/table.styl
```diff
@@ -0,0 +1,78 @@
...
+
+        &.coz-table-row:hover
+            background-color white
+
+    .coz-table-body
+        flex      1 1 auto
+        display   flex
+        flex-direction column
+        // Because it needs a height to be scrollable. Don't ask.
+        height    1px
```
lespacedunmatin: Why not `min-height` ?

stylus/cozy-ui/defaults.styl
m4dz: Ok, I merge, keep the debate guys, it's cool :dancer:

stylus/ui-components/table.styl
```diff
@@ -0,0 +1,78 @@
...
+        font-size       12px
+        text-transform  uppercase
+
+    @media (max-width: (1024/basefont)rem)
+        .coz-table-row
+            max-width  100vw
+
+    @media (max-width: (768/basefont)rem)
+        .coz-table-head
+            display none
```
lespacedunmatin: What ?

stylus/ui-components/table.styl
GoOz: Cause it doesn't work… :/

stylus/ui-components/table.styl
GoOz: What what? "table head" is hidden on mobile according to the mockups.

PR #92, opened 2017-04-15
refactor: Remove unnecessary deprecated msg :hocho:

kosssi:
On Cozy Drive we use this Modal with new parameters but we have lot of deprecated message.

This PR remove it. Message are now only display when we used deprecated parameters.


kosssi:
👍 it's done

PR #94, opened 2017-04-28
Toggle component

y-lohse:
This PR adds a very simple toggle component, such as this:

toggle example

It's designed to be an inline-element that scales with the `font-size` of it's parent. It binds to a simple js value and calls a callback whenever it changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - none yet
 
 ### Added
-- none yet
+- Added a new toggle component
 
 ### Removed
 - none yet
--- a/react/Toggle/index.jsx
+++ b/react/Toggle/index.jsx
@@ -1,0 +1,21 @@
+import styles from './styles'
+
+import React, { Component } from 'react'
+
+class Toggle extends Component {
+  onChange () {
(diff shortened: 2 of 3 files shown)


y-lohse:
I think we need *a* way to configure the size of the toggle, as I expect they will comme in different sizes. But it doesn't have to be font-size.
I like font-size in this case because it gives a lot of flexibility at very little cost (compared to pre-defined `toggle-s`, `toggle-m`, `toggle-l`, ... modifiers for example). But I'm open to alternatives!

GoOz:
Actually i'm questionning the idea of various size of this toggle 😛 
I may be wrong but i can't find use case where it would be a good idea for several sizes. I like my form components to be consistent. Only one size of radio button or checkbox. Only inputs can have different sizes but the fewer the better.
But if @kelukelu says like you, then so be it.

y-lohse:
Yeah I checked with him, and while right now there's only one size... it's not set in stone.

GoOz:
In that case, I would use modifier classes anyway. When it's related to the font-size, it's way too unstable.

y-lohse:
Ok, so I'll change the dimensions to something else than `em`, and that will be the default size. We can add the modifiers when we need them.

m4dz:
Thanks dudes :sparkles:

PR #96, opened 2017-05-03
fix: remove white background-color from more button

GoOz:
Removed white background from "more button" as it is mostly used in parents with already a white background and can cause some inconvenience in some case

Basically, the more button doesn't have a white background by default anymore.

### DIFF
--- a/stylus/ui-components/button.styl
+++ b/stylus/ui-components/button.styl
@@ -116,6 +116,7 @@ $link--upload
 
 $button--more
     width                em(46px)
+    background-color     transparent
     background-image     embedurl('../assets/icons/ui/icon-dots.svg')
     background-position  center center
     background-repeat    no-repeat


GoOz:
[(3) [FIL-38] Cosmétique - Cozy Bar](https://trello.com/c/C8qKQwVL/236-3-fil-38-cosmetique-cozy-bar)

GoOz:
Yes, it's inherited from `$button--secondary` but those have a border which the more button don't have so i removed the white background by adding a transparent one because it doesn't really make sense.

I agree, they shouldn't rely on their parents background but from where i'm standing every more button seems to not have a background-color at all, with an exception for the hover state.
Looking at the big picture, i think it makes more sense that this more button don't have a background and if needed one day on specific case, an overwrite will do the trick instead of overwrite it everytime you don't want a background.

I'll update the Changelog, you're right. 

Thanks. 😄

Issue #97, opened 2017-05-03
Long nav link causes overflow from hover background


ptbrowne:
![image](https://cloud.githubusercontent.com/assets/465582/25665163/0a66b152-301d-11e7-98f0-629c1e3e4b49.png)

ptbrowne:
With a bit of CSS

![image](https://cloud.githubusercontent.com/assets/465582/25665309/81a29ec0-301d-11e7-961b-77484802cc4d.png)

```
.coz-nav-link {
    text-overflow: ellipsis;
    white-space: nowrap;
    overflow: hidden;
}
```

GoOz:
Are we sure we want an ellipsis and not a return? I'm kinda skeptical about cuting nav entries but just asking, really.
@kelukelu what do you think?

kelukelu:
I'm agree. 
A return seems to be a better choice. 
We should ask EDF to find a shorter/better name also: "realiser des economies" seems like an action to me. Not really the best name for a tab...

GoOz:
Indeed, they should find the most suitable short term but as this is not always possible, let's go for a return. Thanks @kelukelu.

ptbrowne:
Return is ugly but OK

![image](https://cloud.githubusercontent.com/assets/465582/25669613/4e4c22fa-302a-11e7-8f97-65c4a5c9bb98.png)

GoOz:
Fixed with #98

PR #98, opened 2017-05-03
fix(nav): fix overflowing text from hover. #97

ptbrowne:
https://github.com/cozy/cozy-ui/issues/97

### DIFF
--- a/stylus/ui-app/nav.styl
+++ b/stylus/ui-app/nav.styl
@@ -17,7 +17,7 @@ $navigation
         height          3em
         display         flex
         align-items     center
-        padding-right   0.5em
+        padding-right   .5em
 
         &:hover::before
             content        ''
--- a/stylus/ui-app/nav.styl
+++ b/stylus/ui-app/nav.styl
@@ -17,6 +17,7 @@ $navigation
         height          3em
         display         flex
         align-items     center
+        padding-right   0.5em
 
         &:hover::before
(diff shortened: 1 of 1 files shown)


ptbrowne:
> Same space on the right for every menu item because of the padding of the ul, then add a little padding (like .5em) on items to avoid the text to touch the bottom of this area because it won't be pretty.

Did not understand that.

GoOz:
Yeah sorry, it wasn't really clear. An image is worth 1 000 words they say so…



See, if the word fits perfectly the available space it won't be pretty, there should be an little padding-right on `coz-nav-item` as well just to be safe.

ptbrowne:
OK thanks Understood.

image

stylus/ui-app/nav.styl
```diff
@@ -9,26 +9,36 @@ $navigation
...
 
     .coz-nav-item
-        position  relative
-        z-index   $app-index
+        position        relative
+        z-index         $app-index
+        height          3em
+        display         flex
+        align-items     center
+        padding-right   0.5em
```
GoOz: Can you just remove the 0 and write `.5em`? That's what Travis doesn't like. :)

ptbrowne: I feel this lint rule is a bit overkill.

GoOz:
Thanks! 👍

PR #102, opened 2017-05-12
feat: add button

ptbrowne:


GoOz:
I don't get it. What is it for? From what I understand you can only have `<button>` but not link `<a>` and you can have only one theme class when you're supposed to be able to add as many as you want if there are compatible.

ptbrowne:
My goal here was to abstract the uses of classes and theming.

IMHO it's nicer to use

```
import { Button } from 'cozy-ui'

<Button theme={ 'regular' }>Salut</Button>
```

(regular should be default by the way ?)

than 

```
import classNames from 'classnames'

<button
    role='button'
    className={classNames('coz-btn', 'coz-btn--regular')}
>
    Contacter EDF
</button>
```

It is true that Button could take a `tag` attribute and that Proptypes would be nice.

GoOz:
OK but we won't be able to use this component for specials buttons like the ones with icons.

enguerran:
Why don't we? By creating the correct abstraction we can cover every button's usages. But to be able to do that, we have to list all kind of button's usage.
IMHO, it is an iterative work, and we could add this PR, then by refactoring projects' code, we could enrich the cozy-ui button component.

ptbrowne:
> OK but we won't be able to use this component for specials buttons like the ones with icons.

What is the problem in this case ? I can modify the component.

GoOz:
It's just that for a special button it needs several theme classes at the same time. I'm not saying it's not possible to do that, it's just that so far your component don't handle that case.

ptbrowne:
True. Do you think it's worth going further in the <Button />-inside-cozy-ui direction or do we wait for an abstraction to emerge through the other apps ?

I can live well with the <Button /> inside my project but the first reason I made the PR was to have a discussion on this very issue :)

gregorylegarec:
I agree with @enguerran, let's work by iteration, we can merge this PR right now without introducing any side effect, and iterate over it in the future to enhance the component.

GoOz:
I'm not against merging it now and iterate later, i just wanted to point out that "issue".

gregorylegarec:
Let's merge this, thanks @ptbrowne ! 👍

PR #104, opened 2017-05-16
fix: modal width

gregorylegarec:
Media query rule was dumb, allowing us to only have a tiny modal not responsive to its content width.

This PR can have side effects on existing code in photos, please check that everything is still correct @enguerran <3

### DIFF
--- a/stylus/ui-components/modals.styl
+++ b/stylus/ui-components/modals.styl
@@ -34,6 +34,7 @@ $modals
         display           block
         box-sizing        border-box
         border-radius     em(8px)
+        max-width         90%
         height            auto
         // Padding is set by inner element,
         // to be able to get horizontal separators.
--- a/stylus/ui-components/modals.styl
+++ b/stylus/ui-components/modals.styl
@@ -56,13 +56,14 @@ $modals
             font-weight  bold
 
         .coz-btn-modal-close
-            position        absolute
-            top             1.3em
-            right           1em
-            margin          0
-            padding         0
-            background-size 2em
-            cursor          pointer
+            position         absolute
+            top              1.3em
+            right            1em
+            margin           0
+            padding          0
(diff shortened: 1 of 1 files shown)


stylus/ui-components/modals.styl
```diff
@@ -34,7 +34,6 @@ $modals
         display           block
         box-sizing        border-box
         border-radius     em(8px)
```
GoOz: I don't get why you remove this line. Without it the modal would take full width of the viewport

GoOz: the `width 90%` line…

gregorylegarec: Mmmh actually it does not. 😐

GoOz: Sorry, i meant it would take full width if the content pushes to the edges.
I think a `max-width: 90%` just to secure some margin from the edges would be better.

gregorylegarec: @GoOz Consider it done because ... actually it is.

enguerran:
@gregorylegarec I'll check, stay tuned.

enguerran:
I guess you can add a `min-width` for very thin content like this:

gregorylegarec:
@enguerran This `min-width` would be arbitrary IMHO. It should be the modal content's own width. What do you think ? @GoOz ?

GoOz:
I don't know… The thing is I agree on the idea that the content should define the width of the modal but in the other hand that fat input and button in a so narrow modal bugs me. 
What was the size of this modal before? 480px wide? If so, does this seems like a too high min-width value?

enguerran:
I updated the Modal's content size:

screenshot at 17-24-10

gregorylegarec:
@enguerran It looks *so* fine now 😍

PR #108, opened 2017-05-17
fix(button): missing import

ptbrowne:
The production build breaks since React was not imported.

It is a bit problematic that the linter does not catch this problem.

If the linter cannot be configured to detect this, maybe the ProvidePlugin could be used ?

```
new webpack.ProvidePlugin({ React: 'preact-compat' })
```

### DIFF
--- a/react/Button.jsx
+++ b/react/Button.jsx
@@ -1,3 +1,4 @@
+import React from 'react'
 import cx from 'classnames'
 
 export default function (props) {


gregorylegarec:
Good point. What do you think @m4dz ? @goldoraf ?

gregorylegarec:
Thanks @ptbrowne !

enguerran:
The lack of *React* is not detected because this project do not include tests or a running platform like storybook or else. I think we should go this way.

ptbrowne:
@enguerran I am not sure that tests would find this as while you are developing mode they would not run in production mode. As for the storybook or running platform, I don't know what they are.

Personally, I prefer to have this as soon as I code, before I do my production build and push it to the test server.

I like `eslint` and the `eslint-plugin-react` can find this problem : 

image


My eslint config :

```
yarn add eslint eslint-plugin-react
```

In `.eslintrc.json`

```
{
  "parserOptions": {
    "ecmaVersion": 6,
    "ecmaFeatures": {
      "jsx": true
    },
    "sourceType": "module"
  },
  "plugins": ["react"],
  "rules": {
    "react/jsx-uses-react": "error",
    "react/jsx-space-before-closing": [0, "never"],
    "react/react-in-jsx-scope": "error"
  }
}
```

PR #109, opened 2017-05-18
fix: there is no button.primary style :lipstick:

enguerran:
I though I already fix that, but it seems to be back.

`s/primary/regular/` to fit button's classes.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -78,7 +78,7 @@ Modal.propTypes = {
 
 Modal.defaultProps = {
   primaryType: 'secondary',
-  secondaryType: 'primary',
+  secondaryType: 'regular',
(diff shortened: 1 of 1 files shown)


enguerran:
It may be the second option.

enguerran:
I looked for what happened, and I found two commits:

1. change the primaryType to `secondary` in https://github.com/cozy/cozy-ui/commit/3ec9a0e91a5199bbcf09dd0ee31d40f47db9e33e#diff-c3e57fd2150089dc0703cdf857587297R96
2. change the type to an unexisting css class in https://github.com/cozy/cozy-ui/commit/8aea847cff23af56a33096d971fd638486728cf6

GoOz:
ok :)

PR #112, opened 2017-05-19
Handle closing modal by outside click and escape key

CPatchane:
In order to save @y-lohse for the next demos ;)

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { Component } from 'react'
 import classNames from 'classnames'
 
 import styles from './styles.styl'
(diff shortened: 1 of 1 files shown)


react/Modal/index.jsx
```diff
@@ -49,9 +49,27 @@ const ModalButtons = ({ secondaryText, secondaryAction, secondaryType, primaryTe
...
+  document.addEventListener('keyup', handleKeyUp)
+
+  const handleOutsideClick = (event) => {
+    if (event.target.className === overlayClassName) props.secondaryAction()
+  }
+
   return (
     <div className={styles['coz-modal-container']}>
-      <div className={styles['coz-overlay']}>
+      <div onClick={handleOutsideClick} className={styles['coz-overlay']}>
```
y-lohse: You might use `overlayClassName` here just in case, but whatever :-)

react/Modal/index.jsx
```diff
@@ -49,14 +49,27 @@ const ModalButtons = ({ secondaryText, secondaryAction, secondaryType, primaryTe
 
 const Modal = (props) => {
   const children = props.children
+  const overlayClassName = styles['coz-overlay']
 
-  const onOutsideClick = (event) => {
-    if (event.target.className === styles['coz-overlay']) props.secondaryAction()
+  const handleKeyUp = (event) => {
+    if (document.querySelector(`.${overlayClassName}`)) {
+      if (event.keyCode === 27) {
```
gregorylegarec: I suggest to use a constant for the keyCode, to clarify code.
```js
if (event.keyCode === KEY_ESC)
```

react/Modal/index.jsx
CPatchane: You're right, I fix that 👍

CPatchane:
Thanks @ptbrowne I was hesitated about making a whole react component, but it seems cleaner when seeing your proposal. It's a component now 👍

PR #113, opened 2017-05-19
chore: export styles so that external components can use it

ptbrowne:
To be able to get styles['coz-modal-content'] from external components.

import Modal, { styles as modalStyles } from 'cozy-ui/react/Modal'

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -2,6 +2,7 @@ import React from 'react'
 import classNames from 'classnames'
 
 import styles from './styles.styl'
+import Content from './Content'
 
 const ModalTitle = ({ title }) =>
   (
@@ -82,5 +83,5 @@ Modal.defaultProps = {
   withCross: true
 }
 
-export { styles }
+export { Content }
 export default Modal
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -82,4 +82,5 @@ Modal.defaultProps = {
   withCross: true
 }
 
+export { styles }
 export default Modal


y-lohse:
Not sure why that's needed, but I have nothing against it!

y-lohse:
Pulling in @gregorylegarec since the PR has changed a bit.

gregorylegarec:
Oops ... missed this one.

I think that @GoOz's objective in the incoming refactoring is to provide all our classes independently from our React components. But it could be a good solution if it is an urgent need. Is it the case @ptbrowne  ?

react/Modal/index.jsx
```diff
@@ -82,5 +83,5 @@ Modal.defaultProps = {
   withCross: true
 }
 
-export { styles }
+export { Content }
```
gregorylegarec: If I understand well, the actual goal is to avoid a double import and write something like :
`import Modal, { Content as ModalContent } from 'cozy-ui/react/Modal'` ? I like it.

GoOz:
I confirm. The future goal is to have properly independant styles from functional component.

react/Modal/index.jsx
ptbrowne: Avoiding the double import is actually a side effect (a nice one). To me it was also to hide the organization of the folders to the end developer.

gregorylegarec: So it works for me. If @GoOz is ok I'll merge it.

GoOz: yeah yeah i'll do my mixture later when refactoring :p
Go for it.

gregorylegarec:
Nice job @ptbrowne 💯

PR #117, opened 2017-05-29
feat: Avoid overflowing for modal

gregorylegarec:
Must be merged this afternoon.

Modal now stick to top when the viewport height is too tiny.

capture d ecran 2017-05-29 a 14 46 56

Requesting review from @GoOz at least for information.

I think that this PR should not trigger regression but I prefer to request reviews for the most people using the modal component as possible.


gregorylegarec:
Thanks for your reviews guys. @GoOz, don't forget to have a look please ❤️

gregorylegarec:
I think both ways have pros and cons. Actually I think we can do both with a smart little flag or something.

PR #118, opened 2017-05-29
feat: remove vertical paddings in modal

gregorylegarec:
The goal is to avoid default vertical paddings and let <ModalContent> component add them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - none yet
 
 ## [3.0.0-beta29] - 2017-XX-XX
+### Changed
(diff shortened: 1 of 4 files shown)


gregorylegarec:
@y-lohse May you have a look please and tell us if it will be ok for the modal in cozy-photos ? 💋

gregorylegarec:
Those changes are breaking UI in at lease Cozy-Drive, I prefer to find a solution for them before merging this PR.

gregorylegarec:
I did add some commit to ensure compatibility with Drive and Photos.
I also integrated `ModalContent` component directly into `Modal`.
Drive will be ok but Photos will need some little changes, my PR is already ready. I may handle this when cozy UI will be released. @y-lohse @kosssi

ptbrowne:
Good to have the ability to specify `classNames` ! Thanks @gregorylegarec !

PR #126, opened 2017-05-31
fix: no hard-coded height for buttons

ptbrowne:
I am currently working on the responsive version of the EDF app.

The height of the button is hard-coded here
https://github.com/cozy/cozy-ui/blob/632a397076b2794953b26e5918457f68db17d7f1/stylus/ui-components/button.styl#L19

I think we can rely on the text setting the height of the button.

Problem
image

Solution
image

### DIFF
--- a/stylus/ui-components/button.styl
+++ b/stylus/ui-components/button.styl
@@ -16,7 +16,6 @@ $button
     margin           0 .25em
     border           1px solid white
     border-radius    2px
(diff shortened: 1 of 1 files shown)


ptbrowne:
Yes I think it could :) Thanks for the idea.

gregorylegarec:
Thanks @ptbrowne ! 👍

Issue #127, opened 2017-06-01
Font smoothing


ptbrowne:
I think it would be good to add that to the default styles.

```css
body {
  -moz-osx-font-smoothing: grayscale;
  -webkit-font-smoothing: antialiased;
}
```

![2017-06-01 10_10_38](https://cloud.githubusercontent.com/assets/465582/26670633/a5499f94-46b2-11e7-9caf-c33e2cbb6104.gif)

To be clear : when the font is bolder (and ugly), font-smoothing is not active.

![image](https://cloud.githubusercontent.com/assets/465582/26670577/7049b824-46b2-11e7-98a7-1f3b86b72bb3.png)

PR #128, opened 2017-06-01
feat: font smoothing 💄

ptbrowne:
Fixes https://github.com/cozy/cozy-ui/issues/127

To be clear : when the font is bolder (and ugly), font-smoothing is not active.

### DIFF
--- a/stylus/ui-base/fonts.styl
+++ b/stylus/ui-base/fonts.styl
@@ -20,7 +20,6 @@ $font-title
 $font-labor
     font-family  Lato, sans-serif
 
-
 // Contextual fonts store, based on the selector checksum hash
 basefonts = {}
 
--- a/stylus/cozy-ui/defaults.styl
+++ b/stylus/cozy-ui/defaults.styl
@@ -20,6 +20,8 @@
 
 body
     font 1em/1.5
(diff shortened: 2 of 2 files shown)


stylus/ui-base/fonts.styl
```diff
@@ -20,6 +20,10 @@ $font-title
 $font-labor
     font-family  Lato, sans-serif
 
+$font-smoothing
+    -moz-osx-font-smoothing grayscale
```
GoOz: Why writing a placeholder? why not writing it as default style for fonts directly on the body? I don't see any usecase where you could re-use that. :/
Just curious really. 😄

ptbrowne: I did the same as for `$font-labor` trying to keep the code coherent :) I can write directly it on the `body`

GoOz: Well, either the font-smoothing is only relevant with Lato font and in that case just add this in the `$font-labor` placeholder, or it's for any font and in that case, just directly write it in body declaration in `default.styl`.

See what i mean?

ptbrowne: I see :) Done :)

PR #130, opened 2017-06-01
feat: Icon component

ptbrowne:
The Icon component can be used like : 

```jsx
import Icon from 'cozy-ui/react/Icon'
<Icon icon='download' />
```

To be able to use an icon with `Icon`, put it in `assets/icons/ui/` and make sure its filename starts with `icon-` and finishes with `.svg`.

It inserts this in the DOM and uses webpack's `sprite-loader`. 

```html
<svg class="icon--1STEd" width="1em" height="1em">
   <use href="#icon-download_53d6a68a9732b5d98886a9737a2a7089"></use>
</svg>
```
☠️  I had to remove the `fill` attribute on the SVG files to be able to style them with CSS afterwards. At the moment they have `fill: currentColor` so they take the color of the surrounding text automagically ✨ . It is a __breaking change__ for the applications depending on cozy-ui icons, if they use directly the SVG, they need to put back the '#FFF' fill themselves.

[EDIT]  Example : 

```jsx
const HelloWorld = function {
   return <span style={{ color: 'red' }}><Icon icon='download'/> Hello world</span>
}
```

Here the download icon will be red like the surrounding text.


ptbrowne:
I think that to be safe, I should duplicate the icons to remove their `fill` attribute.

kosssi:
or maybe you can have a default color and if you have an attribute color you change it.

    <Icon icon='download' color='red' />

gregorylegarec:
I prefer also @kosssi's proposition, is it possible ?

CPatchane:
I agree with @GoOz  about using SVG as background and I don't think that duplicate icons will be a good solution because it will be hard to maintain and to keep compliant everywhere.
If the @kosssi proposal can avoid that, it would be better, if not I would prefer to keep it like that at least for now.

ptbrowne:
> I prefer also @kosssi's proposition, is it possible ?

Yes since it is only CSS at this point.

I could add this to the Component

```
style={{ fill: this.props.color }}
```

Having a default color per app is also possible.

```
.icon {
   fill: tomato
}
```

> I don't think that duplicate icons will be a good solution because it will be hard to maintain and to keep compliant everywhere.

Having duplicated icons is a temporary solution that is not very satisfying for sure. I do not see any advantage to the icons with `background-image` so if the `Icon` component is used, it would be the new default and `background-image` will disappear. I would be happy to know any advantage of `background-images` over `Icon`.

GoOz:
> Having duplicated icons is a temporary solution that is not very satisfying for sure. I do not see any advantage to the icons with background-image so if the Icon component is used, it would be the new default and background-image will disappear. 

Indeed, and if we go with this method, there will be a need for a transition time when both ways coexist and duplication is surely dirty but will do the trick. and when transition time is done, we'll just have to remove old svg and CSS

> I would be happy to know any advantage of background-images over Icon.

Mostly performance, really. As the svgs are embed in the CSS files and cached, it costs less in weight and request

ptbrowne:
> Mostly performance, really. As the svgs are embed in the CSS files and cached, it costs less in weight and request

Here the SVG would be embedded/cached in the JS :)

ptbrowne:
In https://github.com/cozy/cozy-ui/pull/130/commits/48ad10c12c2c90e12dd9064d3758f50a75ad5120, I tried to concile both worlds 🙏 

`Icon.jsx` loads the `SVG` by hard-coding the `sprite-loader` and combines it with the custom `strip-fill` loader which takes care of removing the fill attribute and emiting a deprecation warning. It seems fitting to me to bypass the application defined loaders here as we are explicitly relying on the result of the loader to be a sprite hash.

It means that the code of application relying on `background-image`s does not need to change ✨ !

GoOz:
> Here the big SVG sprite would be embedded/cached in the JS :)

Ah nice but… there's one thing one should not forget. Cozy-UI's React components will only be used by us and external developers that will choose React to code. That was a big concern for us, to not impose a JS lib over an other to developers. 
That being in mind, your way with React does fix an issue that will not be fixed with others lib (unless the external developer does the same trick on his side) that means that we have to have a strong & reliable CSS framework as well. That's my main concern in all this. That's why i'm saying that's not an easy task to change all of this.

ptbrowne:
A way to promote the use of inline SVGs without being reliant on React would be to split the current `Icon` component into 2 files :

* An `icons` file that would export SVG hashes to be used with `<svg><use /></svg>`
* the React Component that would use this file to get icon hashes and then render`<use/>`

A developer that would not use React could do : 

```jsx
const { danger } = require('cozy-ui/icons')

export default function () {
  return `Hello World <svg><use xlink:href="${danger}" /></svg>`
}
```

ptbrowne:
I splitted the code in https://github.com/cozy/cozy-ui/pull/130/commits/9ad0843129760852ab60e3b716bf5b9f434d6c07.

GoOz:
I like this better 😉 
I don't see how it could but could you just check if this doesn't interfere with elements that uses the old way? Just to be sure before merging.

m4dz:
@GoOz do you take this one to get sure it's ok? Or do we merge it as it?

GoOz:
@m4dz I could, indeed. I may have some free time this morning, I'll try to check out this branch on Drive & Photos and check if there's any regression.

GoOz:
Well it looks good to me. I didn't see any glitch as expected. 
@ptbrowne could you just update the changelog? I think you're the best person to explain briefly what's new 😛 Thanks!

ptbrowne:
Done. I also added properties to customize the `Icon` : 

* `className`
* `color`
* `style`
* `width`
* `height`

GoOz:
Nice ! Thanks 👍

PR #132, opened 2017-06-02
feat: better paddings in modal section :lipstick:

gregorylegarec:
capture d ecran 2017-06-02 a 17 29 39

### DIFF
--- a/stylus/ui-components/modals.styl
+++ b/stylus/ui-components/modals.styl
@@ -52,17 +52,23 @@ $modals
         // hide the border-radius
         overflow          hidden
 
-        .coz-modal-content,
-        .coz-modal-section
+        .coz-modal-content
             padding 0 1.5em
+
+        .coz-modal-section
+            padding 1.5em
 
         .coz-modal-section
             border-top 1px solid grey-09
 
+        .coz-modal-content:first-of-type
+                padding-top    1.5em
+
+        .coz-modal-section:first-of-type
+                padding-top    1em
+
         .coz-modal-content,
         .coz-modal-section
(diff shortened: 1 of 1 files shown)


GoOz:
Yeah don't step on each other on this. @y-lohse did you merge you commit on this on Photos yet?

stylus/ui-components/modals.styl
```diff
@@ -52,17 +52,23 @@ $modals
...
+        .coz-modal-content
             padding 0 1.5em
 
         .coz-modal-section
+            padding 1.5em
+
+        .coz-modal-section
             border-top 1px solid grey-09
 
+        .coz-modal-content:first-of-type
```
GoOz: `:first-of-type` is dangerous ground. This is not really maintenance friendly.
I don't know what you're reffering to but it seems to me that better class architecture would be a better option.
I get that the use of `:first-of-type` was here before but it gets more and more messy which just prove that maintenance will be difficult.

y-lohse:
No nothing merged so far onmy end, but il will be pretty soon.

stylus/ui-components/modals.styl
gregorylegarec: What do you suggest ? Use some kind of `coz-modal-content--first` class ? It's a pain to implement and always complicate code.
Do you have some examples of maintenance issues with `:first-of-type` ?

GoOz: I'll have a look, I must have a global look to know what would be best. Stay tuned.

gregorylegarec:
Out of date and not needed anymore.

PR #133, opened 2017-06-07
feat: add tracking helper

y-lohse:
This tracking helper code is now used across our apps, so it would be good to have it somewhere in common. WHile cozy-ui is not ideal, it's the only place we have right now and there are other purely js helpers here, so... ¯\_(ツ)_/¯

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,5 +45,8 @@
     "classnames": "^2.2.5",
     "md5": "^2.2.0",
     "normalize.css": "5.0.0"
+  },
+  "peerDependencies": {
+    "piwik-react-router": "^0.8.2"
   }
 }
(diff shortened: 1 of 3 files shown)


y-lohse:
As you should :D

y-lohse:
Just a heads-up (especially for @m4dz since you already reviewed it): I made a [small change](https://github.com/cozy/cozy-ui/pull/133/commits/efcb2f7b7ad6f97c797717cbdd47c4449799eaaa#diff-f64c82b61ef0be91b3b40db448b8b85dL89) to avoid using he spread operator; turns out our project are not equiped for dealing with them when they come from `cozy-ui`, so instead of updating all the webpack configs I changed that. 

I'm merging this PR in a couple of hours unless someone shouts.

PR #135, opened 2017-06-07
[WIP]chore: Add storybook to display example :newspaper:

kosssi:
I would like add storybook to document react component.

I have few problem with the configuration: 
- [ ] components use stylus
- [ ] Storybook use webpack 2

If you want to continue, do not hesitate

### DIFF
--- a/docs/react/.storybook/config.js
+++ b/docs/react/.storybook/config.js
@@ -1,0 +1,11 @@
+import { configure, setAddon } from '@storybook/react';
+import infoAddon from '@storybook/addon-info';
+
+setAddon(infoAddon);
+
+function loadStories() {
+  require('../stories/Toggle.js')
+  // You can require as many stories as you need.
+}
+
+configure(loadStories, module)
--- a/docs/react/.storybook/webpack.config.js
+++ b/docs/react/.storybook/webpack.config.js
@@ -1,0 +1,16 @@
+module.exports = {
+  resolve: {
+    extensions: ['.styl'],
+    extensions: ['.jsx'],
+    alias: {
+      'react': 'preact-compat',
+      'react-dom': 'preact-compat'
+    }
+  },
+  module: {
+    loaders: [{
+      test: /\.styl$/,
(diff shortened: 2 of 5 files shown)


nono:
I've tried to fix some things: https://github.com/cozy/cozy-ui/commit/2cee848bf90c5f6e7daa10175ac3d0f47724b767.

At least, it compile and we can open the storybook in a browser. But it is still very buggy:

- the `props` are not passed to the `Toggle` component
- `styles` is an empty object (stylus)

enguerran:
@nono @kosssi I think the @nono's commit is very fine as without doing anything I could get this:

screenshot at 13-17-58

@nono I did not understand what you mean by 
> - the props are not passed to the Toggle component
> - styles is an empty object (stylus)

nono:
@enguerran 

> the props are not passed to the Toggle component

https://github.com/cozy/cozy-ui/commit/2cee848bf90c5f6e7daa10175ac3d0f47724b767#diff-b6440b691d0c70e68d9c55920701ffbaR12

Here, `props` is undefined and if I remove this line, it throws an exception.

> styles is an empty object (stylus)

The toggle should be green, not blue like on your screenshot. https://github.com/cozy/cozy-ui/commit/2cee848bf90c5f6e7daa10175ac3d0f47724b767#diff-b6440b691d0c70e68d9c55920701ffbaL1 --> `styles` is an empty object.

enguerran:
@nono 

> > the props are not passed to the Toggle component
>2cee848#diff-b6440b691d0c70e68d9c55920701ffbaR12

Ohhh, that is a preact/react different behavior. Preact's render function accept parameters when React's render don't. Cf. https://github.com/developit/preact/blob/b03bb360b123f95ebe6b2d08944eb32a92863745/src/preact.d.ts#L66

enguerran:
@nono, I'll check on the style object. But it let me believe that webpack's configuration do not let `import styles from './styles.styl'` be loaded with the correct loader.

nono:
I close this in favor of #153

Issue #139, opened 2017-06-12
Need webpack to import cozy-ui ?


ptbrowne:
We need to have `webpack.config.cozy-ui.js` files. I think it should not be the responsibility of an application to know how cozy-ui is coded so that it should know which loaders are necessary.

One solution would be to ship a built version on npm.

It would then be hard for people hacking on `cozy-ui` through another application to use it since they would have to build each time they make a change. Maybe `cozy-ui` should have a webpack file ready for people `linking` directly and wanting to hack on `cozy-ui` while they develop their application. This way, a user wanting to hack on `cozy-ui` could do:

```js
const webpackConfig = merge(
  ...,
  ...,
  require(`cozy-ui/webpack`)
)
```

temporarily in their file ?

What do you think ?

Discussion about that : https://github.com/webpack/webpack/issues/378

gregorylegarec:
I need to make some tests. I am curious about @m4dz and @GoOz opinions.

GoOz:
@ptbrowne do have a point. I don't know about this particular solution as i'm no expert with Webpack but it seems pretty simple and effective to me.

CPatchane:
Seems to be a good point for me, that's also a good way to push a config update only related to cozy-ui through a cozy-ui release. 👍

ptbrowne:
This will be done in a cozy-webpack.

PR #143, opened 2017-06-16
icons.js -> icons.jsx

ptbrowne:
Since JS files are not automatically `babel`ized, we need the .jsx extension for files that use `import`

This is supposed to be temporary since we should find a way
to have cozy-ui tell applications how it should be transpiled.

https://github.com/cozy/cozy-ui/issues/139


enguerran:
Why did you create a exact same PR #145 ?

ptbrowne:
Because Travis was failing as there was no branch cozy:icon-jsx (Travis tries to do `git clone --depth=50 --branch=icon-jsx https://github.com/cozy/cozy-ui.git cozy/cozy-ui`). So I've done my PR from cozy:icon-jsx to cozy:master. It seems that since I've created this branch, the test is now passing on this PR too.

PR #144, opened 2017-06-19
refactor: simplify tabs component

enguerran:
Use it as below:

```jsx
<Tabs initialActiveTab='hello'>
  <Tab name='hello'>
    <h1>Hello World!</h1>
  </Tab>
  <Tab name='night'>
    <p>Godd Night Moon!</p>
  </Tab>
</Tabs>
```

What do you think? Do I forget to migrate some behavior or any specific style?

### DIFF
--- a/react/Tabs/index.jsx
+++ b/react/Tabs/index.jsx
@@ -1,66 +1,52 @@
-import React, { Component } from 'react'
+import React from 'react'
 import styles from './styles.styl'
 import classnames from 'classnames'
 
-export const Tab = function ({ name, children, className, active, activeClass, changeTab }) {
-  const activeStyle = {
-    [styles['coz-tab--active']]: active
-  }
-  if (activeClass) {
-    activeStyle[activeClass] = active
-  }
-  return <div
-    className={classnames(
-      styles['coz-tab'],
-      className,
-      activeStyle)}
-    onClick={() => changeTab(name)}>{
-    children
-  }</div>
-}
+export const Tab = ({ key, children, className }) => (
+  <div
+    role='tabpanel'
+    aria-labelledby={key}
+    className={classnames(styles['coz-tab-panel'], className)}>
+    {children}
+  </div>
+)
 
-export const TabList = function ({ children, activeTab, changeTab, className }) {
-  return <div className={classnames(styles['coz-tab-list'], className)}>{
-    React.Children.map(children, child =>
-      React.cloneElement(child, {
-        active: child.props.name === activeTab,
-        changeTab
-      }))
-  }</div>
-}
-
-export const TabPanel = function ({ children, activeTab, name, changeTab, className }) {
-  return activeTab === name ? <div className={classnames(styles['coz-tab-panel'], className)}>{
-    children
-  }</div> : null
-}
-
-export const TabPanels = function ({ children, activeTab, name, changeTab, className }) {
-  const extra = { activeTab, changeTab }
-  return <div className={classnames(styles['coz-tab-panels'], className)}>{
-    React.Children.map(children, child => React.cloneElement(child, extra))
-  }</div>
-}
-
-export class Tabs extends Component {
+export class Tabs extends React.Component {
   constructor (props) {
     super(props)
-    this.changeTab = this.changeTab.bind(this)
-  }
-  getInitialState () {
-    return { activeTab: this.props.initialActiveTab }
-  }
-
-  changeTab (tabName) {
-    this.setState({ activeTab: tabName })
+    this.state = {
+      active: this.props.initialActiveTab
+    }
   }
 
-  render ({ children, className }, { activeTab }) {
-    const changeTab = this.changeTab
-    return <div className={classnames(styles['coz-tabs'], className)}>{
-      React.Children.map(children, child =>
-        React.cloneElement(child, { activeTab, changeTab })
-      )
-    }</div>
+  changeTab (tab) {
+    this.setState(state => Object.assign({}, state, {active: tab.props.name}))
+  }
+
+  isActiveTab (tabName) {
+    return this.state.active === tabName
+  }
+
+  render () {
+    return (
+      <div className={classnames(styles['coz-tabs'], this.props.className)}>
+        <div role='tablist' className={classnames(styles['coz-tab-list'])}>
+          {React.Children.map(this.props.children, tab => (
+            <div
+              className={classnames(styles['coz-tab'], { [styles['coz-tab--active']]: this.isActiveTab(tab.props.name) })}
+              onClick={e => this.changeTab(tab)}
+              key={tab.props.name}
+              aria-selected={this.isActiveTab(tab.props.name)}
+              aria-controls={tab.props.name}
+              role='tab'>
+              {tab.props.title}
+            </div>
+          ))}
+        </div>
+        <div className={classnames(styles['coz-tab-panels'])}>
+          {React.Children.map(this.props.children, tab => this.isActiveTab(tab.props.name) && tab)}
+        </div>
+      </div>
+    )
   }
 }


ptbrowne:
It was the initial way of doing things from @y-lohse. Together with him, we discussed and agreed that it is very likely that a tab will contain in the future a badge, an icon or something beside simple text. This is why we went with the more verbose version.

enguerran:
My opinion is that it is a nonsense not to simplify API. I think it is not really helpful to produce something like below as it do not remove internal complexity. What if we forget to create `<TabPanels />`? What is the difference between `<TabList />` and `<Tabs />` as their names are not really explicit? Why do I need to declare the tab `name` a lot of time? And so on...?

```jsx
<Tabs initialActiveTab='hello'>
  <TabList>
    <Tab name='hello'>
      Hello
    </Tab>
    <Tab name='night'>
      Night
    </Tab>
  </TabList>
  <TabPanels>
    <TabPanel name='hello'>
      <h1>Hello World!</h1>
    </TabPanel>
    <TabPanel name='night'>
      <p>Good Night Moon!</p>
    </TabPanel>
  </TabPanels>
</Tabs>
```

At [bootstrap](http://getbootstrap.com/javascript/#markup), here is the API for tabs:

```html
<div>

  <!-- Nav tabs -->
  <ul class="nav nav-tabs" role="tablist">
    <li role="presentation" class="active"><a href="#hello" aria-controls="hello" role="tab" data-toggle="tab">Hello</a></li>
    <li role="presentation"><a href="#night" aria-controls="night" role="tab" data-toggle="tab">Night</a></li>
  </ul>
  <!-- Tab panes -->
  <div class="tab-content">
    <div role="tabpanel" class="tab-pane active" id="hello">
      <h1>Hello World!</h1>
    </div>
    <div role="tabpanel" class="tab-pane" id="night">
      <p>Good Night Moon!</p>
    </div>
  </div>

</div>
```

If React do not bring something more powerful, why do we use it instead of well-tried bootstrap?

y-lohse:
Personally I like the simple API a lot more, but if we need to support something else than simple strings in the tab headers, it's simply not an option.
@ptbrowne : it is my understanding that badges were already needed and used, is that correct? Otherwise we could check with @kelukelu whether or not we'll ever need that.

And if we do need to support HTML in tab headers, we could possibly simplify the API to something like this:

```jsx
<Tabs initialActiveTabIndex={0}>
  <TabHeaders>
    <TabHeader>
      Hello
    </TabHeader>
    <TabHeader>
      Night
    </TabHeader>
  </TabHeaders>
  <TabPanels>
    <TabPanel>
      <h1>Hello World!</h1>
    </TabPanel>
    <TabPanel>
      <p>Good Night Moon!</p>
    </TabPanel>
  </TabPanels>
</Tabs>

```

But that's a very marginal simplification and and not necessarily a good idea. Maybe there are other ways to shape this Component so it can support what we need?

enguerran:
You need to remember that a component's `props` could be any JavaScript object you want, even a function or a JSX tree.

y-lohse:
Fair enough. So something like this would work?

```jsx
<Tabs initialActiveTab={MyTabHeader}>
  <Tab name={MyTabHeader}>
    <h1>Hello World!</h1>
  </Tab>
  <Tab name='night'>
    <p>Godd Night Moon!</p>
  </Tab>
</Tabs>
```

I feel like using a JSX Tree as an `active` flag in the state is a bit weird, isn't it? Though I would still prefer that over the current solution.

enguerran:
Lookazat, I'z got freind: https://toddmotto.com/creating-a-tabs-component-with-react/

ptbrowne:
> @ptbrowne : it is my understanding that badges were already needed and used, is that correct? 

No it has not been needed yet.

> What if we forget to create <TabPanels />?

* The Tab component could check its children to produce a very good error message

>  What is the difference between <TabList /> and <Tabs /> as their names are not really explicit? 

I agree that the API here is a bit large. 

> And my opinion is to wait for the needs to add the behavior because an if is not a need.

I see your point. I already have designed a tab component and very soon after designing it with the simple solution of using the name as a string, I had a to add a badge. I do not think it is too premature to add it.

Nonetheless, I agree the API could be simplified.  I do not think that passing a React component as an attribute is such a good idea.

I think we could simplify into something like : 

```jsx
<Tabs>
   <Tab>1</Tab>
   <Tab>2</Tab>
   <Panel>
     1234
   </Panel>
   <Panel>
     4567
   </Panel>
</Tabs>
```

and rely on the index to remember the selection.

enguerran:
The solution with a set of _tab title_ and a set of _tab panel_ is a risk to let source of truth be duplicated and has a consequence of a desynchronization risk. And if we set aside such an architecture, it let us the opportunity to have a set of tab representing a rich/simple title and a rich/simple panel content.

```jsx
<Tabs>
  <Tab title={withBadge('Inbox', unread)}>
    <Emails mails={emails} />
  </Tab>
  <Tab title={withBadge('Trash', trashed)}>
    <Emails mails={trashedEmails} />
  </Tab>
</Tabs>
```

with something like:

```jsx
const withBadge = (title, badgeContent) => (props) => (
  <h1>{title} <span className='badge'>{badgeContent}</span></h1>
)
```

---

La solution avec un ensemble de _titre d'onglet_ et un ensemble de _contenu d'onglet_ est un problème pour garder une source unique de vérité et a ainsi la conséquence d'un risque de désynchronisation. Or si on met de côté cette architecture, ça nous laisse l'opportunité d'avoir un unique ensemble d'onglet représentant un titre riche ou simple et un contenu riche ou simple.

goldoraf:
I agree with @enguerran: the simpler, the better. And besides, we should always have a look at existing APIs when working on that kind of reusable UI components. If we look at [react-toolbox](http://react-toolbox.com/#/components/tabs), their API is similar to what @enguerran proposes, with the exception of the `index`and `onChange` props. Adding the support of badges could simply be a matter of adding a `icon` prop.

Looking beyond that question of tabs, I'd love to see a component with the behavior describe here: http://bkardell.com/common-panel/prototype/panelset-element.html

Basically, tabs are just one way of displaying a set of panels for a specific range of screen sizes. Accordion is another way, etc...

react/Tabs/index.jsx
```diff
@@ -1,66 +1,39 @@
...
-      React.Children.map(children, child =>
-        React.cloneElement(child, { activeTab, changeTab })
-      )
-    }</div>
+  render () {
+    return (
+      <div className={classnames(styles['coz-tabs'], this.props.className)}>
+        <div className={classnames(styles['coz-tab-list'])}>
+          {React.Children.map(this.props.children, tab => (
+            <div
```
lespacedunmatin: SHouldn’t the tabs be buttons ? See https://www.deque.com/blog/a11y-support-series-part-1-aria-tab-panel-accessibility/. You are also missing an `aria-selected` attribute here.

lespacedunmatin: Or a list of buttons (`ul/li`).

lespacedunmatin:
See https://a11y.nicolas-hoffmann.net/tabs/ for keyboards controls.

react/Tabs/index.jsx
goldoraf: they don't need to be buttons as long as they have the `role=tab` attribute. But yes, ARIA is missing here.

ptbrowne: Can this be treated in a separate issue as it is an implementation issue and not an API issue ?

enguerran: @lespacedunmatin you are so right and I am so not sensitized to a11y. I am ashamed and want to improve that skill. I will update this PR to add aria.

ptbrowne:
> The solution with a set of tab title and a set of tab panel is a risk to let source of truth be duplicated and has a consequence of a desynchronization risk.

I see what you mean but I do not think it would be a real problem. If you code manually your tabs, you will see directly that you have 2 tabs and 3 panels (and the Tabs components could warn you if you do that). If you are programmatically implementing tabs, I guess the Panels and Tabs would be generated mapping over the same array.

> Adding the support of badges could simply be a matter of adding a icon prop.

The React has few APIs and they work well together. IMHO, the `children` prop is a very good thing as it enables composability. Adding an `icon` prop is adding complexity, a new prop to know for the developer.

In terms of simplicity, having `Tabs`, `Tab` and `title` is of the same complexity as `Tabs`, `Tab`, `Panel`. I prefer the one where we use `children` as it enables you to put what you wants in there.

react/Tabs/index.jsx
ptbrowne: Could you update the existing `Tab` component with aria also ?

enguerran:
@lespacedunmatin would you give me your opinion on the last commit adding aria attributes? https://github.com/cozy/cozy-ui/pull/144/commits/93892b0d3e136abfadd4138fb537ad7deb6b7380

goldoraf:
FYI: https://medium.com/@RubenOostinga/avoiding-deeply-nested-component-trees-973edb632991

clochix:
Awesome! Years later, you're re-inventing XUL! https://developer.mozilla.org/en-US/docs/Mozilla/Tech/XUL/Tutorial/Tabboxes :-P

enguerran:
This is some kind of HOC btw https://medium.com/@RubenOostinga/avoiding-deeply-nested-component-trees-973edb632991#f224

enguerran:
I need to check once more the code, because I get some weird behavior with this component within preact:
1. when clicking on the tab title, the tab content did not change (the component is not re-render) but it is marked as `_dirty: true`...
1. when I click on another listener HTMLElement, the render is launched and the tabs component is repaint to display the content I click the title earlier...

More on that asap.

PR #148, opened 2017-06-21
Add Spinner component

GoOz:
Added a spinner component because it's used everywhere but never in a same way.
Available options so far:
- colors (default blue)
- size (default 1em (16px))
- noMargin — removes the default left and right margins
- middle — center the whole component both vertically & horizontally like it's used when loading a content view
- loadingType — if you want a message below the Spinner (to be consistent with locales files obviously)

### DIFF
--- a/react/Spinner/index.jsx
+++ b/react/Spinner/index.jsx
@@ -4,13 +4,6 @@ import classNames from 'classnames'
 
 import styles from './styles'
 
-// === Component Options ===
-// loadingType = message type from locale [string] (default: '')
-// middle      = Component is centered both vertically & horizontally [bool] (default: false)
-// noMargin    = Component has no margin [bool] (default: false)
-// color       = Color of the spinner (available colors: blue, grey, white, red) [string] (default: 'blue')
-// size        = Size of the spinner (available sizes: tiny, small, medium, large, xlarge, xxlarge) [string] (default: 'medium')
-
 export const Spinner = ({ t, loadingType, middle, noMargin, color, size }) => {
   return (
     <div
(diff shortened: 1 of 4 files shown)


GoOz:
[(0,5j) Etudier la mutualisation du spinner sur Drive en un composant extensible](https://trello.com/c/nFMXnSzS/373-05j-etudier-la-mutualisation-du-spinner-sur-drive-en-un-composant-extensible)

react/Spinner/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+import styles from './styles'
+
+// === Component Options ===
+// loadingType = message type from locale [string] (default: '')
+// middle      = Component is centered both vertically & horizontally [bool] (default: false)
+// noMargin    = Component has no margin [bool] (default: false)
+// color       = Color of the spinner (available colors: blue, grey, white, red) [string] (default: 'blue')
+// size        = Size of the spinner (available sizes: tiny, small, medium, large, xlarge, xxlarge) [string] (default: 'medium')
+
```
enguerran: I am not a big fan of comments in code. Let's imagine another developer operates on the code and change something, but does not update the corresponding comments. Then, later on, another developer operates on the code and see something strange:
> the comment says you set a title with `title` but it does not work. And when he looks closer in the code, he sees that to set a title you need to use a `name` attribute. What's going on, there? What a mess!

That is why React offers a propTypes attributes to let developers or typechecker knows what the code is waiting for. See https://facebook.github.io/react/docs/typechecking-with-proptypes.html#proptypes for more information.

PR #151, opened 2017-06-23
fix: alert position on mobile

GoOz:
wrong em calculation


GoOz:
Well in this case if we change the font-size of the page it will grow or shrink accordingly which wouldn't happen if we used pixels. I personnaly prefer "rem" but hey… it's historically em :/

y-lohse:
> it will grow or shrink accordingly

Ah right. In my mind the bar was sized in pixels, but actually it's using em's too. That makes sense then, I have to be careful about that (I probably used px already in some places)

(And team rem for the win)

PR #152, opened 2017-06-27
Feat/better modal n tabs

GoOz:
- [x] Added a new modal content component ModalContentOffset  with no margin if needed 
- [x] Changed some tabs units from `px` to `rem`


react/Modal/index.jsx
```diff
@@ -8,6 +8,11 @@ const ModalContent = ({children, className}) =>
     {children}
   </div>)
 
+const ModalContentOffset = ({children, className}) =>
```
y-lohse: I'm very confused about the different types of modal contents, which ones have padding, which ones don't, and when to use what. Nothing wrong with the code, just a comment.

stylus/ui-components/tabs.styl
```diff
@@ -6,7 +6,7 @@ $tabs-base
 
     .coz-tab
         display        inline-block
-        padding        10px 15px
+        padding        .625rem 1rem
```
y-lohse: Are we using rem's now? Just curious because we talked about it the other day.

GoOz: I maybe* get rid of `em` in every other context but font-size and use `rem` instead in the near future. `em` is way too much a pain in the ass even for me and I don't want to make things complicated for everyone.

* that means i strongly want to do that.

react/Modal/index.jsx
GoOz: ModalContent is the basic one with paddings, used in most cases.
ModalContentOffset doesn't have paddings so we can do anything inside (like tabs that take the all with of the modal).

GoOz:
@gregorylegarec that's what I tried first but realised that in order to have the consistent margins between items (title, content, close) I have to rely on the less different item possible. That's why the Offset Content stays a Content (that means that it inherits everything of Content item which it needs) and then just its padding is removed. 
I don't know if I'm clear on this. :/

GoOz:
@gregorylegarec @y-lohse Nevermind, you were right, I missed something. So I rolled back and removed `Content`, is that ok with both of you?

gregorylegarec:
Totally !

PR #156, opened 2017-08-04
feat: Add Table React Component :desert_island:

kosssi:
Voici le composant React `Table`.

Le but étant d'avoir le moins de CSS à faire coté application React.

Voici un exemple d'utilisation :

```javascript
import React from 'react'
import Table from 'ducks/Table'
import glamorous from 'glamorous'

const AccountNameColumn = glamorous.td({ flexBasis: '20%!important' })
const SoldeColumn = glamorous.td({ flexBasis: '20%!important' })
const BankNameColumn = glamorous.td({ flexBasis: '30%!important' })
const AccountNummberColumn = glamorous.td({ flexBasis: '30%!important' })

const Balance = () => (
  <div>
    <h2>Solde actuel</h2>
    <Table>
      <thead>
        <tr>
          <AccountNameColumn>Nom du compte</AccountNameColumn>
          <SoldeColumn>Solde</SoldeColumn>
          <BankNameColumn>Nom de la banque</BankNameColumn>
          <AccountNummberColumn>Numéro du compte</AccountNummberColumn>
        </tr>
      </thead>
      <tbody>
        <tr>
          <AccountNameColumn>BNP - CCHQ (Michel)</AccountNameColumn>
          <SoldeColumn>1 245 234 245 234 245 234 €</SoldeColumn>
          <BankNameColumn>BNP Paribas</BankNameColumn>
          <AccountNummberColumn>47T6178849134</AccountNummberColumn>
        </tr>
      </tbody>
    </Table>
  </div>
)

export default Balance
```

TODO:

- [ ] Ajouter un README.md pour le composant
- [ ] Mettre à jour le projet cozy-drive
- [ ] Mettre à jour le projet cozy-settings
- [ ] Mettre à jour le projet cozy-bank
- [ ] Mettre à jour le projet cozy-sante


GoOz:
Wait what why?
I don't get what's your point here?
Make a better table component? sure i'm up for it, but we're not using HTML `<table>` anymore due to a too restrictive behaviour so I really don't get where you're going. 😕

ptbrowne:
> we're not using HTML <table>

Maybe cozy-drive and cozy-photos are not using it but cozy-bank and cozy-sante are using it.

GoOz:
> What do you mean ?
You can't do whatever you wanna do with tables. Not as much as simple divs can do.
The first time I made this kind of view I used tables but some requests from the designer made the use of table just impossible.

The component shouldn't use table markup only, as settings and drive don't use table for example, if this component is to be used, it must be 'markup agnostic' in some way, or maybe should use params, i don't know. But not table only.

ptbrowne:
Trying to make this issue go forward.

> The component shouldn't use table markup only, as settings and drive don't use table for example, if this component is to be used, it must be 'markup agnostic' in some way, or maybe should use params, i don't know. But not table only.

Could you provide examples of what are your needs and what problems you encountered on Drive/Photos ?

I would accept <Table /> in cozy-ui for the moment being since it has already proven useful in Bank and Sante. @kosssi can you provide examples in the storybook and add a screenshot here ?

The fact that Drive and Photos want to use a different tag should not be a blocker. @GoOz, I think it will be easy to add a RowRenderer prop and a CellRenderer prop but since we do not have the need, I am not sure if the implementation will survive the test of usage. I would prefer if you'd make the PR to add this functionality as you are more aware of your specific needs in Drive/Photos.

GoOz:
Well, it was an issue at some point, there were supposed to be animations when loading content of table, animations that couldn't be done with table. Yet, due to some complications, animations never came as you may know. But I don't know if it's off the table.

Nevertheless, tables are far less flexible than div by nature and they do not, in our use cases, add much more semantic. So I really wouldn't recommend using tables at all. It's way more future proof to use divs. The only way to bypass some (not all) of the default behaviours of tables is to add divs inside divs and that's not a solution, that's more of a trick. I'd rather lose some (not so much) semantic than create a div soup inside <td>s.

As for implementing the renderers, I'm not sure I can do that. But I can tell you the needs for Drive, it's just "no table".

So I don't really know what to say, or do. :/

y-lohse:
> But I don't know if it's off the table.
🤣 

Ok seriously though, this PR and the needs for Drive are different things, aren't they? 

I know we had issues (was it on Safari Mobile?) with `tables` in Drive and even though I actually think they'd be good semantically, I'm ok with using `divs` instead there.

But this is cozy-ui. Obviously `tables` work in some cases since they work in cozy-bank. If cozy-ui offers a `Table` component, I feel it should be based on `table` tags, unless we have a very good reason not to use them.

It might be obvious, but we wouldn't have to use this component in Drive in which case it would have zero effects.

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+            text-transform  uppercase
+
+        tr:hover
+            background-color white
+
+    tbody
+        flex      1 1 auto
+        display   flex
+        flex-direction column
+        // Because it needs a height to be scrollable. Don't ask.
```
y-lohse: I'd actually like to know why we need a `1px` height here. If someone knows it'd be great to have extra details here :)

y-lohse: I'd actually like to know why we need a `1px` height here. If someone knows it'd be great to have extra details here :)

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
+@import 'cozy-ui'
+
+.coz-table
+    position        relative
+    display         flex
+    flex-direction  column
+    flex            1 1 100%
+    height          100%
+    text-align      left
+    color           grey-11
```
y-lohse: Can we replace the `grey-XX` with the color names?

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+
+    thead
+        flex 0 0 em(31px, 16px)
```
y-lohse: I think we should use an `rem`value here.

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+
+        td
+            padding         em(8px, 12px) em(32px, 12px) // @stylint ignore
```
y-lohse: I think we should use more obvious values here, probably in `rem`.

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+            font-size       12px
```
y-lohse: Probably more `rem` or `em`!

react/Table/styles.styl
```diff
@@ -0,0 +1,57 @@
...
+
+    td
+        box-sizing  border-box
+        padding     em(15px, 14px) em(32px, 14px) // @stylint ignore
+        font-size   14px
+        flex 1 0 auto
+
+    @media (max-width: (1024/basefont)rem)
+        tr
+            max-width  100vw
```
y-lohse: Is there an advantage of using `100vw` over `100%` here? I've had a few issues with `vw` and find them a bit unreliable. And since it's a component I find it very arbitrary to say it will take the full viewport. Maybe there will be something next to the `table`, even on mobile.

GoOz:
@y-lohse Yeah, you're right, that was my original point actually. If it's tables only then we won't be using it ourselves on half of our apps. That's mainly why I thought it was sad to do this, but hey, I'm not gonna make Bank & Santé's markup change and we're not gonna change Drive & Photos' markup so let do it like this. It's not that important.

> > But I don't know if it's off the table.

> 🤣

Glad you saw this one 😆

About the changes you asked… this is a old PR, that explains why, but you're right, be careful with the rebase.

react/Table/styles.styl
ptbrowne: rem here :)

react/Table/styles.styl
GoOz: The comment kinda "explains" why. Some webkit based browsers need a defined height on flex item to be scrollable. More a bug than a feature actually. I didn't go into details because I didn't fully get the why, but here's what I know about it.

react/Table/styles.styl
GoOz: nope. 100% is the best option I guess.

ptbrowne:
Should we close this PR ? It has been inactive since December.

GoOz:
Dunno. There's something to be done for sure but I'm still not convinced by this way of doing.

PR #160, opened 2017-08-24
fix: remove overflow hidden on modal

GoOz:
@ptbrowne I tried to find another solution but I couldn't… :/
`overflow: hidden` is too restrictive for every use case. We'll have to find something else for PDF display.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -88,14 +88,14 @@ class Modal extends Component {
   }
 
   render () {
-    const { children, title, withCross } = this.props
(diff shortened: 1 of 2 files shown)


ptbrowne:
Can we add a `className` option to the modal so that an application can apply the `overflow: hidden` class ?

Then the only change required on my side would be to add `className={ 'hiddenOverflow' }` to https://gitlab.cozycloud.cc/labs/cozy-bank/blob/master/src/components/FullscreenIntentModal.jsx

GoOz:
Sure. No Problem. Give me 5min 😄

Issue #161, opened 2017-08-28
Overflow in iOS


ptbrowne:
When you scroll from the bar at the bottom, you can scroll the html container and then the layout is broken.

![bug-ios-scroll](https://user-images.githubusercontent.com/465582/29771149-9a88aca0-8bf2-11e7-95b1-2dbaa30db14b.gif)

y-lohse:
Thanks, @GoOz is actually working on this atm :)

ptbrowne:
I reopen because the layout no longer stays broken thanks to https://github.com/cozy/cozy-ui/pull/162 but you still have the elastic scroll behavior that is weird.

@GoOz Can you confirm that you are aware of the problem ? Are you working on this at the moment ?

GoOz:
@ptbrowne Absolutely not. :(
What do you mean by elastic scroll?

ptbrowne:
![bug-ios](https://user-images.githubusercontent.com/465582/30063531-54cc613a-924f-11e7-9929-19f4dc7c8233.gif)

ptbrowne:
Est ce que t'as une idée pour ça ?

GoOz:
Nope. 😞 
Apparently,
```css
html,
body {
    width: 100%;
    height: 100%;
    overflow: hidden;
}
```
should fix the elastic scroll but since those two elements already have those properties, I guess it doesn't. I have no clue…

GoOz:
The easier way would be to redo the layout but it's still not considered at this point.

GoOz:
New layout for mobile is merged. 🎉

PR #162, opened 2017-08-28
fix: overflow error on html

ptbrowne:
Fixing #161 

After : 

bug-ios-scroll-fixed

The Drive text staying fixed as we scroll is another bug.


GoOz:
I'm keeping this aside for a moment. I'm trying to figure out if a refactoring of the mobile layout is possible in some way.
If I succeed, I won't be needed that.
If I fail, I will.
So far, the odds are not very good with me, but we'll see.

PR #165, opened 2017-08-30
feat: Add lang on translate :globe_with_meridians:

kosssi:
You can get `lang` like this:

```jsx
class MyLang extends Component {
  render ({lang}) {
    return <div>{lang}</div>
  }
}

export default translate()(MyLang)
```


ptbrowne:
It is useful for localization of numbers: 

```js
(123).toLocaleString(lang)
```

PR #170, opened 2017-09-06
doc: add toggle examples

enguerran:
I had examples of how to use `<Toggle />` and the `@storybook/addon-info` to display source code and React.PropTypes... To become a beautiful components' styleguide.

peek 2017-09-06 18-21

See the demo here: https://enguerran.github.io/cozy-ui/react/

### DIFF
--- a/docs/react/.storybook/addons.js
+++ b/docs/react/.storybook/addons.js
@@ -1,0 +1,1 @@
+import '@storybook/addon-actions/register';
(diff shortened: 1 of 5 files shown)


enguerran:
I know for `<ControlledToggle defaultChecked={false} />`. As you can see in the story, this example use a wrapper component to control the `<Toggle />` component. Though Storybook addon `@storybook/addon-info` write down the content of the `return` of the story in a *shallow* mode.

So unless we find either:

- another way to control the component
- a way to update addon-info
- another plugin

I guess we need to go straight into the source code to understand what's happening in this very example.

PR #174, opened 2017-09-07
fix: make default coz-btn more consistent with other button styles

GoOz:
Made the default `.coz-btn` class more consistent with the other button variants so now we can use it on its own.
Basically it's a transparent button with white border & text, and I added behavior for some pseudo-classes (It was already transparent with white borders, I just added the text color).

I'm pretty sure none of all the Drive & Photos' buttons use only the default class but @y-lohse if you could take a quick look in case I missed something, that would be nice of you.

Same for @CPatchane & @ptbrowne, I (almost) never worked on platformers or gangstas projects, could you assure me that this PR won't break anything on your side?

### DIFF
--- a/stylus/ui-components/button.styl
+++ b/stylus/ui-components/button.styl
@@ -21,6 +21,7 @@ $button
     background       transparent
     vertical-align   top
     text-align       center
+    color            white
     font-size        14px /* [1] */
     line-height      1
     text-transform   uppercase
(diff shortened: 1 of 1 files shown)


y-lohse:
I'm about to take a look at it, but why these changes? White on transparent is not going to be readable on most ouf our apps, and changing to white on black on active/focus/hover seems very radical? Aren't you worried that some other button variation accidentally inherits one of these?

y-lohse:
Aw, I'm afraid I was right. Here's the upload button on photos in it's hover state 
capture d ecran 2017-09-07 17 45 44

Not sure this is such a good idea :/

GoOz:
My bad, I may have assumed things too quickly and changed color on `:hover`when I shouldn't.
I changed the way to do it.

To explain my decision, I had a need for a transparent button with white borders and text and transparent background to put in background colored parent, which is exactly the never used `.coz-btn` does except for the text color.
I figured it would be smarter to use this default class like this.

GoOz:
[(2) [FIL] Améliorer la visibilité du téléchargement du client desktop](https://trello.com/c/EgioKRS2/496-2-fil-am%C3%A9liorer-la-visibilit%C3%A9-du-t%C3%A9l%C3%A9chargement-du-client-desktop)

PR #180, opened 2017-09-13
chore: use styleguidist to generate react styleguide

enguerran:
We want to go with styleguidist which give us more control on our styleguide.

Check it out live on https://enguerran.github.io/cozy-ui/react/

Tell me what yo think and what you want, we'll do the best to fit your needs.


docs/styleguide.config.js
```diff
@@ -0,0 +1,8 @@
+const path = require('path')
+
+module.exports = {
+  components: path.resolve(__dirname, '../react/{Button,Toggle}/*.{js,jsx}'),
```
enguerran: You'll have to understand that this configuration entry is going to change as soon as somebody write `README.md` files for the other components.

react/Button/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import cx from 'classnames'
 
-export default function (props) {
+export default (props) => {
```
enguerran: I move the code into a specific folder to include `README.md` file (that I did not write).
It may not change anything in your third-party project because you may `import {Button} from 'cozy-ui/react'`

react/Toggle/README.md
```diff
@@ -0,0 +1,40 @@
...
+    }}
+  >{`Toggle is controlled and ${state.checked
+    ? 'on'
+    : 'off'}`}</label>
+  <Toggle
+    id='0'
+    checked={state.checked}
+    onToggle={() => setState({ checked: !state.checked })}
+  />
+</div>
```
enguerran: I think it's beautiful, more on this on the official styleguidist repository: 

- https://react-styleguidist.js.org/docs/getting-started.html
- https://react-styleguidist.js.org/examples/basic/

enguerran:
`yarn.lock` is always very noisy when we change dependencies. I hope it is good.

docs/styleguide.config.js
ptbrowne: I think we could just put '../react/**/*.jsx' and [`skipComponentsWithoutExample`](https://react-styleguidist.js.org/docs/configuration.html#skipcomponentswithoutexample)

ptbrowne:
Awesome, thanks !

docs/styleguide.config.js
```diff
@@ -0,0 +1,8 @@
...
+  webpackConfig: require('./webpack.config.js'),
+  serverPort: 6161,
+  styleguideDir: path.resolve(__dirname, '../build/react')
```
ptbrowne: I would add `showUsage: true` to show props and methods by default

image

docs/webpack.config.js
```diff
@@ -0,0 +1,37 @@
+const ExtractTextPlugin = require("extract-text-webpack-plugin");
```
ptbrowne: Why not use the already existing `webpack.config.js` ? That's what I did for Bank and it is working well. This way we have only one source of truth ;)

react/Button/index.jsx
ptbrowne: For info, another way to go is to write the documentation in `Button.md`. This way you do not have to create a folder.

docs/styleguide.config.js
enguerran: We may, but I did not seek to far. As said, we should adapt such a configuration with the other component's integration. The main part is here.

react/Button/index.jsx
enguerran: Oh, right, but if every components are in folders, maybe it is a good think to follow this habit.

docs/styleguide.config.js
enguerran: as you wish, no pb

docs/webpack.config.js
enguerran: There is no other `webpack.config.js`, is there?

ptbrowne: There is not, indeed... My bad :)

enguerran:
As PO, I though you will be interested by the styleguide we can produce for the community.

PR #183, opened 2017-09-14
feat: new color & bigger icons on mobile

GoOz:
Nav icons are now bigger on mobile and default text color is a little bit darker.

NB: From now on, the nav style in apps only need to specify the background-image (the icon path actually) on the `.coz-nav-link` element. The remaining properties are handled by Cozy-UI.

### DIFF
--- a/stylus/ui-app/nav.styl
+++ b/stylus/ui-app/nav.styl
@@ -31,17 +31,19 @@ $navigation
             background     silver
 
     .coz-nav-link
-        display          flex
-        left-padded      50px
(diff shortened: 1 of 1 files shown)


GoOz:
[(0) [GLOB] Améliorer la visibilité du menu bottom de l'app mobile](https://trello.com/c/xp0aw4pS/512-0-glob-am%C3%A9liorer-la-visibilit%C3%A9-du-menu-bottom-de-lapp-mobile)

PR #190, opened 2017-09-20
Add Markdown component with default options :sparkles:

CPatchane:


ptbrowne:
Can you add documentation on what purpose serve the watch:react and build:react scripts ?

ptbrowne:
Can you rename the PR and describe what you've done here ? It seems to me that the core of the PR is not about the `<markdown>` component ;)

nono:
@CPatchane I'm not sure to see what is the value of this wrapper over the original react-markdown

CPatchane:
@nono Hum the value of using cozy-ui here is having the same renderers options for the component. The component itself is not very different but I used to make a wrapper component in all my projects to use the same renderers. Maybe it would be better to just export a renderer options for this components and not the component itself. I'll think about that.

y-lohse:
I've also mentioned that I'm not sure about the wrapper being that good. I feel like the default behavior shouldn't be to open a link in a new tab..? If that was react-markdown's default behavior I would find it very unexpected.

ptbrowne:
> Hum the value of using cozy-ui here is having the same renderers options for the component.

What do you mean by renderer in this context ?

nono:
@CPatchane I think we should close this PR. The value is not clear and it has conflicts. Are you OK with that?

CPatchane:
Completely 👍

PR #192, opened 2017-09-25
Big clean up !

GoOz:
Don't really expect a thorough review but really keeping you posted instead to what has been done so far. So here it is:
- [x] Removed mixins `padded` and `spaced`
- [x] Added buttons global classes
- [x] Ensure retro-compatibility
- [x] Removed Lato font from UI & moved to the stack (you'll need to add a `<link>` tag to the index.ejs of your app `<link rel="stylesheet" type="text/css" href="//{{.Domain}}/assets/fonts/fonts.css">`
- [x] Fix modal position on desktop
- [x] Added Nav component
- [x] Clean up comments


stylus/components/alerts.styl
```diff
@@ -19,7 +19,7 @@ $alerts
         box-sizing        border-box
         box-shadow        0 em(6px) em(18px) 0 rgba(50, 54, 63, .23)
         background-color  emerald
-        padded            .8em, 1em
+        padding            .8em, 1em
```
nono: Alignment for the right column is broken

stylus/components/button.styl
```diff
@@ -46,7 +67,8 @@ $button
             top        em(-1px)
 
 
-$button--regular
+$button--regular // depreciated
```
nono: s/depreciated/deprecated/

stylus/components/alerts.styl
GoOz: Good catch :)

stylus/components/button.styl
GoOz: Oh man, it sounded like depreciated in my head so I wrote it that way, thanks.

PR #193, opened 2017-09-26
fix: nav layout on mobile with several items

GoOz:
Nav items had a way too big min-width so I removed it along with useless padding and used flex-basis instead. Now you can have 5 items easily.


GoOz:
About the padding, it could be an issue if we have like 6 or more items which shouldn't be the case. But we'll see if it happens later.

PR #194, opened 2017-09-26
[WIP] Introduce built cozy-ui components

CPatchane:
Issues about using built components still to resolve: 

* Could find an alternative way to [`components/entry.js`](https://github.com/cozy/cozy-ui/pull/194/files#diff-02d27d1862145d9144c56c216abbe410). This way will break all jest tests due to `.css` and `.svg` files `require` statements.
* Find a way to merge 2 sprites generated by `svg-sprite-loader` (the one from cozy-ui to merge with the one from the application) or maybe find another way to build all svg files.

These issues could be solved if there is a way to include/require a file from cozy-ui directly by webpack (to include the components css code and to include all svg), but I didn't find a working way to do that for now.

So I keep that here until I have more time to investigate that.
Feel free to pull, test, challenge or make comments.


ptbrowne:
@CPatchane I have trouble understanding what this PR does and what will be the benefits :/ Are you still working on it ?

CPatchane:
@ptbrowne This PR was a kind of an exploration and is clearly blocked and not complete. I don't think build each components will be a good choice in any cases.
For me, we should keep the cozy-ui components being used inside the application webpack build to handle all necessary files (jsx and files) and get rid of the CSS modules to use styles from cozy-ui or a separate css file (which will be also loaded in the application webpack build).

PR #196, opened 2017-09-27
Document <Spinner /> <Button /> <Icon />

ptbrowne:
Visible on https://ptbrowne.github.io/cozy-ui/react/

Edited: changed the url to show the result on my repo.


enguerran:
I cannot see anything on the link you give

enguerran:
You create a branch on the cozy repository, is that wanted? Why not a fork?
https://github.com/cozy/cozy-ui/tree/doc-spinner-buttons-icons

docs/styleguide.config.js
```diff
@@ -1,8 +1,11 @@
 const path = require('path')
 
 module.exports = {
-  components: path.resolve(__dirname, '../react/{Button,Toggle}/*.{js,jsx}'),
+  components: path.resolve(__dirname, '../react/{Spinner,Icon,Button,Toggle}/*.{js,jsx}'),
```
enguerran: We shall not list components here, if we have 30 000 components, we could not list them anymore, could we?

docs/webpack.config.js
```diff
@@ -33,5 +34,14 @@ module.exports = {
       }
     ]
   },
-  plugins: [new ExtractTextPlugin('[name].css')]
+  plugins: [
+    new webpack.LoaderOptionsPlugin({
```
enguerran: What is the utility to use [LoaderOptionsPlugin](https://webpack.js.org/plugins/loader-options-plugin/) if we can set an option in [the loader itself](https://github.com/shama/stylus-loader/issues/178#issuecomment-331740327)?

docs/styleguide.config.js
```diff
@@ -1,8 +1,11 @@
...
   webpackConfig: require('./webpack.config.js'),
   serverPort: 6161,
-  styleguideDir: path.resolve(__dirname, '../build/react')
+  styleguideDir: path.resolve(__dirname, '../build/react'),
+  require: [
```
enguerran: Cannot fund any documentation about `require` in webpack.
What is used for?

react/Icon/index.jsx
```diff
@@ -1,8 +1,10 @@
-import React from 'react'
+import React, { PropTypes as Types } from 'react'
```
enguerran: https://facebook.github.io/react/docs/typechecking-with-proptypes.html
> React.PropTypes has moved into a different package since React v15.5. Please use the prop-types library instead.

Unless it's added with `preact-compat`. But I think to be more reliable, we should depend on the real [proptypes module](https://github.com/facebook/prop-types).

On the other hand, I cannot understand why you alias the import.

ptbrowne: > On the other hand, I cannot understand why you alias the import.

Oh I'm sure you can :D : it is to type les characters. It is fairly common : https://github.com/search?utf8=%E2%9C%93&q=%22PropTypes+as+Types%22&type=Code

docs/styleguide.config.js
ptbrowne: To require additional files : https://react-styleguidist.js.org/docs/configuration.html#require

docs/styleguide.config.js
ptbrowne: Exactly what I was talking about here : https://github.com/cozy/cozy-ui/pull/180#pullrequestreview-62379744

I am doing the change

ptbrowne: done

docs/webpack.config.js
ptbrowne: Thanks, good catch, done !

react/Icon/index.jsx
ptbrowne: What do you mean by to be more reliable ?

ptbrowne: `preact-compat` is using the `prop-types` module under the hood 👌 https://github.com/developit/preact-compat/blob/master/package.json#L87

enguerran: We are using `preact` but to stay more reliable on `react`, we also use `preact-compat`.
For instance `preact` let us declare the `render` function with `props` as parameters, `react` doesn't. So we usen't.

The same goes to `PropTypes`, since `react` 15.5, `PropTypes` has moved to a different package.

If you look at the `import` part of both below codesandbox, you'll understand:

- preact with preact-compat works with `import { PropTypes } from "preact";` (https://codesandbox.io/s/2okk0663rp)
- react don't (https://codesandbox.io/s/kmj356yp15)

On the other hand, both works with `import PropTypes from "prop-types"`.

In any event, both used the exact same release of `prop-types`.

react/Icon/Readme.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+const icons = require('../../src/icons');
+const colors = ['red', 'blue', 'yellow', 'green']
+let i = 0;
+
+<div style={{ background: '#ccc'}}>{Object.keys(icons).map(icon => <span>
+  { icon } : <Icon icon={ icon } color={ colors[i++] } /><br/>
+</span>) }</div>
+```
```
enguerran: Two problems here:

enguerran:
```jsx
const icons = require('../../src/icons');
const colors = ['red', 'blue', 'yellow', 'green']
let i = 0;

<div style={{ background: '#ccc'}}>
  {Object.keys(icons).map(
    icon => (
      <span>{ icon } : <Icon icon={ icon } color={ colors[i++%colors.length] } /><br/></span>
    )
  )}
</div>
```

See https://codesandbox.io/s/7y0oz2m3v0

enguerran: Beware of the code sample, it lacks a closing parenthesis in the map function.
And format the code as it could be more readable.

react/Icon/Readme.md
```diff
@@ -0,0 +1,31 @@
...
+You can also directly import an SVG to use it. You MUST use svg-sprite-loader
+to load your SVG (either explicitly or, better, implicitly in your `webpack.config.js`).
+
+⚠️ Do not put a `fill` property on your icon
+
+```jsx static
+
+import myIcon from 'my-icon.svg' 
+
+<Icon icon={ myIcon } width='2rem' height='2rem' color='purple' />
```
enguerran: It ain't no purple.

react/Icon/index.jsx
```diff
@@ -35,8 +37,9 @@ const Icon = function ({ icon, width, height, style, color, className, preserveC
   </svg>
 }
 
+
 Icon.propTypes = {
-  icon: React.PropTypes.string.isRequired
+  icon: Types.oneOfType([Types.string, Types.object]).isRequired
 }
```
enguerran: Add other props even if they are not required.
See https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prop-types.md

react/Spinner/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+### Color
+
+```
+<div>
+  blue: <Spinner color='blue' />
+  grey: <Spinner color='grey' />
+  white: <Spinner color='white' />
+  red: <Spinner color='red' />
+</div>
+```
```
enguerran: I cannot see colored spinner, look at the screenshot below:

react/Spinner/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Placement
+
+```
+<div>
+<Spinner noMargin={ true }/>
+</div>
+```
```
enguerran: I see no difference with `noMargin={true}`, see below:

react/Spinner/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+<div>
+  tiny: <Spinner size='tiny' /><br/>
+  small: <Spinner size='small' /><br/>
+  medium: <Spinner size='medium' /><br/>
+  large: <Spinner size='large' /><br/>
+  xlarge: <Spinner size='xlarge' /><br/>
+  xxlarge: <Spinner size='xxlarge' />
+</div>
+```
```
enguerran: `tiny`, `small` and `medium` are the exact same size, see:

react/Spinner/Readme.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+### Message
+
+When you use `loadingType`, `<Spinner />` needs to be in an `<I18n />` wrapper as it uses `t` function.
+
+
+```jsx static
+<Spinner loadingType='salut' />
+```
```
enguerran: I don't want `cozy-ui` to be dependent on `<I18n />`.
But maybe, I am alone on this idea.

So please, consider this comment as a non blocking one, we should update later.

By the way, could you find a way to perform a running demo, maybe you can pass through a props like `t={word => word}`.

react/Spinner/Readme.md
ptbrowne: It is a bug from the Spinner

react/Spinner/Readme.md
ptbrowne: Bug from the spinner

react/Spinner/Readme.md
ptbrowne: Bug from the spinner

react/Icon/Readme.md
ptbrowne: good catch !

ptbrowne: no parenthesis lacking here

react/Icon/Readme.md
ptbrowne: What do you mean it ain't no purple ? Are you testing with an icon with a fill property ?

react/Icon/Readme.md
ptbrowne: > paperplane icon missing

Bug from the icon, not from the docs

> only the first 4 icons

Thanks, good catch ><

react/Icon/index.jsx
ptbrowne: Thanks i added the properties. I presume you use this eslint configuration in your editor ? The lint command of cozy-ui only uses standard js, maybe you can add eslint ?

react/Spinner/Readme.md
ptbrowne: IMHO, this is not a problem of the docs.

react/Icon/index.jsx
ptbrowne: done

react/Spinner/Readme.md
ptbrowne: https://github.com/cozy/cozy-ui/issues/197

react/Spinner/Readme.md
ptbrowne: https://github.com/cozy/cozy-ui/issues/197

react/Spinner/Readme.md
ptbrowne: https://github.com/cozy/cozy-ui/issues/197

react/Icon/Readme.md
enguerran: Oh, I though the icon was the ⚠️
I did not read the doc that explains we need to use `svg-sprite-loader`.

react/Spinner/Readme.md
enguerran: It allows to display a running demo.

enguerran:
Even if we can improve the demo by fixing `<Spinner />` and by adding more running demos, I see no other blocking things for this Pull-Request.

ptbrowne:
Thanks for the thorough review !

PR #199, opened 2017-10-07
PropTypes and active links fixes

CPatchane:


stylus/components/nav.styl
```diff
@@ -68,13 +68,11 @@
...
 
-    .active // deprecated
-    .is-active
     &:hover
         color  charcoal-grey
 
-    .active // deprecated
-    .is-active
+    &.active // deprecated
+    &.is-active
```
GoOz: Did you experience any issue to put those `&` back? I removed them on purpose to reduce the weight of selectors.

ptbrowne: OK for the fix. For the record, we should use BEM here.

GoOz:
@ptbrowne I asked a question to @CPatchane that wasn't answered… I'd love to have an answer before merging, you know.

ptbrowne:
Sorry @GoOz , I should not have merged, you are right...

stylus/components/nav.styl
CPatchane: @GoOz Yep, there wasn't any `active` styles anymore on my side

GoOz: Oh ok. I didn't get this bug on Drive but i suppose it's done differently on your app, that's why. Ok to re-add it then. For now at least. :)

GoOz:
@ptbrowne no biggie. :)

stylus/components/nav.styl
CPatchane: Drive seems to use Link component differently indeed, keep me noticed if you have some issues about this point so.

PR #200, opened 2017-10-09
Make date-fns a dependency instead of a devDependency

CPatchane:
Since it's necessary for I18n component usage, it's could be better to have it directly in cozy-ui.
It avoid to add it in the app without directly using it (confusing).


GoOz:
I guess it makes sense.

CPatchane:
Same thing for stylus

PR #201, opened 2017-10-09
fix: import file only once

enguerran:
see http://stylus-lang.com/docs/import.html\#require
for more details about stylus @import vs @require


enguerran:
From a `3.6 Mo` file built in 2m 48s to a `1.4 Mo` file built in 1m 5s

PR #202, opened 2017-10-12
Fix/v4 improvments

GoOz:
Massive clean up here.

I removed all the `:global` intructions from UI to make it agnostic, as much as possible at least. 
You can use or not CSS-modules on your app but you'll have to `@extend` the  previous global classes, such as `.coz-btn`, and make them global yourself in you app if you want to. 

Besides that there's 2 micro bugfix about buttons and the nav on mobile.

### DIFF
--- a/react/Alerter/index.jsx
+++ b/react/Alerter/index.jsx
@@ -132,7 +132,7 @@ class Alert extends Component {
       >
         <p>{message}</p>
         {buttonText &&
-          <button onClick={buttonAction} className={classNames('coz-btn', `coz-btn--alert-${type}`)}>
+          <button onClick={buttonAction} className={classNames(styles['coz-btn'], styles[`coz-btn--alert-${type}`])}>
             {buttonText}
           </button>
         }
--- a/react/Alerter/styles.styl
+++ b/react/Alerter/styles.styl
@@ -3,4 +3,17 @@
 @require '../../stylus/components/alerts'
 
 .coz-alerter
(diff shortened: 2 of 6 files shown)


ptbrowne:
> I removed all the :global intructions from UI to make it agnostic, as much as possible at least.
You can use or not CSS-modules on your app but you'll have to @extend the previous global classes, such as .coz-btn, and make them global yourself in you app if you want to.

I do not understand, I thought that we were getting rid of placeholders because we did not want the app builders to have to write too much css. It seems we are going in the exact opposite direction here. Am I missing something ?

ptbrowne:
Can we split this PR in 2 ? It seems that there are two things going on here, the fixing of nav and buttons and the refactor for the global classes.

ptbrowne:
OK so I was confused. Since there are no more global classes, if you use css modules when you load cozy-ui AND you want to use cozy-ui classes outside cozy-ui components, you will have to use `@extend`. Another alternative is to load your app styles with css modules, without loading cozy-ui with css modules. This means you will be able to do `<MyComponent className='coz-btn' />` but still will be able to have "type" safety for your app classes.

Bottom line : LGTM 👍  ! Next time, I think it is better not to mix concerns in PRs so try to have PRs that are specific to 1 problem :)

GoOz:
@ptbrowne You're totally right, I just had 5 big commits waiting and handled them poorly. It should be a cleaner PR now :) Thanks

PR #204, opened 2017-10-12
 🚑  Fix the annoying double tap on nav links on iPad 

GoOz:
So I used the MQ parameter `pointer:fine` and `not` which means that ~~every~~ **any** device with**out** an accurate pointing device, such as a mouse, trackpad or pen, should**n't** show an hover effect in this case.
Pretty efficient on iPad & other tablets.

### DIFF
--- a/stylus/components/nav.styl
+++ b/stylus/components/nav.styl
@@ -109,3 +109,11 @@
             box-shadow   none
             font-weight  normal
             color        charcoal-grey
+
+// Prevent double tap on tablets
+@media not all and (pointer: fine)
+    .coz-nav-item:hover::before
+        content none
+
+    .c-nav-link:hover
+        color slate-grey


GoOz:
[[GLOB] 🐛📱Les boutons ne sont pas réactifs sur iOS, particulièrement le faux bouton version pour envoyer les logs](https://trello.com/c/h3dNaizn/530-glob-%F0%9F%90%9B%F0%9F%93%B1les-boutons-ne-sont-pas-r%C3%A9actifs-sur-ios-particuli%C3%A8rement-le-faux-bouton-version-pour-envoyer-les-logs)

GoOz:
@y-lohse so I rewrote the code after realizing that Firefox didn't support `pointer` or any MQ interaction at all (I really thought it did…). The only issue I got is that I couldn't test it. Because, when using an emulator, i do have a mouse pointer so it doesn't work obviously. This needs to be tested on a real device to make sure it is fixed.

y-lohse:
This seems to be working ! So still good for me.

PR #206, opened 2017-10-16
Fix/lastminutefixes

GoOz:
Last minute fixes. 

Some issues after the last refactoring were well hidden.

### DIFF
--- a/react/Button/styles.styl
+++ b/react/Button/styles.styl
@@ -9,7 +9,6 @@
     @extends .c-btn--regular
 
 .coz-btn--secondary
-
     @extends .c-btn--secondary
 
 .coz-btn--danger
--- a/react/Button/styles.styl
+++ b/react/Button/styles.styl
@@ -1,0 +1,22 @@
+@require '../../stylus/tools/mixins'
+@require '../../stylus/settings/palette'
+@require '../../stylus/components/button'
+
+.coz-btn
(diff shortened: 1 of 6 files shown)


stylus/components/popover.styl
```diff
@@ -30,7 +30,7 @@ $popover
     button
     [role=button]
         display          block
-        padding          em(8px) em(32px) em(8px) em(42px)
+        padding          em(8px, 16px) em(32px, 16px) em(8px, 16px) em(42px, 16px)
```
ptbrowne: I wonder if we could use rem here ?

react/Button/styles.styl
```diff
@@ -0,0 +1,22 @@
...
+
+.coz-btn
+    @extends .c-btn
+
+.coz-btn--regular
+    @extends .c-btn--regular
+
+.coz-btn--secondary
+
+    @extends .c-btn--secondary
```
ptbrowne: extra whitespace ?

GoOz: Good catch! Thanks.

stylus/components/popover.styl
GoOz: Indeed, that would in fact be a way better solution. I forgot about this, thanks.

ptbrowne:
Thx !

PR #217, opened 2017-11-22
Add direction to breakpoints

ptbrowne:
Thanks to the direction parameter we can change the direction of the media query if we want.


ptbrowne:
:D

y-lohse:
(I had the same thought but couldn't find a better work than `direction`)

PR #230, opened 2017-12-04
feat: default theme for button

enguerran:
I believe we need to provide default value for __every__ required props.

And I believe we need to set which props are required using [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html).

(I notice we speak about _default value_ in #190 too)

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -3,12 +3,12 @@ import cx from 'classnames'
 import styles from './styles'
 
 export default (props) => {
-  const { theme, busy, className, children, onClick } = props
+  const { theme = 'regular', busy, className, children, onClick } = props
   return <button
     aria-busy={busy}
     role='button'
-    className={cx(styles['coz-btn'], theme ? styles[`coz-btn--${theme}`] : null, className)}
+    className={cx(styles['coz-btn'], styles[`coz-btn--${theme}`], className)}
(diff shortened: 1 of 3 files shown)


enguerran:
I believe this recommandation must be added everywhere. I will update this very PR with proptypes.

react/Button/index.jsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
 import cx from 'classnames'
+import PropTypes from 'prop-types'
 import styles from './styles'
 
-export default (props) => {
-  const { theme = 'regular', busy, className, children, onClick } = props
+const Button = (props) => {
+  const { theme = 'regular', busy = false, className = '', children, onClick = () => { console.log('button was clicked') } } = props
```
GoOz: You forgot to remove the `console.log`. :)

ptbrowne: No default for onClick please :)

enguerran: No, I don't. That is a default value for the `onClick` function.

GoOz: Oh I get it, sorry. I let you two decide what's best then.

enguerran:
Ok, let me know if I get suggestions (approval or not) by code owners.

ptbrowne:
Thanks @enguerran  !

PR #231, opened 2017-12-04
fix: ligter margins for modal elements on mobile

GoOz:
Before:
screenshot_2017-12-04_17-33-23

After:
screenshot_2017-12-04_17-31-21

### DIFF
--- a/stylus/components/modals.styl
+++ b/stylus/components/modals.styl
@@ -63,6 +63,7 @@ $modals
 
         +small-screen()
             font-size 1.25rem
+            margin-bottom .5rem
 
     .coz-btn-modal-close
         position         absolute
@@ -95,6 +96,7 @@ $modals
                 display flex
                 flex-direction column-reverse
                 flex-shrink 0
+                margin-top .5rem
                 button
                     width 100%
                     margin-left 0


ptbrowne:
Screenshot please ?

ptbrowne:
Talking about margins, I am under the impression that we are not particularly strict about the vertical rhythm of our components. I have seen : https://csswizardry.com/2012/06/single-direction-margin-declarations/ and I would like to have a better system concerning margins or guidelines. I feel like we sprinkle `margin-bottom/top` around and it seems quite fragile.

I remember for example that during Friday's demo I have seen a margin bug because we used an h3 in modal.

GoOz:
Totally. There's no vertical rhythm at all, yet. But adding this now would be a lot of work that is not really needed right now. It's in my todo list though.

ptbrowne:
>  But adding this now would be a lot of work that is not really needed right now. It's in my todo list though.

It'd be great if we you could share your vision/guidelines for vertical rhythms during a crosspol, I feel if it can an ongoing effort by everybody instead of a one-off refactor.

GoOz:
[(3) Améliorations de style](https://trello.com/c/MwYIVnn8/639-3-am%C3%A9liorations-de-style)

GoOz:
@ptbrowne I'll share my vision about vertical rhythm when I have one :D
This is something I want to add but as far as I'm concerned, this is more a nice-to-have feature than a mandatory one for now, event though I do agree it should have been here from the beginning (easier to implement that way) but for now there's way more important matters, I think.

PR #236, opened 2017-12-07
Build cozy-ui.min.css for npm publishing

CPatchane:
No need to have it in the repository finally but inside the NPM packages :)
So it will be built using Travis just before being published the package on NPM.


CPatchane:
[Build + push automatique du build.styl de cozy-ui dans un /dist/cozy.ui.min.css](https://trello.com/c/6wwus33L/21-build-push-automatique-du-buildstyl-de-cozy-ui-dans-un-dist-cozyuimincss)

CPatchane:
Seems legit, I'll update the PR 👍

PR #237, opened 2017-12-07
Deprecate with cross

ptbrowne:
- [x] HOC to handle migrations
- [x] `withCross` property is renamed to `closable`. Backward compatibility is preserved via automatic renaming of the property
- [x]  Add `dismissAction` property that will be callled on click outside on on click on the cross. Backward compatibility is preserved by copying `secondaryAction` into this property if it does not exist

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import styles from './styles.styl'
 import Overlay from '../Overlay'
+import migrateProps from '../helpers/migrateProps'
 
 const ModalContent = ({children, className}) =>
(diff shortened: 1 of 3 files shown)


react/Modal/index.jsx
```diff
@@ -92,18 +93,34 @@ Modal.propTypes = {
...
+        newProps.dismissAction = props.secondaryAction
+        const msg =
+          'In the future, `secondaryAction` will not be bound to closing actions (clicking on cross, clicking outside), please use `dismissAction` for that'
+        return [newProps, msg]
+      } else {
+        return [props, null]
+      }
+    }
+  }
+])(Modal)
```
CPatchane: 👍

PR #239, opened 2017-12-07
refactor: Import only lodash functions used

kosssi:


ptbrowne:
OK, for the record, this could be done automatically via `babel-plugin-lodash` during the build.

In: 

```js
import { flow, map, filter } from 'lodash'

const a = flow(map(x => x), filter(x => x))([1, 2, 3])
```

```bash
pbpaste | babel --plugins lodash
```
 
Out : 

```js
import _filter from 'lodash/filter';
import _map from 'lodash/map';
import _flow from 'lodash/flow';


const a = _flow(_map(x => x), _filter(x => x))([1,2,3]);
```

PR #243, opened 2017-12-08
Intent Opener

ptbrowne:
- [x] Wraps a component and provides an `onClick` handler that will open a modal for the intent.
- [x] Intents are opened in a modal taking the majority of the viewport.
- [x] `onComplete` handler
- [x] `onDismiss` handler
- [x] the developer can choose if she wants the modal to be `closable`
- [x] loading indicator while the `iframe` has not fired the `onload` event

Demo : http://recordit.co/P4u5Kmhj9J


ptbrowne:
I rebased since the sidebar PR was merged.

react/IntentOpener/styles.styl
```diff
@@ -0,0 +1,30 @@
+@require 'settings/breakpoints'
+
+// TODO remove the second selector when specificity
+// of cozy-ui selectors has decreased
+.intentModal.intentModal
+    // takes 80% of the screen on desktop
+    max-width unset
+    width 80vw
+    height 80vh
```
GoOz: Beware, using `vh`unit is a dangerous game. FYI on Chrome mobile & Safari mobile (if I remember correctly), `100vh` is more than 100% of the actual viewport's height. No problem with `vw` though.

GoOz:
@ptbrowne I know it opens a "cross-app" modal but the mechanism behind is way too "jsxy" for me.

CPatchane:
Should we handle a loading spinner in this component? It could avoid to wait in front of a blank page during the iframe/intent loading :/

ptbrowne:
> Should we handle a loading spinner in this component

I wanted to add this but the create() method attached to the Promise does not provide a mean to control the `iframe`. Ideally we could put a "ref" callback to `create` that gives the DOM node for `iframe` so that we can listen the `onload` event. I think we should put that in another issue.

react/IntentOpener/styles.styl
ptbrowne: OK, I take note of this, if I see any problem, I'll use another technique. Thanks !

ptbrowne:
> @ptbrowne I know it opens a "cross-app" modal but the mechanism behind is way too "jsxy" for me.

What do you mean ?

react/IntentOpener/IntentIframe.jsx
```diff
@@ -0,0 +1,41 @@
...
+  componentDidMount () {
+    const { action, doctype, options, onComplete } = this.props
+
+    const create = this.props.create || cozy.client.intents.create
+
+    create(action, doctype, {
+        exposeIntentFrameRemoval: true,
+        ...options
+      })
+      .start(this.intentViewer)
```
CPatchane: In fact you can add a callback here to be run on created (https://github.com/cozy/cozy-client-js/blob/master/docs/intents-api.md#cozyclientintentscreate):
```
.start(this.intentViewer, onReadyCallback)
```

GoOz:
> What do you mean ?

I mean I'm no JSX expert.

ptbrowne:
I added the loading indicator.

PR #245, opened 2017-12-11
✨ feat: Media object

ptbrowne:
Visible on https://ptbrowne.github.io/cozy-ui/react/#media


CPatchane:
Should we handle a loading spinner in this component? It could avoid to wait in front of a blank page during the iframe/intent loading :/

ptbrowne:
> What does "bd" stand for already? body? If so, is it explicit enough? Because I wouldn't have guess without the article from Walton.

Yes it stands for `Body` but I want to keep the name short and don't want it to be confusing with0 `<body />`

I understand the possible confusion but when you start to use `Media`, it is very easy to see it and it will always be a child of <Media /> so you'll have the context.

> Why not adding the margin right by default?

Because `` can be used to the right and then the margin-right would not work. So maybe put a margin-right to the first-child and a margin-left to the last-child would be a solution ? But then what about the Img in the middle of two Bd ?

Again, for me it is a question of responsibility and following the Unix principle. It is easier to reason about things when they do one thing and one thing well. `<Media />` is used to manage vertical alignment between several components, not horizontal space.

GoOz:
That's confusing. I thought all of those react components were supposed to be ready-to-use components. It surely is functional but you still have some styling to do, which, if I remember well, is the opposite of what we discussed. I don't think we're on the same track…

ptbrowne:
`<Media />` is an *object* in ITCSS parlance.

> define undecorated design patterns, for example media object known from OOCSS

It is not a *component*. It is low level and will be used to build new *components*. As such, it is not a problem for me if when you use to build *components*, you have to apply a utility class or to do a little bit of styling.

For example, here https://github.com/cozy/cozy-ui/pull/246/files#diff-ea25ad7ecae9b5d0ebfaa8f34b3d24d8R19, I use `.c-menu__item-icon` to apply specific margins so that the horizontal margins are the same for the icon.

GoOz:
OK then, I got nothing against ITCSS object, far from it. 😄 
But It may be confusing to list it next to styled components.
I guess, as more and more components are created, at some point, It will be necessary to organize them in some way. 
Not now, though. I will look into it later, and I will surely ask for input from you about that. 👍

PR #246, opened 2017-12-11
✨ feat: add doc for <Menu /> and improvements

ptbrowne:
https://ptbrowne.github.io/cozy-ui/react/#menu

Prop name | Type | Default | Description
-- | -- | -- | --
position | string | left | Use position='right' to display the menu to the right
disabled | bool | false | Disables the menu
component | element | null | Specifies a custom component for the opener
onSelect | func | null | onClick handler for non disabled items
onSelectDisabled | func | null | onClick handler for disabled items

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -1,7 +1,6 @@
 import React, { Component } from 'react'
 import ReactDOM from 'react-dom'
 import Hammer from 'hammerjs'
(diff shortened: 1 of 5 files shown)


ptbrowne:
> There's something I don't get, why do you make styles to this component only? why not having a $menu in the Stylus files and require it?

I plan to do this later. There was already a `$menu` (relying on bosonic tags) and I don't want to break things.

PR #259, opened 2017-12-15
Add c-nav-icon and c-nav-text to build.styl

CPatchane:


stylus/cozy-ui/build.styl
```diff
@@ -249,6 +255,12 @@
 .c-nav-link
     @extend $nav-link
 
+.c-nav-icon
```
CPatchane: Do you see an example to put for `.c-nav-icon` in the comments above?

GoOz: Not really I haven't use this component yet, but I guess you can get inspiration from the react styleguide.

CPatchane: I will keep it out of the comment for now, I use it with the `<Icon/>` react component. We may need to test it with a simple `img` or `svg`

CPatchane:
I make a better usage of `c-nav-icon`. Tested with the NavIcon component as well and it works.
I add an example in the kss styleguide @GoOz

stylus/cozy-ui/build.styl
CPatchane: Tested with a svg here and it works :)

stylus/components/nav.styl
```diff
@@ -91,7 +91,7 @@ $nav-link
         box-shadow   inset 4px 0 0 0 dodgerBlue
         font-weight  bold
 
-        .nav-link__icon
```
CPatchane: @GoOz Should we need a retro-compatibility here?

GoOz: Well, ask @ptbrowne I never used it myself yet.

CPatchane: Ok, since it was used only for the `<Nav />` component indeed, we don't need one.

PR #260, opened 2017-12-17
Deprecate font size em

ptbrowne:
- [x] Add a deprecate() function to stylus via its plugin API to be able to show deprecation messages that output the selector() using the deprecated function
- [x] Add deprecation messages for em() and font-size()
- [x] Show deprecation messages behind an environment variable flag not to pollute everything while the work on font-size and em is not done. Later we could remove this environment variable flag.

The environment variable flag is `SHOW_STYLUS_DEPRECATION`. If you launch your app watcher with this variable, you'll have deprecation warnings.

```js
env SHOW_STYLUS_DEPRECATION=1 yarn watch:browser
```

### DIFF
--- a/stylus/scripts/deprecate.js
+++ b/stylus/scripts/deprecate.js
@@ -1,0 +1,19 @@
+const nodes = require('stylus/lib/nodes')
+const chalk = require('chalk')
+
+const last = function (arr) {
+  return arr[arr.length - 1]
+}
+const plugin = function () {
+  return function (style) {
+    style.define('deprecate', function (msg) {
+      const parentSel = last(this.selectorStack)[0]
+      if (process.env.SHOW_STYLUS_DEPRECATION) {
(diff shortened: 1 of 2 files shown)


y-lohse:
Todo: once merged, create an issue in our projects so we actually think about using this flag.

ptbrowne:
I was thinking this morning, do you think the flag is necessary ?

GoOz:
I think it would be better if it just works when you launch `yarn watch:drive`. People will forget to use the env flag. I'm pretty sure of it. And if they don't want the warnings, they just can update the code. If it's not annoying it will never be updated :)

ptbrowne:
> And if they don't want the warnings, they just can update the code. If it's not annoying it will never be updated :)

That was my thought. @y-lohse  ?

y-lohse:
Yeah that works for me too.

PR #270, opened 2017-12-20
Spinner classname

ptbrowne:
- Add className to spinner
- Add Proptypes


GoOz:
NAAAAAAH I was doing it… :(

PR #276, opened 2017-12-22
React compatible

ptbrowne:
Since in our tests we use React (I plan to try to use preact with enzyme but wanted to have something running quick), we need to have cozy-ui compatible with React. We also advertise cozy-ui as a React library of components so it should anyway be compatible.

- `props` and `state` are not passed in the `render()` of `Component`s in React so we use `this.props`, `this.state`
- `context` properties are not passed to `CustomComponent` if you do not add a static property `contextTypes` to `CustomComponent`

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -41,7 +41,7 @@ export class I18n extends Component {
   }
 
   render () {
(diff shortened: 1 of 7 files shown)


ptbrowne:
Ping. I would like to have at least 1 approval before merging :)

PR #278, opened 2017-12-28
Feat/enhanced button

GoOz:
Enhanced `<Button />` components
- [x] Addition of `<ButtonLink />` which look like exactly the same as `<Button />` but use a `<a>` tag instead of `<button>` tag (with `target` & `href` params added)
- [x] Added the `disabled` param on `<Button />`
- [x] Made regular theme the default rendering of `<Button />` so there's no need to specify the theme anymore if you want a regular rendered style
- [x] Icon's color is by default the `currentColor` of its button parent. Unless you want to overwrite this color, you don't need to specify a color for your icon anymore.
- [x] Made default size of `<Icon />` `16px` instead of `1em` to avoid  font-size inheritance (`rem` didn't work)
- [x] Removed some icon `fill` attribute
- [x] Removed useless button placeholders that were replaced by the use of `<Icon />` in the button
- [x] minor CSS optimisations


react/Button/ButtonLink.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+ButtonLink.defaultProps = {
+  theme: '',
+  href: '',
+  target: false,
+  className: '',
+  onClick: null
+}
+
+export default ButtonLink
```
ptbrowne: I think `Button` and `ButtonLink` should be in the same file since they share so much.

For example the className function can be shared.

```
const btnClass = function (theme, className) {
  return cx(
        styles['c-btn'], {
          [styles[`c-btn--${theme}`]] : theme
        },
        className)
}
```

```
className={ bntClass(theme, className) }
```

Some PropTypes and defaultPropTypes could be shared too. Since proptypes and defaultProps are regular javascript object, you can use spreading to share attributes.

Example : 

```
const btnPropTypes = {
  children: PropTypes.node.isRequired,
  theme: PropTypes.string,
  className: PropTypes.string,
  onClick: PropTypes.func
}

Button.PropTypes = {
   ...btnPropTypes,
  busy: PropTypes.bool
}

BtnLink.PropTypes = {
   ...btnPropTypes,
  href: PropTypes.string.isRequired,
  target: PropTypes.bool,
}
```

I think it would make the code more maintainable since there would be less risk of divergence between the two files.

GoOz: Well the buttons were in the same file but something breaks with Styleguidist, something related to the fact that those components are not exported as default. I don't really get it because the error i got is `String contains an invalid character` which is not really explicit.
I can try again though, maybe with a clear mind now…

GoOz:
@ptbrowne Better ? (I kinda like it better that way 😉)

PR #281, opened 2018-01-08
fix: selection bar's icons on mobile

GoOz:
Removed obsolete instructions due to the use of `<Icon />` component instead of CSS backgrounds


GoOz:
damn, you're right again… would you please stop that? 😝

y-lohse:
I will do my best 😀

GoOz:
Shoule be better now.

PR #287, opened 2018-01-08
Fix/repair

GoOz:
I messed up with Modal's buttons color, my bad. 
And I saw a Stylus bug that I can't fix right now. 
Basically, with Stylus, when you `@extend` any `$placeholder` within a media query, Stylus just… ignores the MQ and output the instruction as a global one. 
See https://codepen.io/GoOz/pen/yppqmm Yay /o\

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -46,7 +46,7 @@ const ModalButtons = ({ secondaryText, secondaryAction, secondaryType, primaryTe
     (
       <div className={classNames(styles['coz-modal-content'], styles['coz-modal-buttons'])}>
         { displaySecondary &&
-          <button className={styles['c-btn']} onClick={secondaryAction}>
+          <button className={classNames(styles['c-btn'], styles['c-btn--' + secondaryType])} onClick={secondaryAction}>
             {secondaryText}
           </button>
         }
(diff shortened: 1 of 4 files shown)


y-lohse:
FYI [sass has the same limitation](https://stackoverflow.com/questions/14840918/extending-selectors-from-within-media-queries-with-sass), but it also throws a big error if you try to do it.

GoOz:
I know, they should both work the same way, but somehow Stylus won't behave on this. Still working on this though, it got me interested. 🤔

GoOz:
[Publication Android](https://trello.com/c/9U0Dl7qN/712-publication-android)

GoOz:
Turns out you just **can't** `@extend` inside `@media` rules, that's just not possible… so I had to duplicate the code. I don't like it, I think this is silly, especially when you use a preprocessor, but hey… what can I do?

ptbrowne:
Have you tried a mixin approach like @y-lohse suggested ?

GoOz:
I did. It won't work. This issue is duplicated a number of time on Stylus' Github and no solution so far except what I did (some did it dynamically but still the same duplication idea).

ptbrowne:
I do not understand why the mixin won't work. Here it works with a mixin inside a MQ : https://codepen.io/anon/pen/ppaNGd

GoOz:
@ptbrowne It's because you're doing it wrong. https://codepen.io/GoOz/pen/baLqGx I just can't `@extend` inside a MQ, no matter how I do it.

ptbrowne:
I understood the `@extend` part and that it does not work in Stylus.

The suggestion is to use a mixin *instead of* `@extend` to de-duplicate code. LGTM otherwise.

PR #291, opened 2018-01-10
fix: alerts MQ for desktop

GoOz:
Don't know exactly why but on Photos, the `@media all and (min-width alert-width)` turned into `@media all and (min-width: 480em)` which is… bad. With rem, everything is in order.


GoOz:
[Bug du toaster dans Photos](https://trello.com/c/A8tbyIpK/63-bug-du-toaster-dans-photos)

PR #297, opened 2018-01-16
✨ feat: use function to generate NavLink component

ptbrowne:
## Context

We do not export a cozy-ui themed NavLink since it is supposed to build on `NavLink` from react router (`Link` in react-router version 3) and we do not want to ship react-router with cozy-ui.

## Problem
Before, the user had to write its NavLink component by hand. See warning here : https://github.com/cozy/cozy-ui/tree/master/react/Sidebar

## Solution

I provide a function taking the react-router `NavLink`  component (`Link` in react-router 
3) and returns a cozy-ui themed `NavLink`.


react/Nav/index.jsx
```diff
@@ -15,14 +15,16 @@ export const NavText = ({ children }) => (
   </span>
 )
 
-export const NavLink = ({ children }) => (
-  <div className={styles['c-nav-link']}>{ children }</div>
+export const genNavLink = NavLink => ({ to, children }) => (
+  <NavLink to={to} activeClassName={styles['c-nav-link']} className={styles['is-active']}>
```
GoOz: You inverted the two classNames, otherwise it works like a charm. Thanks

ptbrowne:
Thanks should be ok now !

PR #299, opened 2018-01-17
feat: add explicitely One column layout

GoOz:
Added a One Column layout. Well it was already kinda here already but it's more explicit this way.

### DIFF
--- a/stylus/objects/layouts.styl
+++ b/stylus/objects/layouts.styl
@@ -21,57 +21,61 @@
 \*------------------------------------*/
 
 $centerized
-    position   absolute
-    top        50%
-    left       50%
-    transform  translateX(-50%) translateY(-50%)
+    position absolute
+    top 50%
+    left 50%
+    transform translateX(-50%) translateY(-50%)
 
 
 $content-center
-    display          flex
-    align-items      center
-    justify-content  center
+    display flex
+    align-items center
+    justify-content center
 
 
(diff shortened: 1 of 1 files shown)


GoOz:
[Ajouter un layout mono-colonne](https://trello.com/c/fWVp5jcG/69-ajouter-un-layout-mono-colonne)

GoOz:
[Ajouter un layout mono-colonne](https://trello.com/c/fWVp5jcG/69-ajouter-un-layout-mono-colonne)

GoOz:
[Ajouter un layout mono-colonne](https://trello.com/c/fWVp5jcG/69-ajouter-un-layout-mono-colonne)

GoOz:
[Ajouter un layout mono-colonne](https://trello.com/c/fWVp5jcG/69-ajouter-un-layout-mono-colonne)

y-lohse:
Code looks good, although I wonder about something. On very large screens, we don't want 100% of the width, so there should be a max-width somewhere. But maybe it should be in a separate wrapper?

GoOz:
Exactly. If someone wants to build an app that would take 100% of the width available, I think it should be by default. But if you want your content to have a max-width then just add you own style to the wrapper of you content (or override the Layout's style).

y-lohse:
Ok, that makes sense, thanks 👍

PR #303, opened 2018-01-17
✨ feat: Modal is rendered into a Portal

### LINKED ISSUE #301: Mount Modal in a portal
Currently, Modal is mounted in the DOM tree just where we declare it. But it can lead to positionning problems. It would be great to mount it in a portal at the root of the `body`, so it is completly isolated from the rest.

The thing is that even if you use preact-compat, it doesn't add `ReactDOM.createPortal` API. But there is [preact-portal](https://github.com/developit/preact-portal) to use portals in preact.

ptbrowne:
Now, Modals are rendered exactly like other components and their DOM node is rendered inside the DOM node of the component rendering it. It can cause problems when the parent node has `overflow: hidden` or is `position: fixed`. We experienced this first hand in Bank when we rendered a Modal inside a Menu : the UI was broken since the Modal was rendered inside the Menu instead of being rendered inside the body.

Portals solve this problem in React world by transporting their children where you want (most of the time body).

You can see the result on https://ptbrowne.github.io/cozy-ui/react/#modal : 

image

Fixes #301

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,6 +66,7 @@
     "hammerjs": "^2.0.8",
     "md5": "^2.2.0",
     "normalize.css": "^7.0.0",
+    "preact-portal": "^1.1.2",
     "stylus": "^0.54.5"
   },
   "peerDependencies": {
--- a/react/Modal/Readme.md
+++ b/react/Modal/Readme.md
@@ -94,3 +94,18 @@ const { ModalDescription, ModalTitle } = Modal;
     </Modal>}
 </div>
 ```
+
+### Portal modals
+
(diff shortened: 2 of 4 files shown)


ptbrowne:
@y-lohse Which doc ? The doc for into is available under "Props and methods" in the styleguide : 

image

Since we are so close to a launch, I decided to keep backward compatibility and to only render in a Portal if the `into` prop is set.

y-lohse:
My bad, I missed that. All good then !

ptbrowne:
I've added an example in the styleguide alos.

PR #305, opened 2018-01-19
fix: nav mobile was misplaced with smaller icons

GoOz:
Forgot to take a look at the nav with `<Icon />` only on mobile… 😕


CPatchane:
Oops I merged but that was good 👍

PR #306, opened 2018-01-22
feat: add new breakpoint for tablet

GoOz:
Added new default breakpoint for tablet because there was a blank to fill.

### DIFF
--- a/stylus/settings/breakpoints.styl
+++ b/stylus/settings/breakpoints.styl
@@ -146,7 +146,7 @@ mobile()
         {block}
 
 /*
-    np-mobile()
+    no-mobile()
 
     Refers to (min-width: 769px)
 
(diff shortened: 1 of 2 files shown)


stylus/settings/breakpoints.styl
```diff
@@ -158,12 +158,25 @@ no-mobile()
...
+/*
+    no-tablet()
+
+    Refers to (min-width: 1024px)
+
+    Weight: 4
+
+    Styleguide Settings.breakpoints.no-tablet
+*/
+no-tablet()
```
ptbrowne: `larger-than-tablet` seems clearer to me, I was expecting this breakpoint to match (width < small) || (width > medium).

ptbrowne: or `gt-tablet` (gt for greater than)

GoOz: yeah… I keep doubting about the naming. I like your ideas though.

Issue #310, opened 2018-01-30
Buttons, flexbox and safari 10


y-lohse:
[Our buttons use `display inline-flex`](https://github.com/cozy/cozy-ui/blob/master/stylus/components/button.styl#L11) to center their content.

Unfortunately, using `flexbox` on certain elements like `button` is not very well supported in older browser, most notably safari 10, which is still used in the wild. In those cases, the content of the button is aligned to the left.

i've created a [pen](https://codepen.io/anon/pen/oENePQ) with the minimal issue reproduced.

Possible fixes i've found:

- don't use flexbox at all on buttons
- use `display -webkit-inline-box` instead of `display inline-flex` — actually `-webkit-inline-box` is inserted in our code, but safari supports `inline-flex` too so that rule takes precedence
- [using pseudo-elements](https://stackoverflow.com/a/40483475) — seems to work well, but of course if the button `actually` uses pseudo-elements things might very well break

I really have no idea what the consequences of any of these fixes would be, but I'm afraid none of them is backwards compatible.

GoOz:
> use display -webkit-inline-box instead of display inline-flex — actually -webkit-inline-box is inserted in our code, but safari supports inline-flex too so that rule takes precedence

Safari… t(°^°t)

We don't use `::before` or `::after` pseudo-elements in our buttons (so far) so I guess this would do the trick. Thank you for your thorough investigation. 🙏

I'll implement this solution asap.

y-lohse:
Great, thanks!

y-lohse:
![](https://github.trello.services/images/mini-trello-icon.png) [(1) 🐛📱 Libellé des boutons alignés à gauche dans l'app iOS](https://trello.com/c/EhgqPVkX/769-1-%F0%9F%90%9B%F0%9F%93%B1-libell%C3%A9-des-boutons-align%C3%A9s-%C3%A0-gauche-dans-lapp-ios)

GoOz:
Fixed by #311

y-lohse:
@GoOz Welp, it *did* break something. When a button has the `busy` prop, we insert the spinner with an `after`, and it really doesn't like the flex display.
<img width="329" alt="capture d ecran 2018-02-01 09 06 39" src="https://user-images.githubusercontent.com/2261445/35667936-bcfc3a1e-072f-11e8-8463-cbc76c88cc5a.png">

GoOz:
damn I forgot this…

y-lohse:
[...]

PR #311, opened 2018-01-30
fix: center text in buttons on Safari 10

GoOz:
Wide buttons on safari 10 had their text aligned to the left instead of being centered due to a wrong implementation of flexbox in the browser.
This trick fix this issue.

https://trello.com/c/EhgqPVkX

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -56,6 +56,12 @@ $button
             position   relative
             top        -.0625rem
 
+    // Trick for Safari 10 to ensure that
+    // button's text is horizontally centered
+    &::before
+    &::after
+        content ''
+        flex 1 0 auto
 
 $button--regular  // Deprecated
     @extend $button


GoOz:
[(1) 🐛📱 Libellé des boutons alignés à gauche dans l'app iOS](https://trello.com/c/EhgqPVkX/769-1-%F0%9F%90%9B%F0%9F%93%B1-libell%C3%A9-des-boutons-align%C3%A9s-%C3%A0-gauche-dans-lapp-ios)

GoOz:
[#310: Buttons, flexbox ans safari 10](https://trello.com/c/B5e7xSOd/84-310-buttons-flexbox-ans-safari-10)

GoOz:
True but I need flexbox on those button right now and I know there's no need for pseudo elements now or in the near future so… let's deal with it if it pops some day and even so I'm quite confident we can fix this by adding a wrapper, but I don't want to do that just yet. The less the better. :)

PR #312, opened 2018-02-01
Add divider + lowered paddings between cells

GoOz:
Added a "divider"  style in table-like layout. (Used in Contacts, Bank and/or Health apps too I think)
Preview on the [KSS Styleguide](https://gooz.github.io/cozy-ui/styleguide/section-components.html) in the Table section.

Also, due to recent Design Styleguide changes, padding inside table cells has been decreased.

### DIFF
--- a/stylus/components/table.styl
+++ b/stylus/components/table.styl
@@ -54,14 +54,14 @@ $table-row-selected
 
 $table-cell
     box-sizing border-box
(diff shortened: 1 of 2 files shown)


GoOz:
[Intercalaire [Contacts/Bank]](https://trello.com/c/6ip0HqTU/79-intercalaire-contacts-bank)

Issue #309, opened 2018-01-30
Fixed header/footer for IntentModal


enguerran:
For [cozy/cozy-contacts](https://github.com/cozy/cozy-contacts) I needed to create an intent modal with:
- fixed header
- fixed footer
- scrollable main content

The actual code is referenced in the commits from this PR: https://github.com/cozy/cozy-contacts/pull/26

I think it could be very useful to add those CSS rules and create adequat preact components in cozy-ui as every apps could use them.

![screenshot-2018-1-30 cozy-contacts](https://user-images.githubusercontent.com/701648/35555869-db850366-05a0-11e8-911b-897ec2e29890.png)

GoOz:
![](https://github.trello.services/images/mini-trello-icon.png) [mini-layout fixed & scroll](https://trello.com/c/WOkefKL3/90-mini-layout-fixed-scroll)

GoOz:
Fixed #364

PR #314, opened 2018-02-02
feat: Add IntentHeader component

GoOz:
Added a Intent modal header component `<IntentHeader />`
Looks like this (just the title part):
screenshot_2018-02-02_12-32-39

I don't really know if it's possible to easily add an example of this component in Styleguidist as I'm not really aware of how intents works. Does one of you have an idea?


GoOz:
[#309: Intent Header Modal](https://trello.com/c/f2AmgmgB/80-309-intent-header-modal)

ptbrowne:
The IntentOpener is broken in the styleguide, I'll try to fix it and incorporate this.

react/IntentHeader/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+import styles from './styles'
+
+export const IntentHeader = ({ appName, appEditor, appIcon }) =>
+  (
+    <h2 className={styles['intentHeader']}>
+      <div className={styles['intentHeader-title']}>
+        <img className={styles['intentHeader-icon']} src={appIcon} />
+        {appEditor && <span>{appEditor}&nbsp;</span> }
+        {appName}
+      </div>
```
ptbrowne: Why an extra div ?

ptbrowne: Shouldn't the icon and the text be vertically aligned ? I would use a Media here.

GoOz: Now that you pointed that out, actually the `h2` & `div` are inverted. It should be a wrapping `div` and `h2` as a child that wrap the app name, and now that it it an iframe, `h2` should be an `h1`.
Also, the need for an extra wrapper is more because some other stuff may (or may not, not quite validated yet) come in this header that wouldn't be a title.

If I remember correctly `<Media />` doesn't allow to choose the markup, right?

CPatchane:
Is this component used from the client? or the intent service application? If this is a client, how will the application using the intent get the good application icon?

GoOz:
@CPatchane It's meant to be used by the service because the application using the intent can't know this kind of information. :)

ptbrowne:
LGTM

PR #319, opened 2018-02-06
Feat/size modal

GoOz:
Modals no longer have their width defined by its content. 
Things went wild too much so I added a set of predefined sizes.
* `small` (544px/34rem)
* `medium` (576px/36rem)
* `large` (640px/40rem)
* `xlarge` (800px/50rem)
* `xxlarge` (960px/60rem)

`small` being the default size if you don't choose any for your modal.

Preview: https://gooz.github.io/cozy-ui/react/#modal

Also fixed the size of the cross button.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import styles from './styles.styl'
 import Overlay from '../Overlay'
+import { Button} from '../Button'
 import Icon from '../Icon'
 import migrateProps from '../helpers/migrateProps'
 import palette from '../../stylus/settings/palette.json'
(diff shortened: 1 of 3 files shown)


react/Modal/index.jsx
```diff
@@ -26,12 +27,14 @@ const ModalTitle = ({ children, className }) =>
   )
 
 const ModalCross = ({ onClick, className }) => (
-  <button
-    className={classNames(styles['c-btn'], styles['c-btn--close'], styles['coz-btn-modal-close'], className)}
+  <Button
+    theme="secondary"
```
ptbrowne: theme='secondary' for the cross ? Should'nt it be transparent or something ? Why is is a Button instead of a button ? Why is this in this PR ?

GoOz: Oh right, should be `alpha`, not `secondary`.

> Why is is a Button instead of a button ?

Why not? I thought it would better to use our own components, don't you think?

>Why is this in this PR ?

It should have been a really quick fix, but I guess I can extract this into another PR.

ptbrowne: OK for me if theme is alpha. alpha IMHO is a lot less clear than transparent. Since you can have alpha=1 :/

GoOz:
[Review Modal Specs](https://trello.com/c/VX12nwip/45-review-modal-specs)

y-lohse:
I'm not sure if this should be part of this PR or not, but Claire mentioned that we're going to have 2 padding variations for modal, essentially `regular` like now and `fat` for 32px. Just sharing the info here.

GoOz:
@y-lohse well it could be. Did it sound like everything was already settled or that it must be thought through before?

y-lohse:
I think you should double check with her — it sounded pretty clear to me though.

GoOz:
So not quite ready yet. It'll be for another PR later.

Issue #318, opened 2018-02-05
Avatar initial centering


y-lohse:
I just noticed a small problem with avatars on sizes other than `small` — the initials are not properly centered vertically. It seems to be because of the `line-height`which is set to `2.2` and probably needs to be smaller for larger font size. I'd say `1.9` looks right for the `medium` size.

<img width="89" alt="capture d ecran 2018-02-05 18 15 09" src="https://user-images.githubusercontent.com/2261445/35818492-87b63412-0aa0-11e8-9bbf-c3a1beb3af6d.png">

GoOz:
And I guess, if you take a look at Firefox, It will be OK. Those two can't behave the same when it comes to line-height. I need to find a better way to vertically center this. Thanks.

y-lohse:
Ah, I didn't look at another browser. I would set the `line-height`to 1 and let flexbox do the rest of the work? But I might miss something.

GoOz:
Fix with #321

PR #321, opened 2018-02-06
Fix: vertical alignment of initials & add Empty avatar view

GoOz:
- [x] Added empty view for Avatar
- [x] Better alignment for initials
- [x] A little bit of clean up here and there


GoOz:
[#318: Avatar initial centering](https://trello.com/c/N1ZGg2wu/89-318-avatar-initial-centering)

PR #323, opened 2018-02-07
fix increase alert size

GoOz:
Alerts are now wider as requested by the designer.


GoOz:
[larger alerts](https://trello.com/c/YThnF3zP/81-larger-alerts)

Issue #298, opened 2018-01-17
Allow all HTML available targets for ButtonLink


CPatchane:
For now there is only two choices of targets for the `ButtonLink` component. It could be great to be able to use this attribute with the same possible values than the HTML `target` attribute.

GoOz:
Fix with #327

PR #327, opened 2018-02-08
feat: ButtonLink's target attribute becomes a string

GoOz:
…instead of a boolean

It was decided to just pass the target's value as the HTML spec allows and let HTML do its job.


GoOz:
[#298: Allow all HTML available targets for ButtonLink](https://trello.com/c/30NxrQc7/72-298-allow-all-html-available-targets-for-buttonlink)

PR #328, opened 2018-02-09
fix: size overwrite classes have more specificity to avoid conflict

GoOz:
It turns out that, for some reasons I quite don't understand, sometimes CSS builders don't necessarily respect the classes order in the component's styles.

In this case, the base class is `.c-btn` is overwritten by modifier classes like `.c-btn--small` or `.c-btn--narrow`. In order to work properly, the basic class should be at the top of the file and modifier classes beneath it. As CSS builders don't always respect this, modifiers classes that appear before the basic class are overwritten by it. 
So I added a little more weight on modifer classes specificity in order to keep the overwrite effect even if the order is not respected.

### DIFF
--- a/react/Button/styles.styl
+++ b/react/Button/styles.styl
@@ -3,6 +3,21 @@
 .c-btn
     @extend $button
 
+    &.c-btn--tiny
+        @extend $button--tiny
+
+    &.c-btn--small
+        @extend $button--small
+
+    &.c-btn--large
+        @extend $button--large
+
+    &.c-btn--full
+        @extend $button--full
+
+    &.c-btn--narrow
+        @extend $button--narrow
+
 .c-btn--regular
     @extend $button--regular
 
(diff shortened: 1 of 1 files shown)


ptbrowne:
In what project are you witnessing this ? It is important to understand this issue. There are some PostCSS plugins like mqpacker (media-query packer) that reorder things. I think we should remove this plugin as when I tested, it was not bringing much size gain and did things that were completely unexpected.

GoOz:
@ptbrowne Seen on Contacts app. Not sure it's related to MQPacker though, as those classes are not inside MQs if I remember correctly.

enguerran:
`.btn` classes are the only classes that create difficulties?

GoOz:
I don't remember what was about the last case. But I don't see why it could only be buttons.

ptbrowne:
Apart from this PR that I think is mergeable, I think it should be good to open a issue on this problem. It would be good to have a minimal reproducing case so that we can fully understand the problem.

ptbrowne:
@GoOz  where is the issue ?

GoOz:
@ptbrowne https://trello.com/c/Vnx0gN6x You think we should do a Github issue? I honestly don't know what's best with all those tools 😅

enguerran:
[Souci sur cozy/cozy-ui/react](https://trello.com/c/Vnx0gN6x/44-souci-sur-cozy-cozy-ui-react)

enguerran:
[(2) ETU je peux supprimer un contact depuis sa fiche](https://trello.com/c/eD8KjGbM/724-2-etu-je-peux-supprimer-un-contact-depuis-sa-fiche)

PR #329, opened 2018-02-12
fix: repositionned Intent modal cross button

GoOz:
Cross button in Intent modal was misplaced.


GoOz:
[Position de cross button sur header intent modal](https://trello.com/c/jSUxu51b/93-position-de-cross-button-sur-header-intent-modal)

PR #330, opened 2018-02-12
fix: empty views on tablet don't block click events anymore

GoOz:


GoOz:
[Empty view tablet prevents click](https://trello.com/c/DH1buHxA/96-empty-view-tablet-prevents-click)

Issue #171, opened 2017-09-07
Hero style


ptbrowne:
I think it would be good to standardize the way we do onboarding screens. We have several screens like that in labs apps. Do you also have that in Photos, Drive ?

Introducing `.hero`, `.hero__icon`, `.hero__title`, `.hero__subtitle` could be a good way to do that.

![image](https://user-images.githubusercontent.com/465582/30155104-432818e8-93bc-11e7-92b2-9b15d6eafa23.png)

GoOz:
Why not. If it's used a lot it makes sense.

ptbrowne:
We can start from https://gitlab.cozycloud.cc/labs/cozy-bank/commit/c44faac73a499d0dbe43839033d4976b26b523d3

PR #331, opened 2018-02-13
✨ feat: Hero component

ptbrowne:
Desktop

image

Mobile

image

Doc available here : https://ptbrowne.github.io/cozy-ui/react/#hero

Fixes https://github.com/cozy/cozy-ui/issues/171

### DIFF
--- a/docs/utils.js
+++ b/docs/utils.js
@@ -1,3 +1,5 @@
+import translations from './fixtures/en.json'
+import get from 'lodash/get'
 /** Fake cozy.client.intents.create to demonstrate features in Styleguide */
 export const fakeIntentCreate = (action, doctype, options) => {
   let res
@@ -22,4 +24,8 @@ export const fakeIntentCreate = (action, doctype, options) => {
     return p
   }
   return p
+}
+
+export const t = (path) => {
+  return get(translations, path)
 }
--- a/react/Hero/Readme.md
+++ b/react/Hero/Readme.md
@@ -1,0 +1,40 @@
+A `Hero` is used to to showcase key marketing messages. It
+is perfect for onboarding an app.
+
+```jsx
+const Hero = require('.').default
+const { Title, Subtitle, Section, Sections, Paragraph, CTA, Icon} = Hero
+const Button = require('../Button').Button
+const { t } = require('../../docs/utils')
+
+const RED = '#f52d2d'
+const PURPLE = '#a75bcb'
+const BLUE = '#2d8af2';
+
+<Hero>
+  <Title>{t('Onboarding.title.desktop')}</Title>
+  <Sections>
+    <Section>
+      <Icon color={RED} icon='warning' />
+      <Subtitle>{t('Onboarding.manage-budget.title')}</Subtitle>
+      <Paragraph>{t('Onboarding.manage-budget.description')}</Paragraph>
+    </Section>
+    <Section>
+      <Icon color={PURPLE} icon='paperplane' />
+      <Subtitle>{t('Onboarding.save-time.title')}</Subtitle>
+      <Paragraph>{t('Onboarding.save-time.description')}</Paragraph>
+    </Section>
+    <Section>
+      <Icon color={BLUE} icon='folder' />
+      <Subtitle>{t('Onboarding.cozy-assistant.title')}</Subtitle>
+      <Paragraph>{t('Onboarding.cozy-assistant.description')}</Paragraph>
+    </Section>
+  </Sections>
+  <CTA>
+    <Button onClick={() => flash(t('ComingSoon.description'))}>
+      {t('Onboarding.connect-bank-account')}
+    </Button>
+  </CTA>
+</Hero>
+
+```
--- a/react/Hero/styles.styl
+++ b/react/Hero/styles.styl
@@ -1,0 +1,37 @@
+@require 'settings/breakpoints.styl'
+
+.Hero
+    margin-left auto
+    margin-right auto
+
+    &-title
+        text-align center
+        margin 3rem 0
+
+    &-subtitle
+        margin-bottom 0
+
+        +small-screen()
+            margin-top .5rem
+
+    p
+        line-height 1.25
+
+    &-sections
+        display flex
+        justify-content space-around
+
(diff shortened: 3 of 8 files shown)


react/Hero/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+const { Title, Subtitle, Section, Sections, Paragraph, CTA, Icon} = Hero
+const Button = require('../Button').Button
+const { t } = require('../../docs/utils')
+
+const RED = '#f52d2d'
+const PURPLE = '#a75bcb'
+const BLUE = '#2d8af2';
+
+<Hero>
+  <Title>{t('Onboarding.title')}</Title>
```
GoOz: The title doesn't appear in your example, I suppose this is not on purpose, right?

react/Hero/Readme.md
```diff
@@ -0,0 +1,40 @@
...
+      <Paragraph>{t('Onboarding.save-time.description')}</Paragraph>
+    </Section>
+    <Section>
+      <Icon color={BLUE} icon='folder' />
+      <Subtitle>{t('Onboarding.cozy-assistant.title')}</Subtitle>
+      <Paragraph>{t('Onboarding.cozy-assistant.description')}</Paragraph>
+    </Section>
+  </Sections>
+  <CTA>
+    <Button theme='regular' onClick={() => flash(t('ComingSoon.description'))}>
```
GoOz: I guess you can get rid of the `theme` props since it's by default now.

react/Hero/styles.styl
```diff
@@ -0,0 +1,37 @@
+@require 'settings/breakpoints.styl'
+
+.Hero
```
GoOz: I don't mean to be fussy but wouldn't it be better to have the same syntax of CSS all over Cozy-UI?
I mean like lowercase only for classes and `hero-title` instead of `Hera__title`

react/Hero/styles.styl
```diff
@@ -0,0 +1,37 @@
...
+
+    &__title
+        text-align center
+        margin 3rem 0
+
+    &__subtitle
+        margin-bottom 0
+
+        +small-screen()
+            margin-top 0.5rem
```
GoOz: also the `0.5rem` should be `.5rem`

react/Hero/styles.styl
ptbrowne: OK, I verified in the styleguide and it's true that "We use BEM naming convention but with a slight difference, we use this syntax block-element--modifier." :-)

I'll change this but for the record : I am not particularly a fan of following a standard and changing it just a bit. It is not written why there is this change in the CSS styleguide.

One of the advantage (other that doing like everyone else) I see in keeping with BEM for the Block__Element part instead of having Block-Element is that we can write `style={styles.Block__Element}`. With the homemade syntax we have to write `style={styles['Block-Element']}` since dashes would be interpreted as minus signs. It does not work for modifier though :/

I look forward to the day where we use BEM instead cozy-BEM.

GoOz: For what it's worth BEM is not about the syntax, it's about the methodology. 
The syntax "by default" provided by BEM has never been mandatory, it's purely arbitrary, and they say it openly. It's more a "Choose a meaningful syntax and stick to it" than a "Choose my syntax and stick to it".
Respecting the BEM's example like a dogma makes no sense to me. 
Not that I absolutely want to keep the way it is so far in Cozy-UI, but I'll need a little better reason than "everybody does like the example" to change everything we did so far. 😕

react/Hero/Readme.md
ptbrowne: No it's true I must find out why. Good catch !

react/Hero/Readme.md
ptbrowne: done!

react/Hero/styles.styl
ptbrowne: good catch !

react/Hero/styles.styl
ptbrowne: done

ptbrowne: I changed it.

> Respecting the BEM's example like a dogma makes no sense to me.

This is why I gave a reason that makes sense to me since we use CSS modules.

To me, the case to defend is why we've did differently than the others. We are a framework : if we do something else than everybody else, people will have a harder time to learn and adopt it. The change has to bring value to be meaningful.

I also agree that changing it now is not a good idea since a lot of styles are still used manually (instead of via components). This is why I changed it in my PR.

react/Hero/Readme.md
ptbrowne: done

react/Hero/styles.styl
GoOz: Well I don't really agree with everything you said but we could just discuss it on a proper medium later with pleasure. 😄

docs/utils.js
```diff
@@ -23,3 +25,7 @@ export const fakeIntentCreate = (action, doctype, options) => {
   }
   return p
 }
+
+export const t = (path) => {
+  return get(translations, path)
```
CPatchane: You know my point about avoiding adding a dependency for some little needs. How about that:

```javascript
export const t = (path) => {
  return path.split('.').reduce((acc, part) => acc && acc[part], translations)
}
```

ptbrowne:
```

Since lodash is already there, I am not

GoOz: TIL `yarn why` 🎉

y-lohse: That being said, it should be added to the `dependencies` if we're using it directly (or a `devDep` if it's just for the docs).

ptbrowne: True, there you go :)

CPatchane: So we finally added lodash as a dependency :/

gregorylegarec: [](https://media.giphy.com/media/hppWdK8gcmzXq/giphy.gif)

gregorylegarec: :sob:

y-lohse: Well it was already there when you `yarn install`, and it was (probably) already shipped. So it doesn't actually change anything?

PR #332, opened 2018-02-13
feat: Expects any string for a text

GoOz:
How Avatar works now:
1. If an image is given, then show it.
2. If not, search for a text (the app chooses the order of preference if any)
5. If none of the above is there, just show the default icon.

### DIFF
--- a/react/Avatar/index.jsx
+++ b/react/Avatar/index.jsx
@@ -28,11 +28,12 @@ const nameToColor = (name = "") => {
 };
 
 
-export const Avatar = ({ firstname, lastname, image, size, className }) => {
+export const Avatar = ({ text, image, size, className }) => {
   const colored = {
-    backgroundColor: `${nameToColor(firstname)}`,
+    backgroundColor: `${nameToColor(text)}`,
     color: "white"
   };
+
   return (
     <div
       className={classNames(
@@ -41,7 +42,7 @@ export const Avatar = ({ firstname, lastname, image, size, className }) => {
         },
         className
       )}
-      {...{style: firstname ? colored : undefined}}
+      {...{style: text ? colored : undefined}}
     >
       {image &&
         <img
@@ -50,15 +51,12 @@ export const Avatar = ({ firstname, lastname, image, size, className }) => {
           alt=''
         />
       }
-      {!image && firstname && lastname &&
+      {!image && text &&
         <span className={styles["c-avatar-initials"]}>
-          {[
-            (firstname[0] || "").toUpperCase(),
-            (lastname[0] || "").toUpperCase()
-          ].join("")}
+          {text}
         </span>
       }
-      {!image && !firstname && !lastname &&
+      {!image && !text &&
         <Icon icon='people' />
       }
     </div>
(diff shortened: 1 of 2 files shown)


y-lohse:
I feel like this is misguided. The rules of preference can vary from one app to the other and it shouldn't be the `Avatar`s role to decide what to use. I think the order should be:

- use image if given
- if not, use the provided string
- if there's nothing, show the default icon

And the user can decide what string to provide, be it a name, a company, an email or a star sign.

GoOz:
[Update Avatar](https://trello.com/c/kogWfcdX/94-update-avatar)

GoOz:
Now that you say it, I agree. I just need to deal with strings with spaces inside (like firstname and lastname) to only get their first letter, also could be useful for Company name like "Cozy Cloud".

GoOz:
@y-lohse is this more what you had in mind?

y-lohse:
Yes!

react/Avatar/index.jsx
```diff
@@ -28,11 +28,16 @@ const nameToColor = (name = "") => {
...
+export const Avatar = ({ text, image, size, className }) => {
   const colored = {
-    backgroundColor: `${nameToColor(firstname)}`,
+    backgroundColor: `${nameToColor(text)}`,
     color: "white"
   };
+
+  const getInitials = text => {
+    return text.split(' ').map(word => word[0] || '').join('').substr(0, 2).toUpperCase();
+  }
```
ptbrowne: This function could be outside the component.

ptbrowne:
A part the little change, it LGTM for me !

ptbrowne:
Following the discussion on Mattermost, I agree that the Avatar should not have the `getInitialsLogic` by default and that we should split into two components :

1. one that takes only the text to display, the color and/or the image
2. one more high level like InitialsAvatar  that has the `getInitialsLogic`

ptbrowne:
One could also pass a function `getInitials` to `Avatar`.

drazik:
Yep but it gives more surface to Avatar, where it must only be used to show a circle with a certain size, bgcolor/image and content.

IMHO, `Avatar` with props `text`, `bgColor` and `bgImage` is the base component that covers all the needs. If somebody wants something particular, then he/she have to implement another component with its own logic and that renders an `Avatar` underhood. Then if this component is needed elsewhere, it has to be put inside cozy-ui.

So, if we need `InitialsAvatar` only in one app for the moment (no clear plans to use it elsewhere), I think this app have to implement it using a very generic `Avatar` proposed by cozy-ui.

What do you think about it ?

ptbrowne:
Fine by me.

GoOz:
So, let me get it straight, Avatar should be just an empty recipient for image, text or nothing at all and should not take care of the text what so ever, it's the app that should process the text?
If so, should I nevertheless limit the size of the text if any? Because the avatar won't support more than 2 characters, it just crops what's too large but should I secure this or it's definitely not its job?

GoOz:
Sooooooo, I decided to not implement `color` props because we have a set of chosen colors compatible with white text color.
If we let developers choose randomly the background-color, we should have either:
- a `textColor` props but honestly this would be a pain in this ass for the developer and this wouldn't be necessarily accessible.
- a function inside the Avatar component, calculating the brightness of the given color and chose rather it's should be black or white text color, which I'm not comfrotable doint by myself.

But I removed the `getInitials` function.

ptbrowne:
Ping ^ Are we good ?

GoOz:
Now yes :)

Issue #277, opened 2017-12-22
Lato font on Styleguidist


ptbrowne:
Add Lato font on Styleguidist since otherwise it feels different than our apps.

GoOz:
Fixed by #347

PR #347, opened 2018-02-16
fix: add Lato font on Styleguidist

GoOz:
I couldn't stand the lack of Lato font on Styleguidist anymore, so I added it. 
/me feels better 😀

### DIFF
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -52,6 +52,12 @@ module.exports = {
     }
   ],
   components: '../react/**/*.jsx',
+  template: 'template.html',
+  theme: {
+    fontFamily: {
+      base: 'Lato, sans-serif'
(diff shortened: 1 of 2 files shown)


docs/styleguide.config.js
```diff
@@ -52,6 +52,12 @@ module.exports = {
     }
   ],
   components: '../react/**/*.jsx',
+  template: 'templateLato.html',
```
ptbrowne: I would just name it `template.html`.

ptbrowne:
Fixes https://github.com/cozy/cozy-ui/issues/277

docs/styleguide.config.js
GoOz: Sure, I hesitated on this more than i would admit.

PR #348, opened 2018-02-16
fix: export default button component

y-lohse:
Makes `import Button from 'cozy-ui/react/Button'` work!

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -28,6 +28,8 @@ export const Button = props => {
   )
 }
 
+export default Button
+
 export const ButtonLink = props => {
   const { theme, size, extension, href, target, className, children, onClick } = props
   return (


GoOz:
I may have missed something but why the need? and why just the Button component?

ptbrowne:
It is already the case for Icon, Modal, etc... It should work because when you import the default of a module you should get something useful, not some undefined thing. Thanks @y-lohse

GoOz:
Ok then but shouldn't we do the same for `<ButtonLink />`?

ptbrowne:
In this case we would have to put `ButtonLink` in another file since you cannot have two default exports in the same file. I am not against splitting the two. For now you can do `import { ButtonLink } from 'cozy-ui/react/Button'`

GoOz:
oh yeah that's right, only one default :P

PR #354, opened 2018-02-19
feat: modal sizes keep their size unless they don't have the room anymore

GoOz:
And furthermore, Modals get to the top of the screen instead of centered for UX pruposes.

Preview: https://gooz.github.io/cozy-ui/react/#modal

### DIFF
--- a/stylus/components/modals.styl
+++ b/stylus/components/modals.styl
@@ -10,7 +10,15 @@
 \*------------------------------------*/
 
 $modals
-    modal-width = 34rem
+    small = 34rem
+    medium = 36rem
+    large = 40rem
+    xlarge = 50rem
+    xxlarge = 60rem
+    small-padding = .5rem
+    medium-padding = 1.5rem
+    large-padding = 3rem
+
     position relative
     z-index $modal-index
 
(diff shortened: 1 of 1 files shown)


GoOz:
[Modal: max-width -> min-width](https://trello.com/c/izCx9rzw/97-modal-max-width-min-width)

GoOz:
@y-lohse Weird that the linter doesn't warn when there's unused variables…

And I can confirm that it doesn't work on Chrome or Safari, my bad… >_<

GoOz:
@y-lohse Should be ok now.

ptbrowne:
> @y-lohse Weird that the linter doesn't warn when there's unused variables…

This is because it is difficult to recognize a variable from a property name (you basically have to maintain a list of all different CSS properties and keep it up-to-date according to the new properties) in Stylus so I guess it was not implemented.

stylus/components/modals.styl
```diff
@@ -40,19 +49,34 @@ $modals
...
+        width small
+
+        @media (max-width: small + 1rem)
+            width auto
 
     .coz-modal--medium
-        max-width 36rem
+        width medium
+
+        @media (max-width: medium + 3rem)
```
ptbrowne: Where does the 3 come from ? Wouldn't it be easier to have a loop here to output all this code that is very similar.

GoOz: It's the padding of the wrapper. I guess I could use variables here too, but as for the loop, I'm not sure it would be easy to do since sizes and paddings are very contextual.

ptbrowne: And the 6rem below ?

A variable to describe the intention would be good I guess. It is difficult to understand when there are magic numbers.

GoOz: true true. 6rem = 2*3rem of padding.

I'll use variables and make the math easy to read.

GoOz:
@ptbrowne Is it OK with you like this?

ptbrowne:
Perfect, thanks !

PR #360, opened 2018-02-23
feat: forward props to textarea

y-lohse:
Just like `Input`, forwards any remaining props to the underlying element. useful for things like a `name` prop.


ptbrowne:
Thx @y-lohse it's awesome !

Issue #359, opened 2018-02-23
Hidden content on medium screens


y-lohse:
[On medium screens, cozy-ui switches the cozy-bar to `position fixed`](https://github.com/cozy/cozy-ui/blob/b5f0aa6b72af911d9ed8c0690da60460cf13bfc8/stylus/objects/layouts.styl#L11). Why is that?

It's a problem because suddenly the content of the app is *behind* the cozy-bar, so the first 50 pixels or so are invisible.

<img width="1278" alt="capture d ecran 2018-02-23 11 45 48" src="https://user-images.githubusercontent.com/2261445/36590525-3715ac20-188f-11e8-8ef1-8dfb14742017.png">

GoOz:
Ok found it. It's because I forgot something for the one column layout. It's just missing the ghost element replacing the cozy-bar in the layout.

PR #361, opened 2018-02-23
fix: One column Layout looks well on mobile

GoOz:
Extend mobile layout from 2 panes layout to One column layout.

Fix: #359

### DIFF
--- a/stylus/objects/layouts.styl
+++ b/stylus/objects/layouts.styl
@@ -64,6 +64,20 @@ $app
         overflow-x hidden
         overflow-y auto
 
+    +medium-screen()
+        display block
+
+        main,
+        main > [role=contentinfo]
+            display block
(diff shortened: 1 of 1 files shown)


stylus/objects/layouts.styl
```diff
@@ -82,15 +96,7 @@ $app-2panes-sticky
...
-        display block
-
-        main,
-        main > [role=contentinfo]
-            display block
-            overflow visible
-
-        // Those pseudo-elements are "ghost" elements replacing bar and nav
-        &:before
+        // This pseudo-element is a "ghost" element replacing nav
```
y-lohse: Extra comment here?

GoOz: Well not quite the same as the one before but it does means the same.

PR #364, opened 2018-02-27
New Modal with with elastic content

GoOz:
Kinda big PR here.

- [x] Added `$elastic` layout with header and/or footer fixed and scrollable content when there's a lack of available space. 
- [x] Refactoring Modal while using elastic layout
- [x] Clean up the existing code and using new Buttons
- [x] Added `space` props to deal with inner paddings (`small` or `large`)

Preview : https://gooz.github.io/cozy-ui/react/#modal


react/Modal/Readme.md
```diff
@@ -34,6 +47,25 @@ const sizes = [
 </div>
 ```
 
+### Space
+
+Besides the default spacing inside a Modal, you can choose another type from this two avalaible: `compressed` and `spacious`
```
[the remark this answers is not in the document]
ptbrowne: I don't understand when I would choose the spacious mode ?

Can you give some guidelines @Claiw ? I think it would be good to add them to the styleguide.

stylus/objects/layouts.styl
```diff
@@ -40,6 +41,30 @@ $fullbleed
     left 0
     right 0
 
+// Elastic Layout — Usually with fixed header and/or fixed footer and an extensible content zone with scrolling inside
+
+               // linear-gradient(alpha(coolGrey,.2) 25%, alpha(white,0)),
+               // linear-gradient(alpha(white,0), alpha(coolGrey,.2) 75%) 0 100%,
```
ptbrowne: I think those 2 comments should be deleted

stylus/objects/layouts.styl
```diff
@@ -40,6 +41,30 @@ $fullbleed
...
+    background  linear-gradient(white 30%, alpha(black, 0)),
+                linear-gradient(alpha(black, 0), white 70%) 0 100%,
+                linear-gradient(alpha(coolGrey, .25) 0, alpha(coolGrey, .25) 25%, alpha(white, 0) 26%, alpha(white, 0) 100%),
+                linear-gradient(alpha(white, 0) 0, alpha(white, 0) 74%, alpha(coolGrey, .25) 75%, alpha(coolGrey, .25) 100%) 0 100%
+    background-repeat no-repeat
+    background-color white
+    background-size 100% 2rem, 100% 2rem, 100% .5rem, 100% .5rem
+    background-attachment local, local, scroll, scroll
+    background-clip padding-box
+    overflow auto
```
ptbrowne: image

I loaded ada.description.long in the "Action" modal and it resulted in this. It lacks some padding I guess ?

ptbrowne: The margin surrounding the buttons seem too big :/

stylus/objects/layouts.styl
```diff
@@ -40,6 +41,30 @@ $fullbleed
...
+
+$elastic-content
+    background  linear-gradient(white 30%, alpha(black, 0)),
+                linear-gradient(alpha(black, 0), white 70%) 0 100%,
+                linear-gradient(alpha(coolGrey, .25) 0, alpha(coolGrey, .25) 25%, alpha(white, 0) 26%, alpha(white, 0) 100%),
+                linear-gradient(alpha(white, 0) 0, alpha(white, 0) 74%, alpha(coolGrey, .25) 75%, alpha(coolGrey, .25) 100%) 0 100%
+    background-repeat no-repeat
+    background-color white
+    background-size 100% 2rem, 100% 2rem, 100% .5rem, 100% .5rem
+    background-attachment local, local, scroll, scroll
```
ptbrowne: Can you leave a comment to say that this is used to show when the user has scrolled ?

stylus/components/modals.styl
```diff
@@ -9,146 +10,205 @@
...
+    padding small-padding
+
+    +tiny-screen()
+        padding tiny-padding
+
+$modal-footer--spacious
+    padding large-padding
+
+    +tiny-screen()
+        padding medium-padding
```
ptbrowne: Why do we have to repeat the padding for header, content, and footer ? Could we not just apply it to the entire modal ? It feels like we have a lot of code here.

react/Modal/Readme.md
```diff
@@ -34,6 +47,25 @@ const sizes = [
...
+const spaces = [
+  'compressed',
+  'spacious',
+];
+
+<div>
+  {spaces.map(space => <button onClick={()=>setState({ space, modalDisplayed: !state.modalDisplayed, })}>
+    { space }
+  </button>)}
+  {state.modalDisplayed && <Modal title={ state.space + ' modal'} space={state.space} description={content.ada.short} dismissAction={() => setState({ modalDisplayed: false })} /> }
```
ptbrowne: I would call this property `spacing` and use `large` and `small` instead of `compressed` and `spacious`. I think it is simpler and will be memorized more easily.

react/Modal/Readme.md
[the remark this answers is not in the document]
ptbrowne: For now, in the styleguide we tell what is possible. It is easy and already helpful but I think what would be even more useful would be to guide the user towards the right solution.  Here, if I were to read this, I would think "5 sizes and 3 possibles spacing mode ? That gives me 15 different possibilities to choose from when I am designing the modal. How can I choose ?".

Do you have any idea @Claiw on how to improve this ?

I guess what I'm trying to say here is that I feel we start to have too much options.

stylus/components/modals.styl
GoOz: I kept the existing strategy that would prefer adding padding to elements inside the modal than the modal itself.
So when using complex modals you don't have padding by default when you want to take advantage of all available space and also it's easier when a section with a top border should touch the borders and style have the padding inside.

react/Modal/Readme.md
GoOz: I had a hard time finding the right naming 😆 but I like yours better.

react/Modal/Readme.md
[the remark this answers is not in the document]
GoOz: This was indeed complicated to code but the Design Team seemed to need this.
Although I agree with you, maybe there should be a very highlighted default solution and let those options… very optional.

stylus/components/modals.styl
ptbrowne: OK I see 👍

PR #369, opened 2018-03-05
Add panels

ptbrowne:
### Goal

- Useful to display a main content along a side content.

### API 
- `Panel.Group`
- `Panel.Main`
- `Panel.Side`

It is showcased in the modal here : https://ptbrowne.github.io/cozy-ui/react/


stylus/components/panels.styl
```diff
@@ -0,0 +1,38 @@
...
+
+$panel-group
+    display         flex
+    flex-direction  row
+    justify-content flex-start
+    align-items     stretch
+    box-sizing      border-box
+
+$panel-main
+    flex-basis 65%
```
GoOz: You could use the shorhand here. `flex 1 0 65%`

stylus/components/panels.styl
```diff
@@ -0,0 +1,38 @@
...
+    flex-grow 1
+    flex-shrink 0
+    overflow auto
+
+$panel-side
+    flex-basis 35%
```
GoOz: If I remember correctly some browsers don't have the same default values for `shrink` and/or `grow` and you might as well write them explicitely with `flex 0 0 35%`

react/Panel/styles.styl
```diff
@@ -0,0 +1,10 @@
+@require '../../stylus/components/panels'
+
+.Panel-group
```
GoOz: Is the nesting really necessary? It doesn't do much besides inscreasing the specificity I think.

ptbrowne: You're right.

PR #373, opened 2018-03-05
fix: button children

y-lohse:
TIL: this test is useless, and even harmful, since preact creates a `<undefined>` element around the children.
[Here's](https://codesandbox.io/s/pjvwxpwjzx) a codesandbox with a demo, make sure you inspect the renderer page to see the extra element (highlight through css).

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -27,7 +27,7 @@ export const Button = props => {
       <span>
         {icon && <Icon icon={icon} />}
         {label && <span>{label}</span>}
-        {children && {children}}
+        {children}
       </span>
     </button>
   )


ptbrowne:
> TIL: this test is useless, and even harmful, since preact creates a <undefined> element around the children.

Actually, it is a bit more interesting than that and makes for a little excursion into how JSX is rendered and the internals of Preact. 

When you do `<div attr={value}>content</div>`, it is transpiled as `React.createElement('div', { attr: value }, 'content')`. And here since we use Preact, `React.createElement` is actually the [`h` function](https://github.com/developit/preact/blob/master/src/h.js). 

The `h` function returns a VNode. VNode have four attributes : 

* `nodeName`
* `key`
* `attributes`
* `children`

`h('NodeName', { key: 'myKey', className: 'cls' },  'Text')`
↓ 
 ```
VNode{
    nodeName: 'NodeName',
    key: 'myKey',
    attributes: { key: 'myKey', className: 'cls' },
    children: 'Text'
}
```

So this means that we would return `VNodes` in our `render` functions  and everything would be fine. It would be a little cumbersome but perfectly correct. Actually plain old javascript objects would also work given how simple [VNodes are](https://github.com/developit/preact/blob/master/src/vnode.js). 

`const HelloWorld = () => ({ nodeName: 'div', children: 'HelloWorld !' })`

> "Look Ma, no JSX !" said Little Bobby Tables
 

Back to our problem : when you we write `{ children && {children}}` we are actually doing `{children && { children: children }}`.  Here, the "VNode" is "incomplete" because `nodeName` is not present. But Preact does not care and when it will create the DOM node, it will use `document.createElement` that is itself pretty tolerant and will create an `<undefined>` node.

I opened https://github.com/developit/preact/issues/1029 ;)

**Comment by ptbrowne** [comment]
Everyone is encouraged to add

```
if (process.env.NODE_ENV === 'development') {
  require('preact/debug')
}
```

and 

```
    new webpack.DefinePlugin({
      'process.env': {
        'NODE_ENV': JSON.stringify('development')
      }
    }),
```

to benefit from runtime warning from preact.

CPatchane:
@ptbrowne, if you use `cozy-scripts` package it will be added by default using the new `Webpack` 4 development mode. I am still waiting the webpack 4 support for `svg-sprite-loader` to release it.

y-lohse:
Thank you for the explanation @ptbrowne ! Perfectly clear and good to know.

PR #374, opened 2018-03-07
✨ feat: mkComponent supports className when forwarding props

ptbrowne:
When forwarding className, it would shadow the className defined
when creating the Component. Now it merges classNames


ptbrowne:
ping @cpa

PR #377, opened 2018-03-13
Button improvements

### LINKED ISSUE #376: Multiple problems with the new Button API
I tried to transition the Cozy Drive codebase to the new button API, and it did not go well.

In no particular order:

- There's a bug with `ButtonLink` labels [here](https://github.com/cozy/cozy-ui/blob/master/react/Button/index.jsx#L50). It should be `{label}` and not `label`.
- We had a few instances of buttons containing an icon with other props than just the name of the icon. For example `<Icon icon="cross" width="18" color="kaki" />`. This isn't possible in the new API anymore.
- Just to be sure — using `children` is still ok and will stay so, yes? It's not deprecated?
- `label` is supposed to be a [string](https://github.com/cozy/cozy-ui/blob/master/react/Button/index.jsx#L60). Would it be possible (and would it make sense) to pass a JSX object too? Some of our buttons contain a spinner sometimes, so I don't know if it's better to use `label` or `children` for that.

ptbrowne:
Following the remarks of @y-lohse [here](https://github.com/cozy/cozy-ui/issues/376).

Fixes #376 

* ✨ feat: label can be a string or a React element
*  🐝 fix: label should be passed as variable
*  🐝 fix: children should not be passed as object
*  ✨ feat: let React element be passed for Icon

### DIFF
--- a/react/Button/Readme.md
+++ b/react/Button/Readme.md
@@ -38,8 +38,8 @@ const { Button } = require('./index');
 
 #### Width
 
-* `narrow` to ignore Button's
-* `min-width`  & `full` to enable full width
+* `narrow` to ignore Button's `min-width`
+* `full` to enable full width
 
 ```
 const { Button } = require('./index');
--- a/react/Button/Readme.md
+++ b/react/Button/Readme.md
@@ -1,5 +1,5 @@
 There's two kinds of Button at your disposal : `<Button />` & `<ButtonLink />`.
-The First is a basic `<button>` for a click event, the second is a `<a>` a link.  
+The first is a basic `<button>` for a click event, the second is a `<a>`, a link.
 Both look exactly the same, they share the same `theme`, `className` & `onClick` parameters but `<Button>` has also: 
 
 - `busy` that adds a spinner (default `false`)
@@ -9,7 +9,7 @@ when `<ButtonLink>` has:
 - `href` to add an URL (default `''`)
 - `target` to pass the link's `target` value (default `''`)
 
-#### Different themes:
+#### Themes
 
 ```
 const { Button } = require('./index');
@@ -22,7 +22,7 @@ const { Button } = require('./index');
 </div>
 ```
 
-#### Different sizes from the default one (tiny, small, large):
+#### Sizes
 
 ```
 const { Button } = require('./index');
@@ -35,7 +35,10 @@ const { Button } = require('./index');
 </div>
 ```
 
-### Different width, `narrow` to ignore Button's `min-width`  & `full` to enable full width:
+#### Width
+
+* `narrow` to ignore Button's
+* `min-width`  & `full` to enable full width
 
 ```
 const { Button } = require('./index');
@@ -67,7 +70,7 @@ initialState = { busy:false };
 <Button onClick={() => {setState(state => ({busy: !state.busy}))}} busy={state.busy} label='Toggle busy'/>
 ```
 
-#### To disable a button
+#### Disable a button
 
 ```
 const { Button } = require('./index');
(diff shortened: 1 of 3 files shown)


edas:
Seen, and no comment

y-lohse:
Thank you <3

GoOz:
Shouldn't we update the doc though? about what `icon` & `label` can expect.

react/Button/index.jsx
```diff
@@ -56,8 +56,8 @@ export const ButtonLink = props => {
 // Proptypes
 const commonPropTypes = {
   children: PropTypes.node,
-  label: PropTypes.string,
-  icon: PropTypes.string,
+  label: PropTypes.node,
+  icon: PropTypes.node,
```
GoOz: For my own culture, when the PropType is `node` but you send a `string` it doesn't "break" because a string can be related to a text node, is that it?

react/Button/index.jsx
```diff
@@ -25,9 +25,9 @@ export const Button = props => {
       onClick={onClick}
     >
       <span>
-        {icon && <Icon icon={icon} />}
+        {typeof icon === 'string' ? <Icon icon={icon} /> : icon}
```
CPatchane: It could be a breaking change here since we were able to pass the svg symbol as icon prop here, couldn't it?

ptbrowne: ?

If the icon property is a string, we pass it to `<Icon />` otherwise we render the variable so I do not think it is a breaking change.

react/Button/index.jsx
ptbrowne: https://reactjs.org/docs/typechecking-with-proptypes.html

ptbrowne:
Which doc ?

Since styleguidist will pickup the proptypes, it will be updated.

react/Button/index.jsx
CPatchane: Except that the `<Icon />` [can also take a SVG Symbol object](https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L10). This object is provided by the SVG loader when you import a SVG icon in the code. Before that, we were able to just pass this object to the Button to display using `<Icon />`, now we can't. I don't think we used a lot this way but it was just a warning about a potential breaking change.

react/Button/index.jsx
GoOz: OK TIL.
funny, out of instinct, I was expecting `node` to be elements or even text node but an array!? 
Well ok then. 
Thank you. 🙏

GoOz:
@ptbrowne Yes it will be updated, but I guess a little sentence explicitely saying what those props are expecting wouldn't hurt.
I know the code should explain itself but not everybody has the same level of understanding and if a sentence can clear any doubt I think it's not a bad thing to do.

ptbrowne:
image

GoOz:
Where is that from? Never seen this view.
And I don't think that `Icon of the button` is explicit enough. It doesn't explain how, just what, and since we have two different ways of doing it with the same prop I guess it would be good to explain it a little bit more in the readme.
I may be old fashion on this one but what's the purpose of a documentation if it only says with words what the code says and nothing more?
Do you understand my concerns or not at all? 😕

ptbrowne:
> Where is that from? Never seen this view.

As stated in my previous message, "It is also visible in the styleguide by clicking "Props and Methods".

> And I don't think that "Icon of the button" is explicit enough. It doesn't explain how, just what, and since we have two different ways of doing it with the same prop I guess it would be good to explain it a little bit more in the readme.

I've added a section.

react/Button/Readme.md
```diff
@@ -35,7 +35,10 @@ const { Button } = require('./index');
 </div>
 ```
 
-### Different width, `narrow` to ignore Button's `min-width`  & `full` to enable full width:
+#### Width
+
+* `narrow` to ignore Button's
+* `min-width`  & `full` to enable full width
```
GoOz: It should be 
```
* `narrow` to ignore Button's `min-width`
* `full` to enable full width
```

react/Button/index.jsx
ptbrowne: Good catch : https://github.com/cozy/cozy-ui/pull/377/commits/0ab5217a8618ff57b02dc37e39cbd3d4afa3600c . Is it ok ?

PR #378, opened 2018-03-13
fix: buttonlink label and children

y-lohse:
Fixes the labels and the children for `ButtonLink`


y-lohse:
Woops, i'm late to the party. Replaced by #377

PR #379, opened 2018-03-13
fix: export modal buttons

GoOz:
I guess this one was missing too, isn't it.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -40,7 +40,8 @@ const ModalHeader = ({ children, className }) =>
     </h2>
   )
 
-const ModalTitle = (props) => {
+const ModalTitle = props => {
+  console.log('ModalTitle is a deprecated component, use ModalHeader instead')
   return <ModalHeader {...props } />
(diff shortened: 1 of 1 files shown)


react/Modal/index.jsx
```diff
@@ -72,7 +72,7 @@ const ModalFooter = ({ secondaryText, secondaryAction, secondaryType, primaryTex
     )
 }
 
-const ModalButtons = () => {
+const ModalButtons = (props) => {
   console.log('ModalButtons is a deprecated component, use Modalfooter instead')
```
ptbrowne: use ModalFooter instead

react/Modal/index.jsx
```diff
@@ -72,7 +72,7 @@ const ModalFooter = ({ secondaryText, secondaryAction, secondaryType, primaryTex
     )
 }
 
-const ModalButtons = () => {
+const ModalButtons = (props) => {
```
ptbrowne: `props =>` is OK

PR #380, opened 2018-03-13
fix: svg symbol detection

y-lohse:
Unfortunate side-effect of the fix from earlier : if a button has no icon, the `icon` variable is `null` and reading `icon.id` produces a type error and crashes the whole thing :(


y-lohse:
Yeah no worries, shit happens when things are rushed.

Issue #269, opened 2017-12-20
Selector component ?


ptbrowne:
We have this component in Bank and Santé. Current behavior : 

- Menu should take all screen on mobile
- Menu should be opened from the selector on Desktop
- It should have a small chevron to the right
- There should be an Overlay when opened

So it seems it is quite like a Menu except that it has an overlay and a different behavior on mobile. What do you think @gooz ?

![image](https://user-images.githubusercontent.com/465582/34207533-2daf15cc-e58b-11e7-957b-ee94c5350638.png)

![image](https://user-images.githubusercontent.com/465582/34207548-3b41e494-e58b-11e7-85f9-53c5c1453db9.png)

I am open for a different name.

GoOz:
I don't really know this one.
Before anything I'd rather talk about it with @Claiw, but yeah it seems it could belong to UI.
Not sure about the name though.

ptbrowne:
>  Before anything I'd rather talk about it with @Claiw, but yeah it seems it could belong to UI.

👌

Claiw:
The behavior seems ok for me.
In the mockups, this component is called "Popover"; I have too a mockup named as "selection open" which describes the behavior 
Perhaps Selection-menu or Selection-list ?

goldoraf:
Regarding to the naming issue, I think what we have here is a Menu in a Popover, like here: http://www.material-ui.com/#/components/popover

y-lohse:
We have similar use-cases on Contacts:

<img width="409" alt="capture d ecran 2018-03-16 09 06 12" src="https://user-images.githubusercontent.com/2261445/37509874-7a258c7c-28f9-11e8-94c6-9713286cfcda.png">

A button with a dropdown indicator, a list of selectable items, a form at the end.

<img width="409" alt="capture d ecran 2018-03-16 09 05 55" src="https://user-images.githubusercontent.com/2261445/37509887-8bcf5520-28f9-11e8-9ffd-4bad5771bfaa.png">

Same thing, but each option has a checkbox net to it and you can pick more than one.

<img width="904" alt="capture d ecran 2018-03-16 09 09 07" src="https://user-images.githubusercontent.com/2261445/37509930-b1ec4772-28f9-11e8-99b4-b995b766a5ed.png">

An interactive option that contains buttons, and you can turn it into an editable field.

________________

[...]

PR #389, opened 2018-03-19
feat: advanced select component

y-lohse:
Following up on #269, her's my proposal for our `Select` component.

It basically exports `react-select` and lets you override it. The default options give it a Cozy look, although I'm sure there's more to be done on that front.

Should it be called `Select` though? Maybe we'd like to keep that for the regular HTML select, but I can't find a suitable alternative. `AdvancedSelect`? `ReactSelect`?

edit: live doc http://yannick-lohse.fr/cozy-ui/react/#select (there's a `z-index` problem with the code blocks but no idea where to fix it)


y-lohse:
[(3h) [UI] composant de gestion de groupes d'un contact](https://trello.com/c/ZWXLfheS/875-3h-ui-composant-de-gestion-de-groupes-dun-contact)

y-lohse:
[(5) Gérer les groupes dans la fiche de contact](https://trello.com/c/NiAXu1HW/729-5-g%C3%A9rer-les-groupes-dans-la-fiche-de-contact)

GoOz:
About the z-index issue, it may be just for the styleguide but I'll take a look when pimping the component in case it happens on real situation

GoOz:
Oh I just noticed, did we lose keyboard interactions?

react/Select/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+import PropTypes from 'prop-types'
+import classNames from 'classnames'
+import ReactSelect from 'react-select'
+import styles from './styles.styl'
+import Icon from '../Icon'
+import palette from '../../stylus/settings/palette.json'
+
+const withCustomControl = CustomContol => ({ innerProps, children }) => (
+  <div {...innerProps}>
+    {CustomContol}
```
ptbrowne: s/CustomContol/CustomControl

react/Select/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        <input
+        type="checkbox"
+        checked={isSelected}
+        className={styles['select-option__checkbox']}
+      />
+    }
+    {label}
+    { (isSelected && !withCheckbox) && <Icon icon="check-circleless" color={palette['dodgerBlue']} className={styles['select-option--checkmark']} /> }
+  </div>
+);
```
ptbrowne: Since the majority of statements here have a check on withCheckbox, I wonder if it would not be simpler to split it into two components.

react/Select/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  innerProps: PropTypes.object.isRequired,
+  withCheckbox: PropTypes.bool,
+}
+
+Option.defaultProps = {
+  withCheckbox: false
+}
+
+const CheckboxOption = ({ ...props }) => (
+  <Option {...props} withCheckbox />
```
ptbrowne: Ah OK, it is done here.

react/Select/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+);
+
+CheckboxOption.propTypes = {}
```
ptbrowne: Why not copy Option proptypes ?

src/icons.jsx
```diff
@@ -9,6 +9,7 @@ const importIcons = function () {
     'icon-album-remove',
     'icon-back',
     'icon-check',
+    'icon-check-circleless',
```
ptbrowne: Mmh, should have been the otherway around :/

icon-check and icon-check-circle

react/Select/Readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+<Select options={options} />
+```
+
+### Custom control component
+
+You can use the `withCustomControl` HOC to turn an existing, custom component into the control button for the `Select`:
+
+```
+const withCustomControl = require('../Select').withCustomControl;
```
ptbrowne: Do you think withCustomControl could be renamed reactSelectControl ? IMHO it would be clearer.

react/Select/index.jsx
y-lohse: Yeah, I kept one "main" component in order to avoid duplicating class names.

react/Select/index.jsx
y-lohse: Weeeell. I'm not even sure about the `Option proptypes`. Most of the props defined here (such as `innerProps`, `label`, ect) are actually injected by `react-select` and not by the user. The only addition is the `withCheckbox`. You're also never really supposed to instanciate these components, you just pass their definition to `react-select`.

So, should we describe the props that `react-select` uses? I'd say probably not, and then `CheckboxOption` has no props, and `Option` has only the `withCheckbox` prop. What do you think?

src/icons.jsx
y-lohse: Exactly :/ Hopefully we can do this when we next break backward compatibility.

react/Select/index.jsx
ptbrowne: Ah yes OK, sorry I was confused. I agree.

Issue #387, opened 2018-03-19
ModalHeader font size


y-lohse:
Since `ModalHeader` is a `<h2>`, elements in it inherit rules like `font-size` or `font-weight`. This is good when there's only text, but when there are buttons for example, it's not great.
Not sure what the solution is (and I managed to override what I wanted), but maybe someone can come up with a fix !

GoOz:
True, I didn't think about that when I transformed this component.
Tell me if you think I'm right but I'm guessing something like that:
* If `title` props is defined we put a `h2`with the `title` string in it.
* If  just `children` then we just got a div and you have to, as a developer, put yourself your title in a heading through the `children` (if needed).

y-lohse:
I think that would mostly work. The one situation where it wouldn't is `<ModalHeader>My Title</ModalHeader>`, but not sure that's a big problem..?

GoOz:
Well… that would be a regression indeed, the text would lose its size & weight but I can't see how to avoid this collateral damage.

ptbrowne:
I don't understand : the majority of the properties of `<Button/>` are defined absolutely in rems.

https://github.com/cozy/cozy-ui/blob/master/stylus/components/button.styl#L12

If I replace the `h2` by a `div` via the inspector the result is like that : 

![image](https://user-images.githubusercontent.com/465582/37613868-6a4d3144-2ba9-11e8-9670-28f6bff6bd98.png)

So apart from the horizontal margin which is a bit bigger (and which could be dealt with by setting an absolute size), I do not see the problem :/

Are you using elements other than Buttons that have the problem ?

y-lohse:
Mh right, maybe `Button` was a bad example. I had the problem with a [`MenuItem`](https://github.com/cozy/cozy-contacts/blob/master/src/components/Modals/ContactCardModal.jsx#L18) element.

But I don't think it's safe to assume that everything inside a `ModalHeader` will have an absolute font-size defined. Or if an element used `em` for sizing I think it would be impacted too.

GoOz:
Still, h2 tag is not supposed to have any tag as children, at least as a semantic point of view, so there's something to do here anyway.

ptbrowne:
OK I see.

To avoid breaking backward compatibility, typeof could be used to check if children is a string, and in this case wrap it in a h2. This way we have the butter and the money of the butter ;)

GoOz:
Oh yeah I didn't think of that. good call 🎉

PR #390, opened 2018-03-19
modalHeader have h2 only if title props defined

GoOz:
But still works if `children` is just a string.

Fix: #387


GoOz:
[#387: ModalHeader font size](https://trello.com/c/1PZqgWpy/119-387-modalheader-font-size)

PR #399, opened 2018-03-23
Add 0 size

kosssi:
Pour créer toutes les classes utilitaires de margin padding 0


GoOz:
@kosssi Just added some documentation for standalone CSS.
BTW, now, `.u-m-0` for margin & `.u-p-0` for padding exist along with single direction classes like `.u-mt-0` for margin-top, `.u-mb-0` for margin-bottom, etc.

Issue #263, opened 2017-12-18
doc for Alerter + easier usage


ptbrowne:
There is no doc for the Alerter component in styleguidist.

enguerran:
We must see with @goldoraf if we rewrite the `cozy-ui/react/Alerter` with `cozy-drive/photos/Alerter`. I think the later has a better API.

ptbrowne:
Yes, it seems you have a bit of debt there :/

```
 drive master  rg 'Alerter' | grep import | sort
src/drive/components/Layout.jsx:import Alerter from 'photos/components/Alerter'
src/drive/components/Layout.jsx:import { Alerter as OldAlerter } from 'cozy-ui/react/Alerter'
src/drive/components/PublicLayout.jsx:import { Alerter } from 'cozy-ui/react/Alerter'
src/drive/ducks/files/Toolbar.jsx:import { alertShow } from 'cozy-ui/react/Alerter'
src/drive/reducers/index.js:import alerterReducer from 'cozy-ui/react/Alerter'
src/photos/components/Layout.jsx:import Alerter from './Alerter'
src/photos/containers/AddToAlbumModal.jsx:import Alerter from '../components/Alerter'
src/photos/containers/AlbumPhotos.jsx:import Alerter from '../components/Alerter'
src/photos/containers/SelectionBarWithActions.jsx:import Alerter from '../components/Alerter'
src/photos/ducks/albums/components/AlbumToolbar.jsx:import Alerter from '../../../components/Alerter'
src/photos/ducks/albums/components/PhotosPicker.jsx:import Alerter from '../../../components/Alerter'
src/photos/ducks/upload/index.js:import Alerter from '../../components/Alerter'
src/photos/reducers/index.js:import alerterReducer from 'cozy-ui/react/Alerter'
src/sharing/components/ShareByEmail.jsx:import Alerter from 'photos/components/Alerter'
```

We have to implement some Alerts in our sprint so I guess it will be a good time for us to take the Alerter from Photos and put it in cozy-ui. It's true that the API from photos/Alerter is nicer.

To me the old `Alerter` should not be used as it needs the user to install a custom state slice in his redux reducer. I guess this is why you refactored it in Photos.

What do you think if I import photos' Alerter in cozy-ui as Flasher and then deprecate Alerter ? I guess the action creators from cozy-ui/Alerter could even call `Flasher.{success,info,error}` for a smoother transition path.

goldoraf:
[...]

PR #400, opened 2018-03-26
Alerter ❗️

y-lohse:
Following up on : #263, I've essentially copied over the Alerter component from Photos, updated it and added some documentation. [Here's the preview!](http://yannick-lohse.fr/cozy-ui/react/#alerter)

### DIFF
--- a/react/Alerter/index.jsx
+++ b/react/Alerter/index.jsx
@@ -1,142 +1,179 @@
 import React, { Component } from 'react'
-import classNames from 'classnames'
-import { connect } from 'react-redux'
-
-import { translate } from '../I18n'
+import PropTypes from 'prop-types'
+import Portal from 'preact-portal'
 
 import styles from './styles.styl'
+import classNames from 'classnames'
 
-const ALERT_SHOW = 'ALERT_SHOW'
-const ALERT_DISMISS = 'ALERT_DISMISS'
-const ALERT_CLEAR = 'ALERT_CLEAR'
+const createStore = () => {
+  const notifications = []
+  const listeners = []
 
-const DEFAULT_AUTOCLOSE_DELAY = 3500
-const DEFAULT_ALERT_LEVEL = 'info'
-const DISMISS_DELAY = 500
-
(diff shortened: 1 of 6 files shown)


stylus/components/button.styl
```diff
@@ -207,8 +206,10 @@ $button-alert--info
 
     &:active
     &:hover
+    &:not([disabled]):not([aria-disabled=true]):hover
```
y-lohse: This is not cool, but there's another selector like that earlier in the file that forces a blue background on all buttons. Any recommendation @GoOz ?

y-lohse:
[(1) Ajouter une modale de confirmation lors de la suppression d'un contact](https://trello.com/c/pnIW5A5u/823-1-ajouter-une-modale-de-confirmation-lors-de-la-suppression-dun-contact)

react/Alerter/index.jsx
```diff
@@ -1,142 +1,203 @@
...
+import PropTypes from 'prop-types'
+import Portal from 'preact-portal'
 
 import styles from './styles.styl'
+import classNames from 'classnames'
 
-const ALERT_SHOW = 'ALERT_SHOW'
-const ALERT_DISMISS = 'ALERT_DISMISS'
-const ALERT_CLEAR = 'ALERT_CLEAR'
+const createStore = () => {
```
enguerran: Can't we use redux and react-redux? Or something like that https://github.com/developit/state-machine-component?

ptbrowne:
Great, thanks yannick ! I wondered why the transition was a bit choppy and saw that top/bottom are used for the transition. Can you change them to a transform ? It will be much smoother.

https://csstriggers.com/

> Changing top alters the geometry of the element. That means that it may affect the position or size of other elements on the page, both of which require the browser to perform layout operations.

Once those layout operations have completed any damaged pixels will need to be painted and the page must then be composited together.

> Changing transform does not trigger any geometry changes or painting, which is very good. This means that the operation can likely be carried out by the compositor thread with the help of the GPU.

PR #401, opened 2018-03-26
feat: add Checkbox component

GoOz:
Added `<Checkbox />` component

Preview: https://gooz.github.io/cozy-ui/react/#checkbox


GoOz:
[Add Input Checkbox component](https://trello.com/c/xudXmQhI/126-add-input-checkbox-component)

stylus/components/forms.styl
```diff
@@ -275,6 +275,71 @@ $input-text--medium
...
+            height checkbox-size
+            border-radius .125rem
+
+        &::before
+            transition box-shadow 350ms cubic-bezier(0, .89, .44, 1)
+            box-shadow inset 0 0 0 .125rem alpha(charcoalGrey, .25)
+            background-color white
+
+            &:hover
+                box-shadow inset 0 0 0 .125rem dodgerBlue
```
ptbrowne: I am not able to make the hover work. I guess the hover should be put on the parent ?

stylus/components/forms.styl
```diff
@@ -275,6 +275,71 @@ $input-text--medium
...
+                opacity 1
+                transform scale(1)
+
+        &:not(:checked) + span::after
+            opacity 0
+            transform scale(0)
+
+    &.is-error input[type=checkbox]:not(:checked) + span
+        color pomegranate
+
```
ptbrowne: If I understand well, a checkbox can be in error only if it's not checked ? I think it should be possible to have it in error even if it is checked.

GoOz: Well I asked myself the question and I couldn't find any usecase when checking a checkbox could be an error. Do you think of one?
Like this it's more of a security but I can change this is if needed.

ptbrowne: No I don't have a precise usecase here, but I find the lack of symmetry between the two cases surprising.

GoOz: Yeah I know what you mean, it's just that having a checked checkbox in red just bothered me.
Although if checked, the prop `error` should be removed by JS, so I guess it won't happen in real life.

Issue #402, opened 2018-03-26
:hover inheritance on buttons


y-lohse:
As noted in [this comment](https://github.com/cozy/cozy-ui/pull/400/files#r177103966), il I have a class that extends `$button` and `$button-alerter`, I can't simply change the background color with a `&:hover` rule, because [this rule](https://github.com/cozy/cozy-ui/blob/master/stylus/components/button.styl#L56) takes precedence.

PR #409, opened 2018-03-29
fix: button :hover styles are simpler to overwrite

GoOz:
Fix: #402


GoOz:
[#402: :hover inheritance on buttons](https://trello.com/c/Ox5RUY1x/129-402-hover-inheritance-on-buttons)

PR #410, opened 2018-03-29
fix: input radio size

GoOz:
As discussed with the designers, checkbox and radio should have the same size.

### DIFF
--- a/stylus/components/forms.styl
+++ b/stylus/components/forms.styl
@@ -9,7 +9,6 @@
 @require '../components/button'
 
 checkbox-size = 1rem
-radio-size = .8125rem
 
 $form
     .coz-form
@@ -344,12 +343,6 @@ $radio
     @extend $checkbox
     span
         &::before
-        &::after
-            top .15rem
-            width radio-size
-            height radio-size
-
-        &::before
             border-radius 50%
 
         &::after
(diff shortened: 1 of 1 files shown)


GoOz:
[Add Input Radio component](https://trello.com/c/AzWaVK7n/127-add-input-radio-component)

stylus/components/forms.styl
```diff
@@ -357,8 +350,8 @@ $radio
             content '•'
             color white
             text-align center
-            font-size .7rem
-            line-height .7rem
+            font-size 1rem
+            line-height .9rem
```
ptbrowne: Why is the line-height different from the font-size ? The line-height does not round to pixels cleanly (0.9*16 = 14.4px) , I suggest to change it to 0.9375rem or 0.875rem.

GoOz: Well, browsers round up but you're right, it'll be cleaner.

ptbrowne: > Well, browsers round up

All browsers do not have the same strategy for rounding up or down, and it seems some css properties have subpixel rendering.

Source: https://stackoverflow.com/questions/34676263/sub-pixels-calculated-and-rendered-differently-among-browsers and any "subpixel rendering" search in Google

PR #411, opened 2018-03-30
feat: add disabled attribute for ButtonLink ✨

gregorylegarec:
This PR adds a `disabled` attribute for `<ButtonLink />`. In a few case we may display a `<ButtonLink />` having an `href` fetched asynchronously. Or an `href` with no value. In those case, the `<ButtonLink />` could be displayed as disabled an the `onClick` handled ignored.

Example : in [Cozy-Collect](http://github.com/cozy/cozy-collect/) we are fetching the cozy-store URL. In some environment this app may not be installed, so the `<ButtonLink />` is disabled.

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -43,13 +43,14 @@ export const Button = props => {
 export default Button
 
 export const ButtonLink = props => {
-  const { theme, size, extension, href, target, className, children, label, icon, onClick } = props
+  const { disabled, theme, size, extension, href, target, className, children, label, icon, onClick } = props
   return (
     <a
+      aria-disabled={disabled}
       href={href}
       target={target}
       className={btnClass(theme, size, extension, className)}
-      onClick={onClick}
+      onClick={disabled ? (event) => event.preventDefault() : onClick}
     >
       <span>
         {usableByIcon(icon) ? <Icon icon={icon} /> : icon}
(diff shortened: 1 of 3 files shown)


react/Button/index.jsx
```diff
@@ -43,13 +43,15 @@ export const Button = props => {
 export default Button
 
 export const ButtonLink = props => {
-  const { theme, size, extension, href, target, className, children, label, icon, onClick } = props
+  const { disabled, theme, size, extension, href, target, className, children, label, icon, onClick } = props
   return (
     <a
+      aria-disabled={disabled}
+      disabled={disabled}
```
ptbrowne: The disabled attribute is not valid for anchors. https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes

gregorylegarec:
@ptbrowne Well spotted, thanks 👍

ptbrowne:
Since Button and ButtonLink now are so similar, I think it would be simpler if Button could support an href. ButtonLink would then be superfluous.

gregorylegarec:
@ptbrowne Seems legit. Do you think it should be done in this PR ? The tasks seems to be : 
[] Add an `href` attribute to `<Button />`
[] Deprecate `<ButtonLink />` but keep it "internally" ?

ptbrowne:
No, I do not think it should be done in this PR.

Yes I agree with the checklist, I guess the Button should switch to render `<a>` tags when it has an href.

GoOz:
That's the way I did it at first… one and only component to deal with both use cases ¬_¬

PR #413, opened 2018-04-03
feat: add Empty component

GoOz:
Added Empty component.
preview: https://gooz.github.io/cozy-ui/react/#empty

⚠️ This PR has some breaking changes towards $empty styles used so far. So wait before merge.


ptbrowne:
I think it would be best if Empty reused other standard components instead of defining its own title, text etc...

GoOz:
But we don't have standard components for this yet.

ptbrowne:
So let's write them ?

ptbrowne:
The goal of the styleguide is to provide coherence, it will be more difficult to have coherence by doing top-down than doing bottom up IMHO.

GoOz:
I have another card for the titles to do later. But for the paragraphs I don't have guidelines yet.

CPatchane:
IMHO the purpose of this PR is to add a new `<Empty />` component. Since the standards text and title components don't exist for now, it should be done in another PR without blocking this one instead.
If we add subcomponents after this PR, it shouldn't be a breaking change but an internal change (like my previous PR about adding IntentModal component for example). So final user of the `<Empty />` won't be affected so.

ptbrowne:
Fine, this will not be the first time we do things in the wrong order.

nono:
What is the next step for this PR?

GoOz:
@nono It has to be reworked, hopefully next sprint

GoOz:
[Add Empty component](https://trello.com/c/AhfPr2He/132-add-empty-component)

ptbrowne:
> ⚠️ This PR has some breaking changes towards $empty styles used so far. So wait before merge.

Since it has breaking changes, do you have advice on how to adapt existing Empty components ? What are the breaking changes ?

GoOz:
@ptbrowne Our apps use the `$empty` placeholder which drastically changed. So it **will** break.
Apps should use this component instead. 
But I'll do it, later. I'm holding this PR until next week and take care of it myself then.

CPatchane:
If it will add breaking changes, please bump a major version for this.

GoOz:
@CPatchane will do, I promise 😛

PR #415, opened 2018-04-05
✨ feat: able to customize delay of alert and auto delay

ptbrowne:
- [x] Ability to customize the duration before closing the alert
- [x] If no duration is passed, the duration is computed from the number of words in the message

### DIFF
--- a/react/Alerter/index.jsx
+++ b/react/Alerter/index.jsx
@@ -29,7 +29,7 @@ class Alert extends Component {
     hidden: true
   }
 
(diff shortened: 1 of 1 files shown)


ptbrowne:
> I wondered if we should maybe have a minimum duration too, but even with 3 words it's already one second so it's probably fine.

There is a Math.max : the duration has already a minimum duration of 2 seconds.

y-lohse:
Ah yes of course 👍 . Not quite awake yet.

PR #416, opened 2018-04-05
🔧 chore: update react-styleguidist

ptbrowne:


docs/styleguide.config.js
```diff
@@ -1,4 +1,5 @@
 const path = require('path')
+const fs = require('fs')
```
GoOz: I don't see what it is used for.

ptbrowne:
There are some things that would be interesting to test : 

* The pagePerSection setting if the styleguidist is too long : https://github.com/styleguidist/react-styleguidist/releases/tag/v6.5.0

* webpack 4 support

PR #419, opened 2018-04-09
feat: Add MidEllipsis component

GoOz:
Added MidEllipsis component to enable an ellipsis in the middle of a string.

preview: https://gooz.github.io/cozy-ui/react/#midellipsis
Example inside `<ListItemText />` component: https://gooz.github.io/cozy-ui/react/#listitemtext


GoOz:
[Middle Ellipsis Component](https://trello.com/c/yuS3Pe0Q/137-middle-ellipsis-component)

GoOz:
@ptbrowne I'm open to suggestions.
I can't fade because `text-overflow: fade` isn't supported by browsers or use a pseudo element width a gradient because it would mean setting a background color to this component which won't be generic. I don't have other solutions. Do you think of something I missed?

ptbrowne:
Non mais OK ca marche, pas de soucis.

PR #420, opened 2018-04-09
fix: fill strip loader lowercase

y-lohse:
Some icons (such as `rename`) were not properly stripped of their fill color because they use lower-case color values.


y-lohse:
[Correction d'un bug de couleurs d'icônes dans cozy-ui](https://trello.com/c/MkA8qXRZ/923-correction-dun-bug-de-couleurs-dic%C3%B4nes-dans-cozy-ui)

PR #421, opened 2018-04-09
feat: extract IntentIframe from IntentModal 🔨

gregorylegarec:
This PR extracts the `IntentIframe` from the `IntentModal`, for cases when the intent should not be opened in a modal. That is for example the case in Collect when it performs an intent from an intent.


gregorylegarec:
[(5) [COL] ETU je peux installer/configurer UN connecteur depuis de l'inter-apps](https://trello.com/c/PnUUxk1F/898-5-col-etu-je-peux-installer-configurer-un-connecteur-depuis-de-linter-apps)

react/IntentIframe/styles.styl
```diff
@@ -0,0 +1,17 @@
+iframe
+    width 100%
+    height 100%
+    border 0
+
+
+.intentIframe
+    height 100vh
```
GoOz: Watch out, `100vh` on Safari/Chrome is actually longer than 100% of the viewport's height. 
Sometimes it doesn't render as you wished for. Did you test it on those browsers, preferably on mobile?

gregorylegarec: Thanks, Is using 100% safer ?

gregorylegarec:
@GoOz Thanks, I renamed the markdown file. Isn't weird to use a capital as first letter for all `Readme.md`?

react/IntentIframe/styles.styl
GoOz: well yeah but a little more verbose as you need this element's parents to have their height set to `100%` too but I guess that's already the case, so it's good. :)

GoOz:
@gregorylegarec I… maybe yes… I'm just a fan of consistency and this uppercase doesn't bother me that much but I guess you're right.
Indeed, other intent components' readme are named the same way… I don't know.
The way I see it, it should be `readme.md` by default when there's only one component file in the folder and maybe renamed to the according filename when there's a bunch of them.

But don't bother too much with this, this is non-blocking stuff.

PR #422, opened 2018-04-12
feat: hide overflowing content from Bd component

drazik:
This just add `overflow: hidden` in the `Bd` component.

In general, since this component must constraint its content's width, I think it should hide what's overflowing.

More particulary, in banks we use the `ListItemText` component inside a `Bd` component. Since the `ListItemText` has `white-space: nowrap`, it expands to fit the wider line of text is contains. But in that case, if `Bd` doesn't have `overflow: hidden`, its content may overflow : 

image

You may think this is a particular case we have here. But on the contrary I think it's very generic : `Bd` must fill the available space, and what it contains must never overflow.

### DIFF
--- a/react/Media/Media.styl
+++ b/react/Media/Media.styl
@@ -10,6 +10,7 @@
 
 .bd
     flex 1 1 auto
+    overflow hidden
 
 .img
     line-height 0


GoOz:
@drazik Isn't the `<ListItemText />` supposed to have ellipsis by default? It shouldn't render this way to begin with. Did you put the props to false? Or did I messed up somewhere?
I'm not sure adding an `overflow: hidden` to `<Bd>` would hurt but I think there's something to look inside `<ListItemText />` first.

drazik:
I'm using `ListItemText` in the most simple way : 

```
<ListItemText primaryText={operation.label} secondaryText={operation.account.label} />
```

`ListItemText` has `text-overflow: ellipsis` and therefore theorically everything should be good. But I think because `Bd` have neither `overflow: hidden` nor a fixed `width` or `max-width` it can grow indefinitely, so it takes the width of the text. Then `ListItemText` takes 100% of the computed width, and from the point of view of the browser don't overflows.

I don't think the problem is just with the `ListItemText` component, but more with the context in which I use it. It should constrain its width, and it doesn't.

GoOz:
@drazik ah ok, I see your point. 
The thing is if you `overflow: hidden` on `Bd` you'll never get the ellipsis.
Isn't it best to "control" the width of `Bd` instead ? Like a `width: 100%`?
Sorry to be a pain in the ass, it's just that having `overflow: hidden` by default seems very strict to me and if it doesn't give us an issue now, it could in the future.

drazik:
Actually, by adding a `overflow: hidden` on the `Bd`, I get the ellipsis : 

image

My first attempt to solve the issue was to set a `max-width: 100%` on the `Bd`. But in that case, if it takes `100%` on its parent width, it still overflows by the width of its sibling `Img`. Another solution would be to compute the available space in JS and dinamically set the width. But it's very ugly.

That's why I think everything that overflows from `Bd` should just be hidden.

GoOz:
Ok then. That must be a way more complicated context than I thought. 
Go for it. We'll deal with overflow issues if and when it come. 😄

PR #423, opened 2018-04-13
feat: add ButtonAction component

GoOz:
Added `<ButtonAction />` component

preview: https://gooz.github.io/cozy-ui/react/#buttonaction


react/ButtonAction/styles.styl
```diff
@@ -0,0 +1,26 @@
...
+
+.c-actionbtn-label
+    @extend $actionbtn-label
+
+.c-actionbtn-icon
+    @extend $actionbtn-icon
+
+
+
+
```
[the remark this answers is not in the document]
GoOz: indeed. 😊

react/ButtonAction/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+import cx from 'classnames'
+import PropTypes from 'prop-types'
+import styles from './styles.styl'
+import Icon from '../Icon'
+
+const usableByIcon = icon => {
+  const isSvgSymbol = icon && !!icon.id
+  const isIconIdentifier = typeof icon === 'string'
+  return isSvgSymbol || isIconIdentifier
+}
```
ptbrowne: I think this should be the responsibility of Icon to check fo this and not crash if there is a problem. Otherwise this helper should be used in every possible component that uses `Icon`.

react/ButtonAction/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+        className)}
+      onClick={onClick}
+      title={compact ? label : undefined}
+      {...restProps}
+    >
+      <span>
+        {usableByIcon(leftIcon) ? <Icon icon={leftIcon} /> : leftIcon}
+        {label &&
+          <span
+            data-action="label"
```
ptbrowne: I see that data-action is used to target this span in CSS, why not using a class ?

react/ButtonAction/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+          </span>
+        }
+        {children}
+      </span>
+    </button>
+  )
+}
+
+ButtonAction.propTypes = {
+  type: PropTypes.string.isRequired,
```
ptbrowne: a `oneOf` could be used here.

Since `type` has a default, it is not really required here, ie the user does not have to provide it.

stylus/components/button.styl
```diff
@@ -348,3 +350,109 @@ $button--full
...
 $button--narrow
     min-width auto
+
+/*------------------------------------*\
+  Action button
+\*------------------------------------*/
+$actionbtn
+    @extend $button
+    border-color silver
+    text-transform none
```
ptbrowne: Why is it necessary ? Isn't it the default ?

stylus/components/button.styl
```diff
@@ -348,3 +350,109 @@ $button--full
...
+    max-width 12.5rem
+    min-height 2rem
+    width 100%
+    padding-right .5rem
+    text-align left
+
+    > span
+        justify-content flex-start
```
ptbrowne: For my culture, why is it necessary to add ?

react/index.js
```diff
@@ -1,5 +1,6 @@
 export {default as Alerter} from './Alerter'
 export { Button, ButtonLink } from './Button'
+export { default } from './ButtonAction'
```
kosssi: Il y a une erreur ici @GoOz:
```
export { default ad ButtonAction } from './ButtonAction'
```

stylus/components/button.styl
GoOz: because I `@extend $button` which overwrite the default value

stylus/components/button.styl
GoOz: nope, because I `@extend $button` which overwrite the default value :)

react/ButtonAction/index.jsx
GoOz: Yeah I know, it seems odd, but I try very hard not using classes in stylus `$placeholder` so it isn't forced classes and it reduces the risk of name conflict. 
The thing is I cannot make parent > children relationship between placeholders but for this one I really needed it, so I add something as neutral as possible.

react/ButtonAction/index.jsx
GoOz: I do agree, I took that form the `<Button />` component and it felt redundant.

react/ButtonAction/index.jsx
ptbrowne: > I try very hard not using classes in stylus $placeholder so it isn't forced classes

ptbrowne:
Juste après le click on a : 

image

c'est un peu bizarre :/

GoOz:
ah j'ai oublié de faire du `:focus`, bien vu.

ptbrowne:
Est ce qu'on veut le -webkit-outline ?

GoOz:
@patrick On m'a jamais dit au design de l'enlever et quand on me dit pas, je me fais un plaisir de laisser les fonctionnalités d'accessibilité par défaut du navigateur.

PR #427, opened 2018-04-17
Feat: various minor tweaks and additions

GoOz:
- [x] Added `fullwidth` props on `<Input />` component 
- [x] Added `fullwidth` props on `<Textarea />` component
- [x] Added fullwidth style for default `select` element
- [x] Ensure font-family on form elements is respected
- [x] Increase modal fullscreen's selector specificity

### DIFF
--- a/react/Input/index.jsx
+++ b/react/Input/index.jsx
@@ -4,7 +4,7 @@ import PropTypes from 'prop-types'
 import styles from './styles.styl'
 
 const Input = props => {
-  const { type, id, className, value, placeholder, error, size, ...restProps } = props
+  const { type, id, className, value, placeholder, error, size, fullwidth, ...restProps } = props
   return (
     <input
       type={type}
@@ -12,7 +12,8 @@ const Input = props => {
       className={cx(
         styles['c-input-text'], {
           [styles[`is-error`]] : error,
-          [styles[`c-input-text--${size}`]] : size
+          [styles[`c-input-text--${size}`]] : size,
+          [styles[`c-input-text--fullwidth`]] : fullwidth
         },
         className)}
       placeholder={placeholder}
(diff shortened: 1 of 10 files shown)


react/Textarea/Readme.md
```diff
@@ -23,6 +23,16 @@
 </div>
 ```
 
+#### Full width textarea
+
+```
+<div>
+  <p>
+    <Textarea placeholder="I'm have a tiny size" fullwidth></Textarea>
```
ptbrowne: same here, let's leverage the size attribute ?

The placeholder should be changed to I'm full width

react/Input/index.jsx
```diff
@@ -4,15 +4,16 @@ import PropTypes from 'prop-types'
...
   return (
     <input
       type={type}
       id={id}
       className={cx(
         styles['c-input-text'], {
           [styles[`is-error`]] : error,
-          [styles[`c-input-text--${size}`]] : size
+          [styles[`c-input-text--${size}`]] : size,
+          [styles[`c-input-text--fullwidth`]] : fullwidth
```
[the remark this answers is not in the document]
GoOz: Well no, `size` is mostly about the padding inside and the height, `fullwidth` is just about the width.
You can have both props on a component.

stylus/components/forms.styl
```diff
@@ -384,6 +384,15 @@ $select
...
         display  none
 
+$select--tiny
+    @extend $input-text--tiny
+
+$select--medium
+    @extend $input-text--medium
+
+$select--fullwidth
+    @extend $input-text--fullwidth
```
ptbrowne: Why do we need the extra level of indirection ? Why can't the select directly extends from input-text--fullwidth ?

PR #428, opened 2018-04-17
🚑  fix: avoid IntentModal dismissAction to be called twice

gregorylegarec:
Spent one hour tonight dealing with this.

In the future we shoud improve this fix, by separating the logic between modal closing and intent cancellation. But it implies deprecating properties like `dismissAction`, ensure legacy and avoid regression.

I think we should wait for the release of cozy-interapp to make Cozy-UI components up to date and improve this fix.

### DIFF
--- a/react/IntentModal/IntentModal.jsx
+++ b/react/IntentModal/IntentModal.jsx
@@ -1,8 +1,11 @@
 import React, { Component } from 'react'
-import Modal from '../Modal'
-import IntentIframe from '../IntentIframe'
 import styles from './styles.styl'
 import PropTypes from 'prop-types'
+
+import once from 'lodash/once'
+
+import IntentIframe from '../IntentIframe'
+import Modal from '../Modal'
 
 /**
  * Render a modal for the specified intent.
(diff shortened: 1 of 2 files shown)


react/IntentModal/IntentModal.jsx
```diff
@@ -9,36 +9,59 @@ import PropTypes from 'prop-types'
...
+  dismiss = () => {
+    const { status } = this.state
+    if (status === 'dismissed') return
+
+    const { dismissAction } = this.props
+    this.setState({
+      status: 'dismissed'
+    })
+    dismissAction && dismissAction()
+  }
```
ptbrowne: It would be better to use _.once for this usecase, it boils down to : 

```
const once = fn => {
    const called = false
    let result
	return function () {
      if (called) { return result }
      called = true
      result = fn.apply(this, arguments)
      return result
	}
}

dismiss = once(() => {
```

gregorylegarec: @ptbrowne Well spotted, thanks !

ptbrowne:
What do you mean by "more fix needs deprecating some props"

react/IntentModal/IntentModal.jsx
```diff
@@ -1,44 +1,64 @@
-import React from 'react'
-import Modal from '../Modal'
-import IntentIframe from '../IntentIframe'
+import React, { Component } from 'react'
 import styles from './styles.styl'
 import PropTypes from 'prop-types'
 
+import { once } from 'lodash'
```
CPatchane: Could you, at least, use directly `lodash/once` here? We should install only lodash helpers necessary in the library and not globalize the lodash whole usage bin adding the `lodash` whole package IMHO.

PR #430, opened 2018-04-18
Fixed swipe down on ActionMenu

goldoraf:
The gestures handler must be attached to the wrapper node instead of the menu itself, so that the user can swipe from wherever he wants on the screen (today it's necessary to begin the swipe on the menu).


goldoraf:
[(5) Trier les fichiers dans Drive](https://trello.com/c/TgqIiJfX/869-5-trier-les-fichiers-dans-drive)

PR #431, opened 2018-04-18
feat: add 'into' prop on IntentOpener and IntentModal

drazik:
Since `Modal` now has a `into` prop to render it in a portal somewhere in the DOM instead of in a direct child, we can use it on the `IntentOpener` and `IntentModal`.


ptbrowne:
LGTM, just update the snapshot for Travis :)

PR #434, opened 2018-04-18
feat: remove dimensions and border on compact version of ButtonAction

drazik:
I updated the compact version of the `ButtonAction` component to make it match the requirements banks.

It means : 

* Remove the border on the left (we need to handle it at the application level since even if a transaction row does'nt have an action, it shows this border)
* Remove specific size for the compact variant

See https://drazik.github.io/cozy-ui/react/#buttonaction , and the result in banks : 

image

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -379,10 +379,7 @@ $actionbtn
 $actionbtn--compact
     @extend $button--narrow
     position relative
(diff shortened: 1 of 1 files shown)


drazik:
Yeah I know you didn't have so many spec, that's why I did the changes since I know the context of the app in which we'll use it ;)

PR #436, opened 2018-04-18
bug: Ensure legacy by keeping exporting Alerter

gregorylegarec:


gregorylegarec:
Actually maintaining legacy is far more complicated than just exporting the Alerter, and it's useless too maintain it as the only apps impacted are internal apps.

PR #437, opened 2018-04-19
feat: add Badge component

drazik:
This adds a `<Badge />` component that can wrap any other component to add a little circle with some content to it.

See https://drazik.github.io/cozy-ui/react/#badge

### DIFF
--- a/stylus/components/badge.styl
+++ b/stylus/components/badge.styl
@@ -1,6 +1,6 @@
 @require '../settings/palette'
 
-radius = .4375rem
+size = .875rem
 
 $badge-root
     position relative
@@ -8,10 +8,10 @@ $badge-root
 
 $badge
     position absolute
-    bottom - radius
-    right - radius
-    width radius * 2
-    height radius * 2
+    bottom - size / 2
+    right - size / 2
+    width size
+    height size
     border-radius 50%
     background-color black
     color white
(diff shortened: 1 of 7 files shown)


stylus/components/badge.styl
```diff
@@ -0,0 +1,38 @@
+@require '../settings/palette'
+
+radius = .4375rem
```
GoOz: I was confused about the name of this `radius` variable and saw that it's not used for radius at all.
I wonder if it wouldn't be best to name it something like `size` and prefer something like this:
```
size = .875rem
[…]
bottom - size / 2
right - size / 2
width size
height size
```

stylus/components/badge.styl
```diff
@@ -0,0 +1,38 @@
...
+    color white
+    z-index 1
+    display flex
+    flex-direction row
+    flex-wrap wrap
+    justify-content center
+    align-content center
+    align-items center
+    font-size .625rem
+    border 1px solid white
```
GoOz: Even for border, I prefer using `rem` so  it should be `.0625rem` even if, I know, it isn't read easily. :)

stylus/components/badge.styl
drazik: Yeah the idea behind `radius` was more the radius of the circle. I understand it can be confusing, I change it.

stylus/components/badge.styl
drazik: Yeah I missed this one

PR #440, opened 2018-04-19
chore: Use prettier formatting

gregorylegarec:
theshining-1


.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": ["cozy-app"],
+  "rules": {
+    "no-console": "off",
+    "react/display-name": "off"
```
gregorylegarec: A lot of occurrences of these two rules, so I preferred to keep them for now, as their usages are convenient.

PR #445, opened 2018-04-23
feat: add 3px margin on caption inside ListItemText

drazik:
Following @joel-costa review, I added 3px top margin on the captions inside `ListItemText`

### DIFF
--- a/stylus/components/list.styl
+++ b/stylus/components/list.styl
@@ -5,3 +5,6 @@ $list-text
     flex-direction column
     justify-content center
     min-height 2.5rem
+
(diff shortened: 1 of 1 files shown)


stylus/components/list.styl
```diff
@@ -5,3 +5,6 @@ $list-text
     flex-direction column
     justify-content center
     min-height 2.5rem
+
+    .g-caption
+        margin-top 3px
```
GoOz: `.1875rem` instead of `3px` and we're good.

Issue #447, opened 2018-04-23
span on ButtomAction is truncate


kosssi:
On peut voir sur https://cozy.github.io/cozy-ui/react/#buttonaction que la version longue est tronqué en bas

![capture d ecran 2018-04-23 a 17 19 29](https://user-images.githubusercontent.com/1135513/39136226-0858d97e-471b-11e8-9cff-b021fb95eecd.png)

GoOz:
ooooh good catch, I didn't see it. Thanks.

PR #448, opened 2018-04-23
fix: buttonAction line height

GoOz:
FIX: #447


GoOz:
[#447: span on ButtomAction is truncate](https://trello.com/c/32ojxrok/148-447-span-on-buttomaction-is-truncate)

PR #450, opened 2018-04-24
WIP 🚨 test: auto test from Readme.md

ptbrowne:
- Automatic snapshot from the Readme of the Button

More to follow !

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,5 +3,9 @@
   "rules": {
     "no-console": "off",
     "react/display-name": "off"
+  },
+  "globals": {
+    "mount": true,
+    "shallow": true
   }
 }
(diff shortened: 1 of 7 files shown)


gregorylegarec:
If I understand well, the goal is to automatically generate a snapshot in the styleguidist page ?

It looks cool but do we agree that it cannot replace jest snapshots because Travis cannot run styleguidist tests (as far as I know) ?

ptbrowne:
> because Travis cannot run styleguidist tests (as far as I know)

It is a normal Jest test generated from the Readme.md so Travis will be able to run it.

ptbrowne:
The snapshot is here ;) https://github.com/cozy/cozy-ui/pull/450/files#diff-e435f7015f0a11b9538c80ba6faeaaec

CPatchane:
Oops thanks 🎉

PR #455, opened 2018-04-30
Add chip

ptbrowne:
image

Need it to implement 

image

Preview at https://ptbrowne.github.io/cozy-ui/react/#chip

Also fixed Avatar to be able to use `style` and to have different background color even when initials are the same.


ptbrowne:
😂😂😂

CPatchane:
Can we add this new components in the css build as well?

PR #456, opened 2018-04-30
Fix toggle state return from onToggle + add tests

CPatchane:
The toggle was always returning `!props.checked` instead of returning the real new toggle state.

### DIFF
--- a/react/Toggle/index.jsx
+++ b/react/Toggle/index.jsx
@@ -4,9 +4,9 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
 class Toggle extends Component {
-  onChange() {
+  onChange(e) {
     if (this.props.onToggle) {
-      this.props.onToggle(!this.props.checked)
+      this.props.onToggle(e.target.checked)
     }
   }
   render() {
--- a/react/Toggle/test/toggle.spec.js
+++ b/react/Toggle/test/toggle.spec.js
@@ -1,0 +1,26 @@
+'use strict'
+/* eslint-env jest */
+
+import React from 'react'
+import { shallow } from 'enzyme'
+
+import Toggle from '../'
+
+describe('Toggle component', () => {
+  it(`handles correctly onToggle function`, () => {
+    const handleOnToggle = jest.fn()
+    const component = shallow(
+      <Toggle id="mock-toggle-id" onToggle={handleOnToggle} />
+    )
+    component
+      .find('#mock-toggle-id')
+      .simulate('change', { target: { checked: true } })
+    expect(handleOnToggle.mock.calls.length).toBe(1)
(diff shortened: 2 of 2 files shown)


react/Toggle/index.jsx
```diff
@@ -4,9 +4,9 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 
 class Toggle extends Component {
-  onChange() {
+  onChange(e) {
     if (this.props.onToggle) {
-      this.props.onToggle(!this.props.checked)
+      this.props.onToggle(e.target.checked)
```
CPatchane: Maybe it was used in another way by changing the props from the parent. I can't know for now.

react/Toggle/test/toggle.spec.js
```diff
@@ -0,0 +1,52 @@
...
+    expect(component).toMatchSnapshot()
+  })
+
+  it(`renders correctly if disabled`, () => {
+    const component = shallow(
+      <Toggle id="mock-toggle-id" disabled />
+    ).getElement()
+    expect(component).toMatchSnapshot()
+  })
+
```
ptbrowne: les 3 cas au dessus sont gérés via les snap des exemples non ?

react/Toggle/index.jsx
ptbrowne: The usage as a controlled component is shown in the styleguide.

react/Toggle/test/toggle.spec.js
CPatchane: Indeed 👍

react/Toggle/index.jsx
CPatchane: Perfect so.

react/Toggle/test/toggle.spec.js
```diff
@@ -0,0 +1,33 @@
...
+import { configure, shallow } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-15'
+
+import Toggle from '../'
+
+configure({ adapter: new Adapter() })
+
+describe('Toggle component', () => {
+  beforeEach(() => {
+    jest.resetModules()
```
ptbrowne: why is it necessary ?

react/Toggle/test/toggle.spec.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+/* eslint-env jest */
+
+import React from 'react'
+import { configure, shallow } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-15'
+
+import Toggle from '../'
+
+configure({ adapter: new Adapter() })
```
ptbrowne: the adapter is already configured in the jest.setup file

ptbrowne:
Thanks @CPatchane 🌮 !

PR #466, opened 2018-05-02
Expose I18n extend() method

gregorylegarec:
Expose an `extend(dict)` method to be able to extend locales dictionnary. It could be useful for adding locales related to an app or a konnector.

Usage:

```js
import { extend } from 'cozy-ui/react/I18n'

// [...]

extend({
  'such': {
    'useful': 'very i18n'
   }
 })

// [...]

console.log(t('such.useful'))
// very i18n
```

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -74,4 +74,6 @@ export const translate = () => WrappedComponent => {
   return Wrapper
 }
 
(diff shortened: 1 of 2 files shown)


ptbrowne:
It would be a bit better with doc, but I am not sure where to put it since the I18n does not yet have a doc.

PR #468, opened 2018-05-02
 ✨ feat: transpilation

ptbrowne:
This is a work in progress. The goal is to transpile cozy-ui jsx and stylus files so that user of the library do not have to add something to their webpack config, thus improving usability of the library. This should also improve performance as the host webpack would have less to do.

## What's been done

JS files
* JSX files are transpiled into the `transpiled/` directory via `yarn transpile`

CSS
* A CSS stylesheet is extracted from the stylus files imported
by the JSX components during the `yarn transpile` phase. It should be imported by the host application.
* The CSS module classes are inserted into the components by the `css-module-transform plugin`

Polyfills
* Handled by babel through the [transform-runtime plugin](https://babeljs.io/docs/plugins/transform-runtime/)

Icons
* Bundled into a sprite just before transpilation. The application will have to add the sprite manually

## What's next

* Test on a real app, being done in Banks

## Breaking changes

1. Stylesheet must be inserted manually
2. `IconSprite` must be inserted manually
3. Icons lose their `icon-` prefix in their filename 

Check this for more information : https://github.com/cozy/cozy-ui/pull/468/files#diff-e7ea890a8dc2db7dbffd36fdc3d2226b

## How to test

On cozy-ui side

```
yarn link
yarn transpile --watch # will create the icon-sprite and babel watch
```

On app side

```
yarn link cozy-ui
```

You will need to add the icon sprite manually in the app (it was previously inserted by the svg-loader). [Check the example](https://github.com/cozy/cozy-ui/pull/468/files#diff-9777d1d36879cbac443b9da130b95b67R184).

## On polyfills

From the [documentation](https://babeljs.io/docs/plugins/transform-runtime/)

> This plugin is recommended in a library/tool.

> Babel uses very small helpers for common functions such as _extend. By default this will be added to every file that requires it. This duplication is sometimes unnecessary, especially when your application is spread out over multiple files.

> This is where the transform-runtime plugin comes in: all of the helpers will reference the module babel-runtime to avoid duplication across your compiled output. The runtime will be compiled into your build.


sebn:
@ptbrowne When running `yarn transpile`, at `dist/react/Icon/index.js:35` I see:

```js
var _icons = require('../../src/icons');
```

Is it normal?

sebn:
(I mean, in order to work, shouldn't this be `../../../src/icons` since I'm in `dist/`? Or should the `icons` file have been transpiled as well instead?)

sebn:
Also my `dist/stylesheet.css` is empty.

But except for those, I managed to import and display a cozy-ui Button! :tada:

sebn:
(changing the `yarn transpile` script to `babel react/ --out-dir dist/react && babel src/ --out-dir dist/src` did the trick)

sebn:
Maybe the preprocess function should not be async?
https://github.com/css-modules/css-modules-require-hook#preprocesscss-function

But the stylus API doesn't seem to include a `renderSync()` function...
http://stylus-lang.com/docs/js.html
(at least it didn't work when I tried to call it)

sebn:
https://github.com/stylus/stylus/issues/230

sebn:
My bad, `render()` actually returns the css when no callback given:
https://github.com/stylus/stylus/blob/dev/lib/renderer.js#L117

sebn:
Extracting CSS to separate files... e.g.:

```json
    extractCss: {
        "dir": "./dist/stylesheets/",
        "relativeRoot": "./react/",
        "filename": "[path]/[name].css"
      }
```
...then concatenating them by hand... e.g.:

```bash
find dist/stylesheets -name '*.css' | xargs cat > dist/stylesheet.css
```

...seems to work though (means the button is now styled as expected).

I'll try to play with more components to see if everything else looks fine!

sebn:
Resolution of `../../stylus/settings/palette.json` is failing too (e.g. as required by the `Avatar` component). But the stack I'm using may possibly be unable to load json files...

sebn:
:+1: Not getting an empty stylesheet anymore! 🎉 
But the transpiled selectors are prefixed/hashed.
Passing `generateScopedName: "[local]",` to `css-modules-transform` in `.babelrc` fixes the issue.
But I suppose we don't want to use this setting when we aren't transpiling...

drazik:
Babel 7 now supports JS config files (see https://babeljs.io/blog/2018/08/27/7.0.0#javascript-config-files). Maybe this can help here ?

ptbrowne:
I don't understand what do you mean by "when we are not transpiling" ? Do you mean that we should ship on npm a CSS with unhashed CSS classes and another version meant to be consumed by cozy UI React components with hashed CSS classes ?

sebn:
> I don't understand what do you mean by "when we are not transpiling" ? Do you mean that we should ship on npm a CSS with unhashed CSS classes and another version meant to be consumed by cozy UI React components with hashed CSS classes ?

My bad, I was not clear at all. There are actually 2 scenarios:
1. reusing (p)react components without depending on webpack (should work now)
2. reusing component styles without depending on react (not possible yet because of the randomly prefixed/suffixed css selectors)

I mistakenly tested scenario 2 while current focus is (and should definitely be) on scenario 1.
Please forget about my previous comment :)

I'll give scenario 1 a try again as soon as possible. Thanks!

ptbrowne:
> Please forget about my previous comment :)

No worries :D
I think to support scenario 2, we should ship on npm a styles directory with unhashed CSS classes. Would it support your needs ?

sebn:
> I think to support scenario 2, we should ship on npm a styles directory with unhashed CSS classes. Would it support your needs ?

I was thinking of a single css file, same as bootstrap, to make it very easy to include it without even requiring any build step.

Maybe we could have:
- `dist/stylesheet-react.css` (with hashed classes, to be used with the react components)
- `dist/stylesheet.css` (unhashed, to be used as is)

WDYT?

sebn:
Regarding the js side, I believe there is still one remaining *src* imports in the transpiled code, e.g.:

```
$ grep -rn 'src' node_modules/cozy-ui/index.js
node_modules/cozy-ui/index.js:1:import icons from './src/icons'
```

Not sure why it worked when we tried with Parcel...

sebn:
@ptbrowne I finally found a way to require all cozy-ui components. Everything seems to be working for now :tada: 

However I'm still unable to use the icons (typical console error: `Icon not found warn.`).

sebn:
I've set up a ClojureScript [demo](https://sebn.github.io/cljs-cozy-ui/) + [repo](https://github.com/sebn/cljs-cozy-ui) using the transpiled react components. I'll try to progressively add more of them.

ptbrowne:
This is ready to review.

sebn:
```
$ grep -r node-polyglot transpiled/react/I18n/translation.js
var _nodePolyglot = require('node-polyglot');
```

But `node-polyglot` is currently a dev dependency. Does it mean one should add it as a dependency in the consumer project to use it?

scripts/copy-files.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -eu
+
+# Palette needs to be accessible from stylus and from JS
+JS_PALETTE='react/palette.js'
```
sebn: I was expecting this to be `transpiled/react/palette.js` but it seems to work, not sure why (I'm probably missing something).

sebn:
I updated the ClojureScript repo & demo (although GitHub Pages builds are still disabled currently).
Icons seems to be working now! :tada: 
(although they're white, I probably missed something)

Also got a weird shadow-cljs build warning:

```
:shadow.build.npm/js-invalid-requires - {:resource-name "node_modules/cozy-ui/transpiled/react/I18n/format.js", :requires [{:line 23, :column 63} {:line 26, :column 22}]}
```

Which looks like a shadow-cljs or Google Closure bug (failing to handle some specific case of dynamic require). Probably not related to this PR.

scripts/copy-files.sh
ptbrowne: I want it to be both in source `react` and transpiled `react`.

ptbrowne:
For node-polyglot : this is a mistake, good catch.

sebn:
Huge work anyway! I can't wait to use it :heart_eyes_cat:

PR #476, opened 2018-05-03
feat: add calendar icon

drazik:
Add a calendar icon : image


ptbrowne:
I think you have to update the snapshots

`yarn jest -u`

ptbrowne:
Thanks drazik !

PR #479, opened 2018-05-03
<IntentHeader /> little improvements

gregorylegarec:
* Improve flex properties to avoid having a crushed header.
* Allow passing a className prop


ptbrowne:
When was there a crushed header ?

gregorylegarec:
@ptbrowne In a flex context like this one:

capture d ecran 2018-05-04 a 13 00 02
capture d ecran 2018-05-04 a 13 00 16

PR #484, opened 2018-05-07
Add <AsButton /> wrapper

gregorylegarec:


gregorylegarec:
[(2) ETU je peux accéder aux permissions de mon application [0]](https://trello.com/c/KjoheF3M/866-2-etu-je-peux-acc%C3%A9der-aux-permissions-de-mon-application-0)

gregorylegarec:
[(3) Amélioration Collect [1]](https://trello.com/c/BabTBSB8/941-3-am%C3%A9lioration-collect-1)

react/Button/index.jsx
```diff
@@ -59,6 +51,30 @@ export const Button = props => {
...
+export const AsButton = props => {
+  const { children } = props
+  const Wrapped = React.Children.only(children)
+  console.debug(Wrapped.props)
+  const buttonProps = props.isLink
+    ? {
+        ...ButtonLink.defaultProps,
+        ...actualProps(ButtonLink.propTypes, props)
+      }
+    : { ...Button.defaultProps, ...actualProps(Button.propTypes, props) }
```
ptbrowne: const base = props.isLink ? ButtonLink : Button

ptbrowne: const base = props.isLink ? ButtonLink : Button

ptbrowne: const base = props.isLink ? ButtonLink : Button

GoOz:
I'm just gonna ask a naive question but wouldn't it be better to make a `NavLink` props to display a button instead of a text link than wrapping as Button ? 
Unless you think there would be other usecases outside of the Nav component that would also need this wrapper. In that case just ignore my question.

gregorylegarec:
@GoOz That actually was my first thought, but it implies to add a dependency on `NavLink` from `react-router` into Cozy-UI. And I did not want to add such dependency. With this actual wrapping solution we just abstract the button props and we are able to wrap any component that render as a `<a />` tag. Even if for now, the only known case is for `<NavLink />`.

GoOz:
OK seems legit to me then :)

ptbrowne:
What do you think about https://github.com/cozy/cozy-ui/pull/488 ?

IMHO, it neatly refactors Button and ButtonLink together and fills up the `AsButton` requirements.

PR #486, opened 2018-05-09
Fix <ButtonLink /> className prop format

gregorylegarec:
The call to `btnClass()` was still made with several arguments instead of an object. The tests did not spot this because of a bug in it, see #485.


CPatchane:
Thanks @gregorylegarec ⚡️

PR #488, opened 2018-05-09
feat: Button can render with a different Tag [breaking]

ptbrowne:
It is useful for a Button to render with a different tag.

Another solution for #484 that does not extend the API of Cozy UI.

For example if you want a Button to render with a NavLink : 

```
<Button tag={NavLink} to='/toto' />
```

It is a breaking change since `<AsButton />` is removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "prebuild:css": "mkdir dist/",
     "prebuild:css:app": "mkdir -p build/styleguide",
     "prebuild:doc:kss": "run-s clean:doc:kss build:css:app",
+    "prepush": "yarn lint",
     "test": "npm-run-all --serial lint test:jest",
     "test:jest": "jest --verbose --coverage",
     "watch:doc:react": "styleguidist server --config docs/styleguide.config.js"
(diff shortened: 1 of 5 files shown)


gregorylegarec:
Great, I love it. 👍
And it simpler that the pseudo-HOC `<AsButton />`.
I am not familiar with passing components as props, but now I'll use this pattern.


Does the use of `{...restProps}` fit well with PropTypes validation ?

ptbrowne:
> Does the use of {...restProps} fit well with PropTypes validation ?

Props passed with `...restProps` are validated with PropTypes. Was it your question ?

gregorylegarec:
@ptbrowne In your example you're passing a `to` property to pass it to the `<NavLink />` element. As I am not yet very familiar with PropTypes mechanism, is it ok from a React point of view, if this property `to` is not declared as a `<Button />`'s PropType ?

ptbrowne:
Yes it's OK. React does not emit any warning if a property is passed that is not declared in the PropTypes.

Issue #491, opened 2018-05-12
top right close button in modal overlaps title


Benibur:
it happens for any width (here size='xxlarge')
![image](https://user-images.githubusercontent.com/1266420/39957131-c3876ff4-55ed-11e8-829f-2e2b56a8e386.png)
![image](https://user-images.githubusercontent.com/1266420/39957132-c7cd5e0c-55ed-11e8-8243-ce6022d9a89f.png)

ptbrowne:
I fixed : it will be available in the next version of Cozy-UI.

PR #492, opened 2018-05-13
fix: add padding for cross in modals

ptbrowne:
fix https://github.com/cozy/cozy-ui/issues/491

The cross is positioned absolutely so it has no effect on the layout of the header. For the header not to overflow on the cross, I added right padding on the header when the modal is showing the cross (= when it is closable). 

View on https://ptbrowne.github.io/cozy-ui/react/, the complex modal has a long title.

image

### DIFF
--- a/react/Modal/Readme.md
+++ b/react/Modal/Readme.md
@@ -146,6 +146,11 @@ For more complex modals, you can use individual components.
 
 ```
 const { ModalDescription, ModalHeader, ModalFooter } = Modal;
+const headerStyle = {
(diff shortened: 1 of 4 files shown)


CPatchane:
I don't understand, is it for the cross or for the header itself? The code seems to say that is related to the header.

ptbrowne:
Sorry, the description of the issue was not very clear. I edited : 

> The cross is positioned absolutely so it has no effect on the layout of the header. For the header not to overflow on the cross, I added right padding on the header when the modal is showing the cross (= when it is closable).

PR #495, opened 2018-05-14
tests: use one file per automatic snapshot

ptbrowne:
1 inconvenient of the automatic snapshots was the big file that was difficult to review
when there was a lot of changes. Here the big snapshot file is splitted into several
snapshot files (one per example) so that it is easy to review. `jest-specific-snapshot`
is used to achieve this. The title for each example is taken from the nearest previous
markdown title.

jest-specific-snapshot blogpost : https://medium.com/sears-israel/adding-multi-snapshot-testing-to-jest-b61f23cf17ca


react/testFromStyleguidist.js
```diff
@@ -55,6 +58,33 @@ jest.mock('react-styleguidist/lib/rsg-components/PlaygroundError', () => {
...
+  }
+  const title = find(
+    markdownChunk.content.split('\n'),
+    x => x.indexOf('#') > -1
+  )
+  return title ? title.replace(/^#+\s/, '') : ''
+}
+
+const codeAndTitle = chunks => {
+  return withLastMatching(chunks, chunk => chunk.type == 'markdown')
```
gregorylegarec: `==` instead of `===`

y-lohse:
@ptbrowne Were there any major reasons why this shouldn't be merged, or does it just need a rebase? Were you planning on doing any of that?

ptbrowne:
I had some problems  but I don't remember what they were. I prefer to close. It seems it is not so important anyway.

PR #496, opened 2018-05-15
Add 'button' value for 'type' in Button

gregorylegarec:
Edit: Let's just add the `button` value for prop `type` in `<Button />`.

~~Default type should be `button`, to avoid having several button with type `submit` in the same form, for example. The type `submit` is supposed to send form data.~~

From https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button:
> The type of the button. Possible values are:
> * `submit`: The button submits the form data to the server. This is the default if the attribute is not specified, or if the attribute is dynamically changed to an empty or invalid value.
> * `reset`: The button resets all the controls to their initial values.
> * `button`: The button has no default behavior. It can have client-side scripts associated with the element's events, which are triggered when the events occur.

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -137,7 +137,7 @@ Button.propTypes = {
   /** Disables the button */
   disabled: PropTypes.bool,
   /** Type of the underlying `<button />` */
-  type: PropTypes.oneOf(['reset', 'submit']),
+  type: PropTypes.oneOf(['button', 'reset', 'submit']),
   /** Use the `subtle` alternative look for the Button  */
   subtle: PropTypes.bool
 }


y-lohse:
Thanks for the PR! I'm not a code owner here, but I don't think we should make this change 🤷‍♂️   :

1. This is a breaking change, we have `<button>` elements inside forms that act as submit button right now, they would need to be changed. I have no idea where they all are, but pretty sure there are some in Drive and Contacts.
2. I don't think we should have a default `type` that is different from the spec's default.

I think the correct approach is to specify the `type` for buttons inside forms when they are not supposed to act as submit buttons, just as we would when doing regular HTML.

ptbrowne:
I agree with @y-lohse , since the HTML default is type="submit", our Button should not change this behavior.

https://fr.wikipedia.org/wiki/Principe_de_moindre_surprise

GoOz:
I couldn't agree more.

CPatchane:
Also completely agree about keeping the consistency with the HTML spec with the default to `submit`.

gregorylegarec:
Ok, I think you're totally right guys. I suggest to just add the `button` value in the PropTypes.

PR #497, opened 2018-05-15
fix: close button in selection bar

y-lohse:
On mobile, the button to close the selection bar was hidden.

This is because there was a rule hiding `span` inside `button`, which was meant to target the text labels in the selection bar, but accidentally also targeted the `span` inside our `Button` component.

I hope I'm somewhat clear? Anyway, it's fixed.


ptbrowne:
There are lots of coverage related files ? Is it normal ?

y-lohse:
Oh god no.

I ran the tests locally, I think we're missing a `gitignore` entry. i'll clean this up and do a separate PR.

PR #498, opened 2018-05-15
Set intent modal overflow to 'hidden' to preserve border radius

gregorylegarec:


ptbrowne:
What is your problem exactly ? Do you have a reproducing case ? There is a overflowHidden property that can be set on the modal. Would it solve your problem ?

gregorylegarec:
@ptbrowne Here's an example of an intent to Collect into a intent modal. The border should be rounded.

capture d ecran 2018-05-15 a 19 57 08

An alternative to setting `overflow: hidden` directly on modal is maybe to set it on `.intentModal` class instead. What do you think ? Or event force `overflowHidden` prop to `true` on intent modals ?

gregorylegarec:
Plot twist : it isn't.

capture d ecran 2018-05-15 a 22 13 12

gregorylegarec:
Second plot twist : the scrollbars seem to be a regression in cozy-ui. An investigation is ongoing.

gregorylegarec:
Investigation concluded to a deprecated local version of Collect, which was displaying the scrollbars.

I updated my commit to only add `overflowHidden` on every `<IntentModal />`.

PR #501, opened 2018-05-15
feat: add SelectBoxWithFixedOptions

ptbrowne:
This components supports options with the `fixed: true` attribute to put them in a non scrollable area below other options.

Visible on https://ptbrowne.github.io/cozy-ui/react/#!/SelectBox/7 (cropped by the end of the body but it is not a problem when body has height 100%).

##### Implementation

`react-select` does not provide an easy way to do this as it is not possible to extend a component that is inside the white area of the menu (MenuList) but outside the scrollable area. The fixed area also needed to have real `Option`s inside with the selected/OnSelect behavior.

To achieve this, I needed to rearrange directly the children. This seems a bit brittle but it does the job and I tried to keep the ugly bits tucked inside a single function.

##### Screenshots

In styleguide : 

In Banks :


ptbrowne:
> And then we could render the FixedOptions, the MenuList, and whatever else needs to be rendered there.

The thing is that the `options` received are not yet rendered and we would have to duplicate the render method here https://github.com/JedWatson/react-select/blob/v2/src/Select.js#L1310

react/testFromStyleguidist.js
```diff
@@ -134,7 +135,8 @@ const testFromStyleguidist = (name, markdown, require) => {
       Tabs,
       Text,
       Textarea,
-      Toggle
+      Toggle,
+      true
```
CPatchane: I may miss something out, but why is it just `true` here?

react/SelectBox/SelectBoxWithFixedOptions.js
```diff
@@ -0,0 +1,86 @@
...
+  if (!menuList) {
+    return { fixed: [], newChildren }
+  }
+  const options = menuList.children
+  const { true: fixed, false: normal } = groupBy(
+    options,
+    vnode => vnode.attributes.data.fixed === true
+  )
+  newChildren.children[0].children = normal
+  return { fixed, newChildren }
```
CPatchane: Just my personal point but I think this part is not readable. We have to understand that you will get an object with two properties `true` and `false` (not usual as property names) to become then the `fixed` and `normal` objects.

I would (personally) prefer a classical `map() usage` here:
```javascript
const fixed = []
const normal = []
options.map(vnode => {
  if (vnode.attributes.data.fixed === true) { fixed.push(vnode) }
  else { normal.push(vnode) }
})
newChildren.children[0].children = normal
return { fixed, newChildren }
```

But this point is not blocking since it does the job.

docs/styleguide.config.js
```diff
@@ -22,7 +22,7 @@ module.exports = {
         '../react/Input/index.jsx',
         '../react/Label/index.jsx',
         '../react/Radio/index.jsx',
-        '../react/SelectBox/index.jsx',
+        '../react/SelectBox/SelectBox.jsx',
```
CPatchane: Do you test the component with the fixed option so?

ptbrowne: Since it is in the styleguide, it gets snapshotted.

react/SelectBox/SelectBoxWithFixedOptions.js
gregorylegarec: More a `forEach` than a `map`, even a `reduce`

react/testFromStyleguidist.js
ptbrowne: This is the value of the isTesting option. Here the function that will execute the code inside the codeblocks is defined. For example if you have a code block in the Readme.md like that : 

```
<Button>My Button</Button>
```

, it gets wrapped inside this function like that : 

```
function (require, ..., Button, ... Toggle, isTesting) {
  <Button>MyButton</Button>
}
```

and this function is bound so that each argument is correctly defined. Here the `true` is the value bound to the argument `isTesting` so that you can use `isTesting` inside your codeblocks, in the jest context. In the styleguide context, this is achieved via the context option of styleguidist.

The `isTesting` variable is useful to have the selectBox opened when snapshotting, so that `<options>` are snapshotted.

react/SelectBox/SelectBoxWithFixedOptions.js
ptbrowne: I changed to 

```
```patch
-  const { true: fixed, false: normal } = groupBy(
+  const { fixed, normal } = groupBy(
     options,
-    vnode => vnode.attributes.data.fixed === true
+    vnode => (vnode.attributes.data.fixed === true ? 'fixed' : 'normal')
   )
```

I do not find `reduce` appropriate. Sure, eventually you are reducing the array to an object, but it seems like an implementation detail. The intention here is to group the elements via a predicate and this is what groupBy does and imply.

PR #516, opened 2018-05-27
Default `htmlFor` to undefined for `<Label/>`

CPatchane:
It will avoid these warnings in console `Empty string passed to getElementById().` when we just want to use `<Label/>` without the `for` attribute.

### DIFF
--- a/react/Label/index.jsx
+++ b/react/Label/index.jsx
@@ -29,7 +29,6 @@ Label.propTypes = {
 }
 
 Label.defaultProps = {
-  htmlFor: '',
   className: '',
   error: false
 }
--- a/react/__snapshots__/examples.spec.jsx.snap
+++ b/react/__snapshots__/examples.spec.jsx.snap
@@ -451,7 +451,7 @@ exports[`Field should render examples: Field 1`] = `
   <div>
     <form>
       <div class=\\"o-field\\">
-        <label for=\\"\\" class=\\"c-label\\">I'm a label</label>
+        <label class=\\"c-label\\">I'm a label</label>
         <textarea placeholder=\\"I'm a textarea\\" class=\\"c-textarea\\" id=\\"idField\\"></textarea>
       </div>
     </form>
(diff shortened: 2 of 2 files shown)


ptbrowne:
And it will not add unmeaningful attributes to snapshots :)

GoOz:
Sorry to insist but… a label is never to be used on its own, it's always with an input-(like) element and those two have to be linked some way.

3 things: 
* You need an element that look like a `label` but isn't semantically one, then don't use a `label` element, use a `span` or a `div` and apply to it the label's style.
* You need a `label` with an `input` but don't want to add an `id` to the input ? Do it anyway. It **has** to be done.
* Your `label` is wrapping the `input`, that's actually the only case where the `for` attribute is not needed

Which is your case?

CPatchane:
I didn't have a specific case to not use the `htmlFor` and I don't have any preference about default value `false` or `undefined`. I just used the label with an input without linking them and got some weird warnings `Empty string passed to getElementById()`. I had to figure out what these was from and wanted to avoid these unmeaningful warnings with potential searches about them in future usages.
@GoOz Ok about your point, but we have to put that in the documentation so. I think labels can usually be used with inputs and forms but not necessarily with the `linked behaviour`. And if it's a good practice to always use the `for` attribute, we maybe have to say it in our guidelines as well.
Another solution can be to make the label an option of an input cozy-ui component. In this case maybe we will always have the elements linked.

GoOz:
It's in my opinion way more than a good practice.
Although I agree the error message isn't explicit enough.

ptbrowne:
> Your label is wrapping the input, that's actually the only case where the for attribute is not needed

`<Label>` should be able to accomodate this usecase. Principle of least surprise again :)

> I don't have any preference about default value false or undefined.

Here it is not a matter of preference : `false` has no meaning in this context, we merely want `htmlFor` not to be defined.

GoOz:
> <Label> should be able to accomodate this usecase. Principle of least surprise again :)

I agree I just don't know how to properly do that. 
And I'm afraid that if we don't make this attribute required then the good practice will never be used.

ptbrowne:
> if we don't make this attribute required 

This attribute is not required.

> And I'm afraid that if we don't make this attribute required then the good practice will never be used.

I understand your point but the `defaultProps` mechanism here cannot be used. This good practice is only relevant when the `Label` does not wrap the `Input` and the `defaultProps` mechanism cannot check that.

GoOz:
I know. I'm just sad about it. But I can deal with it.

CPatchane:
It's done. Thanks again @ptbrowne @GoOz and @drazik 👍

drazik:
Sorry to be so meticulous, but setting `undefined` as a default value is redundant : it is basically `undefined` if you don't pass a value.

PR #518, opened 2018-05-30
Correctly import css code to build

CPatchane:
Previously it was an `import` statement for `normalize.css` from `node_modules`, broking usage in apps. Now the required content of `normalize.css` is correctly imported.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,8 @@
   },
   "homepage": "https://github.com/cozy/cozy-ui",
   "scripts": {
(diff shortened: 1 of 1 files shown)


ptbrowne:
> broking usage in apps. 

For my culture, what was broken ?

CPatchane:
> For my culture, what was broken ?

@ptbrowne If we want to use only css without stylus, the normalize import from the build uses the wrong path from the application styles file. Now it's all included as css code.

PR #519, opened 2018-05-31
chore: use commitlint to ensure all commit messages are valid

drazik:
Using [commitlint](https://github.com/marionebl/commitlint) and [husky](https://github.com/typicode/husky), we ensure that the commit messages are well formatted. The rules are the one from [Conventional Commits](https://conventionalcommits.org/), that we already follow.

Failing : 

image

Success : 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "clean:doc:kss": "rm -rf build/styleguide",
     "deploy:doc": "git-directory-deploy --directory build/ --branch gh-pages",
     "lint": "npm-run-all 'lint:*'",
+    "lint:commit": "commitlint --from=$(git merge-base HEAD master) --to=HEAD",
     "lint:js": "eslint 'react/**/*.jsx' 'react/**/*.js'",
     "lint:stylus": "stylint stylus --config .stylintrc",
     "prebuild:css": "mkdir dist/",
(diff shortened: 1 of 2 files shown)


ptbrowne:
Can this be done before push instead of at each commit ?

gregorylegarec:
Does it handle correctly the cute emojis we are using as prefix ? 🌻

drazik:
@ptbrowne I don't know if it's possible, since `commitlint` just take a commit message as parameter and returns if it's well formed or not. I don't know if  we can get all commit messages in a `prepush` hook. So I asked here : https://github.com/marionebl/commitlint/issues/361

drazik:
After discussing it with @ptbrowne, the idea would be to get the commits between `branch` and `origin/branch` in a `prepush` hook and pass them all to `commitlint` in a loop. I will look at it and propose something :)

ptbrowne:
Thanks Cyrille, that's great !

drazik:
I got an answer on the issue on commitlint : https://github.com/marionebl/commitlint/issues/361#issuecomment-394083870

I still have some investigation to do, because : 

* What happen if there's no remote branch for the local branch we are going to push ? We can look for a common commit in the tree, and lint the commit messages between this one and `HEAD`
* What happens if we changed the history and want to push force ? We shouldn't check the remote branch anymore since we are going to overwrite it completely

So instead of linting from `origin/branch`, maybe we should always lint from the common ancestor, to `HEAD` ?

ptbrowne:
> But we need to find a way to find the commit at which our branch derived.

`git merge-base HEAD master`

PR #520, opened 2018-05-31
chore: setup semantic-release

drazik:
This setups semantic-release.

What we get on each PR that get merged : 

* A new release of cozy-ui published on npm
* A new release of cozy-ui created on Github (with changelog in it)
* `CHANGELOG.md` is updated (with this format : https://github.com/drazik/my-awesome-module/blob/master/CHANGELOG.md)

What we need to do : 

* Respect the [conventional commit messages format](https://conventionalcommits.org/) (we are working on #519 to lint the commit messages and throw an error if a message is not valid). It means that emojis at the beginning of commit messages are no longer valid :disappointed: 
* Discuss if we really want to release a new version on each merged PR or not

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,8 @@ after_success:
 - test $TRAVIS_BRANCH = "master" && test $TRAVIS_REPO_SLUG = "cozy/cozy-ui" && yarn
   build:doc && yarn deploy:doc -- --username cozycloud --email contact@cozycloud.cc
   --repo=https://$GITHUB_TOKEN@github.com/cozy/cozy-ui.git
+- test $TRAVIS_BRANCH = "master" && test $TRAVIS_REPO_SLUG = "cozy/cozy-ui" && yarn travis-deploy-once "yarn semantic-release"
 
-deploy:
-  provider: npm
-  email: npm@cozycloud.cc
-  skip_cleanup: true
-  api_key:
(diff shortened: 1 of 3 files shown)


ptbrowne:
Great :)

ptbrowne:
Do you plan to still use `lerna-changelog` ? I think a combination of lerna-changelog and handlebars would work right ?

drazik:
Actually the original changelog plugin (https://github.com/semantic-release/changelog) says it can handle this. So I will just do some tests to see that.

drazik:
The plugin generates a CHANGELOG.md file like this one : https://github.com/drazik/my-awesome-module/blob/master/CHANGELOG.md

There's no links at the end to compare between each versions. If we want to stick with the same format that we have today, I think we will have to write a plugin.

ptbrowne:
I think it's great. Personally I seldom use the links to compare. You can compare two versions manually through the [compare page](https://github.com/cozy/cozy-ui/compare). 🎉

Thanks @drazik !

ptbrowne:
Last thing to think of is whether we want to work on a dev branch and only push a release when we merge it to master. What do you think ?

drazik:
I personnaly prefer to release for every PR that get merged. If we want to release a big milestone with a lot of new features / breaking changes, we can still create a branch that synchronizes these PRs, then get merged when everything is ready to be shipped.

But I'm open to discuss pros and cons of course, if anyone have an opinion about this.

drazik:
@gregorylegarec because I know you love emojis, I did a test with an emoji at the beginning of a commit message, and it doesn't work with semantic-release either :

gregorylegarec:
💔

ptbrowne:
Thanks @drazik ! Let's merge ?

drazik:
I just rebased after merging #519, and removed `changelog` script since it's not needed anymore. It should be OK. We will have to keep an eye on it to see if it's really working.

PR #521, opened 2018-06-04
WIP feat(SelectBox): show an Overlay on mobile

drazik:
On banks we use the `SelectBox` for our date selector component. We found it's tricky to close it without clicking on another element of the page that does another action (a transaction row that opens a modal, for example).

That's why we propose to add a transparent overlay when a `SelectBox` is focused. This way, we can click everywhere outside the select to close it, without misclicking on another element and do something unexpected.

https://drazik.github.io/cozy-ui/react/#selectbox

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -1,15 +1,18 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import classNames from 'classnames'
 import ReactSelect from 'react-select'
 import styles from './styles.styl'
 import Icon from '../Icon'
 import palette from '../../stylus/settings/palette.json'
+import Overlay from '../Overlay'
+import withBreakpoints from '../helpers/withBreakpoints'
 
 const customStyles = {
-  container: base => ({
+  container: (base, state) => ({
     ...base,
-    maxWidth: '30rem'
+    maxWidth: '30rem',
+    zIndex: state.isFocused ? styles.focusedZIndex : 10
   }),
   control: (base, state) => ({
     ...base,
@@ -42,9 +45,9 @@ const customStyles = {
     ...base,
     color: 'black'
   }),
-  menu: base => ({
+  menu: (base, state) => ({
     ...base,
-    zIndex: 10
+    zIndex: state.isFocused ? styles.focusedZIndex : 10
   })
 }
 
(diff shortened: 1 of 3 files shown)


y-lohse:
Thanks, sounds good!
I was wondering if it should be the default behavior or a config option etc, but it looks like it's the default behaviors for regular select boxes too : https://codesandbox.io/s/l96r22q5zm

So 👍 for me

react/SelectBox/SelectBox.jsx
```diff
@@ -1,15 +1,18 @@
...
 import palette from '../../stylus/settings/palette.json'
+import Overlay from '../Overlay'
+import withBreakpoints from '../helpers/withBreakpoints'
 
 const customStyles = {
   container: base => ({
     ...base,
-    maxWidth: '30rem'
+    maxWidth: '30rem',
+    zIndex: 51
```
ptbrowne: mmh 51, must be coming from `$overlay-index + 1`, could this be explicited above ?

I think you can even get the value of overlay-index from a CSS variable that could be instantiated from the stylus variable.

https://github.com/css-modules/css-modules/issues/106

react/SelectBox/SelectBox.jsx
```diff
@@ -100,13 +103,63 @@ const CheckboxOption = ({ ...props }) => <Option {...props} withCheckbox />
...
+    const {
+      components,
+      styles,
+      breakpoints: { isMobile },
+      ...props
+    } = this.props
+    const showOverlay = this.state.isOverlayVisible && isMobile
+
+    return (
+      <div>
```
ptbrowne: The wrapping in a `div` can change the display of the resulting component from `inline-block` (if it is its default value at the moment) to `block`, is it OK ?

drazik: I had to add a `className` prop to `<SelectBox />` and apply it to this div. It may be a breaking change if the app is passing style to the `<SelectBox/>` root element via react-select's styles API. It now have to pass it via `className`.

drazik:
I put this in WIP status because there's a problem when integrating this in banks for now

y-lohse:
Since this hasn't moved for several month's, I'm closing the PR. Feel free to re-open it though.

However I think that it should be possible to use an `Overlay` component combined with the `SelectBox` to achieve what you want?

PR #522, opened 2018-06-05
fix: enables only overflow-y on small screen

GoOz:
To avoid horizontal scrollbar on Windows' browsers, especially inside intent modals.

The horizontal scrollbar appears because there's a vertical one. In WIndows OS, the scrollbar adds itself to the element it scrolls so it increases the element's width. Why ? Because fuck you, that's why.

### DIFF
--- a/stylus/elements/defaults.styl
+++ b/stylus/elements/defaults.styl
@@ -43,7 +43,7 @@ body
     overflow-y  auto
 
     +medium-screen()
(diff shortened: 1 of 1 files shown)


GoOz:
[Interapp modal scrollbars](https://trello.com/c/cV0Pyc72/176-interapp-modal-scrollbars)

PR #524, opened 2018-06-06
feat: add global utility classes

GoOz:
Made utility classes global so you don't have to create classes and extend them anymore. You can just use the class without CSS-Modules now.

Before:

```jsx
<Element className={styles['u-ellipsis']}>
      {children}
</Element>
```

After:

```jsx
<Element className='u-ellipsis'>
      {children}
</Element>
```

### DIFF
--- a/.stylintrc
+++ b/.stylintrc
@@ -56,10 +56,7 @@
     "error": true
   },
   "mixins": ["hide","reset"],
-  "namingConvention": {
-    "expect": "lowercase-dash",
-    "error": true
-  },
+  "namingConvention": false,
   "namingConventionStrict": false,
   "none": {
     "expect": "never",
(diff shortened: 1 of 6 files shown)


ptbrowne:
Is there a reason why we do not use loops to generate the classes ?

ptbrowne:
If extending does not work from a lookup, I think it would be better if the global classes were declared in each file. For example, $color- placeholders are declared in `text.styl` and at the same place, utility classes could be declared. This way we do not have to change two files for the same functionality and we do not have to repeat ourselves since we will be able to do : 

```patch
for color in keys(colors)
    $color-{color}
        color: lookup(color) !important // @stylint ignore

+   .u-color-{color}
+        color: lookup(color) !important // @stylint ignore
```

GoOz:
Because I didn't think of this. I usually don't. Because it's not CSS I can't add this in my way of thinking… :/

GoOz:
New PR https://github.com/cozy/cozy-ui/pull/525

PR #525, opened 2018-06-06
feat: add global utility classes

GoOz:
Made utility classes global so you don't have to create classes and extend them anymore. You can just use the class without CSS-Modules now.

Before:

```jsx
<Element className={styles['u-ellipsis']}>
      {children}
</Element>
```

After:

```jsx
<Element className='u-ellipsis'>
      {children}
</Element>
```
📌  [previous merge](https://github.com/cozy/cozy-ui/pull/524)

### DIFF
--- a/.stylintrc
+++ b/.stylintrc
@@ -56,10 +56,7 @@
     "error": true
   },
   "mixins": ["hide","reset"],
-  "namingConvention": {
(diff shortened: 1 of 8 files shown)


GoOz:
[Make utilities global](https://trello.com/c/ohAV8lhM/138-make-utilities-global)

PR #526, opened 2018-06-06
feat: added download to phone icon

y-lohse:
It's an extra icon.


y-lohse:
[(1) Erreur lors d'un appui long pour sélectionner sur Android](https://trello.com/c/tywfjzfH/958-1-erreur-lors-dun-appui-long-pour-s%C3%A9lectionner-sur-android)

GoOz:
It's the same that used to be in Drive repo, right?

y-lohse:
> It's the same that used to be in Drive repo, right?

…possibly? I think we may have a variation of it somewhere.

GoOz:
I know there's those two used for offline status on mobile.
https://github.com/cozy/cozy-drive/blob/master/src/drive/assets/icons/icon-phone-download-white.svg
https://github.com/cozy/cozy-drive/blob/master/src/drive/assets/icons/icon-phone-download.svg

If it's the same, it could be nice to use this one instead and remove those two.

y-lohse:
Good catch, thanks. I'll make sure to look into that.

Issue #478, opened 2018-05-03
Utility classes for cursors


ptbrowne:
It would be nice to have

```
.u-c-help {
  cursor: help
}

.u-c-pointer {
  cursor: pointer
}
```

What do you think ?

GoOz:
why not. :)

ptbrowne:
I've only ever used `help` and `pointer`, so I think we can stick with only those two and wait if the need arise of others :)

GoOz:
Cool, I was more inclined to do that too :)

PR #529, opened 2018-06-07
fix: global styles are compiled only when used with CSS-modules

GoOz:
`:global()` shouldn't appear in CSS files as this is not a valid CSS syntax and therefore browsers would ignore the entire declaration making the utility classes for the standalone build ignored.

Instead I used a mixin `global()` to output a normal utility class for the standalone build and a version with `:global()` when used with CSS-modules.

`global()` takes two parameters:
- `selector` which should be a string for the CSS class you want
- `placeholder` which is the placeholder you want the class to `@extend`

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -674,24 +674,13 @@ Display an avatar of a user or contact with initials or an image
     Markup:
     <p class="{{modifier_class}}">I'm visible! 🎉</p>
 
-    .u-hidden - Element is visually hidden but still readable by screen-readers
+    .u-visuallyhidden - Element is visually hidden but still readable by screen-readers
     .u-hide - Element is totally hidden (even for screen readers)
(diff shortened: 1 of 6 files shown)


GoOz:
[#478: Utility classes for cursors](https://trello.com/c/KItE9y9c/160-478-utility-classes-for-cursors)

stylus/cozy-ui/build.styl
```diff
@@ -674,24 +674,13 @@ Display an avatar of a user or contact with initials or an image
     Markup:
     <p class="{{modifier_class}}">I'm visible! 🎉</p>
 
-    .u-hidden - Element is visually hidden but still readable by screen-readers
```
CPatchane: Is it a breaking change?

GoOz: No, I forgot to update the doc a long time ago and just noticed today… 😞

GoOz: It was like 8 month ago…

PR #531, opened 2018-06-08
Update react select

ptbrowne:
`react-select` new version enables `className` customization for each component.

I exported the components from `react-select` because if the host application has another version of `react-select` and start passing components from a newer version of `react-select` to "our" `react-select`, it causes very hard to debug problems.

### DIFF
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -69,3 +69,29 @@ options[options.length - 1].fixed = true;
   menuIsOpen={isTesting ? true : undefined}
   options={options} />
 ```
+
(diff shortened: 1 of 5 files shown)


react/SelectBox/Readme.md
```diff
@@ -69,3 +69,29 @@ options[options.length - 1].fixed = true;
   menuIsOpen={isTesting ? true : undefined}
   options={options} />
 ```
+
+### Customize ReactSelect components
+
+Do not directly another version of react-select, cozy-ui exports the components
```
y-lohse: Missing a word here, I guess "install".

Issue #514, opened 2018-05-25
Remove duplication from `build.styl` and the stylus files in each React component


ptbrowne:
In each React component, there is a stylus file that is "instantiating" the stylus placeholders. The `build.styl` also does that. Maybe we could use the react component stylus files to generate the build.css (compile each stylus file and concatenate the result). This would remove a lot of code that is in `build.styl`. Some code would still need to be there to instantiate global classes (utility classes for example).

GoOz:
Good idea 👍

PR #532, opened 2018-06-11
Chore/clean doc

### LINKED ISSUE #463: Add real classes using placeholders in the vanilla build
Some classes are lacking from the build

- [ ] Chip

GoOz:
Clean up documentation.
- [x] Use components styles to extend standalone classes to avoid duplication
- [x] Add missing documented classes.
- [x] global clean up & optimisation

Fixes #463 #514

### DIFF
--- a/react/Chip/styles.styl
+++ b/react/Chip/styles.styl
@@ -1,16 +1,16 @@
 @require '../../stylus/components/chip'
 
-.Chip
+.c-chip
     @extend $chip
 
-.RoundChip
+.c-chip--round
     @extend $round-chip
 
-.ChipSeparator
+.c-chip-separator
     @extend $chip-separator
 
-.ChipButton
+.c-chip-button
     @extend $chip-button
 
-.ChipButton--disabled
+.c-chip-button--disabled
     @extend $chip-button--disabled
--- a/react/Nav/styles.styl
+++ b/react/Nav/styles.styl
@@ -12,8 +12,10 @@
 .c-nav-link
     @extend $nav-link
 
+.c-nav-icon // Deprecated
 .c-nav-item__icon
     @extend $nav-item-icon
 
+.c-nav-text // Deprecated
 .c-nav-item__text
     @extend $nav-item-text
--- a/react/Nav/styles.styl
+++ b/react/Nav/styles.styl
@@ -12,10 +12,8 @@
 .c-nav-link
     @extend $nav-link
 
-.c-nav-icon // Deprecated
-.c-nav-item__icon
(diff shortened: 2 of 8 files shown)


GoOz:
[#514: Remove duplication from `build.styl` and the stylus files in each React component](https://trello.com/c/8Dg5vclH/173-514-remove-duplication-from-buildstyl-and-the-stylus-files-in-each-react-component)

GoOz:
[#463: Add real classes using placeholders in the vanilla build](https://trello.com/c/zEpTLEFb/154-463-add-real-classes-using-placeholders-in-the-vanilla-build)

react/Chip/styles.styl
```diff
@@ -1,16 +1,21 @@
...
 .ChipSeparator
+.c-chip-separator
     @extend $chip-separator
 
 .ChipButton
+.c-chip-button
     @extend $chip-button
 
 .ChipButton--disabled
+.c-chip-button--disabled
```
ptbrowne: We should remove the `ChipButton` classes (and thus modify the ChipButton jsx), this way there is no duplication :)

react/Nav/styles.styl
```diff
@@ -12,8 +12,10 @@
 .c-nav-link
     @extend $nav-link
 
+.c-nav-icon // Deprecated
 .c-nav-item__icon
     @extend $nav-item-icon
 
+.c-nav-text // Deprecated
 .c-nav-item__text
```
ptbrowne: Same here

react/Chip/styles.styl
GoOz: Yes, It's just that this component is done differently and I'm not sure I understand it all. I tried something and failed. So I figured it would have to be done later. But maybe I can give it another try.

react/Nav/styles.styl
GoOz: Well, the react component doesn't use the same classes and every app doesn't use the React component, I didn't want to make it a breaking change out of it and block UI updates just for a classname change.

react/Chip/styles.styl
GoOz: Nevermind… I found why I could do it. It's done :)

react/Nav/styles.styl
ptbrowne: > the react component doesn't use the same classes

Yes, but we can change the component so it uses the same classes as the vanilla css version. It it not a breaking change as the classnames are private to the component and not part of its API.

GoOz: Oh yes you mean the other way around… true

PR #536, opened 2018-06-12
feat: add rem() mixin

GoOz:
Adding `rem($value, $basefont)` mixin to ease rem calculation.
`$value`: The value in pixel you want to translate in rem. For multiple values you can use `rem(14 12)` for shorthand properties such as `margin`. ⚠ Multiples values should be separated by spaces.
 `$basefont`: If you need to overwrite the default `$basefont` value of `16`

The interesting part to review is [this one](https://github.com/cozy/cozy-ui/compare/master...GoOz:feat/remMixin?expand=1#diff-efc2168eed258d98c3b2ee3aa64e5c04), every other files is just using the mixin.

### DIFF
--- a/react/Hero/styles.styl
+++ b/react/Hero/styles.styl
@@ -1,4 +1,5 @@
 @require 'settings/breakpoints.styl'
+@require 'tools/mixins'
 
 .Hero
     margin-left auto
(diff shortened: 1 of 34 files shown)


GoOz:
[rem() mixin](https://trello.com/c/4jmBp57z/149-rem-mixin)

PR #540, opened 2018-06-13
Exposed initTranslation() and allowed polyglot prop for I18n

goldoraf:
In Drive, we want to have a `t` instance in the redux context. Therefore, we need to init i18n first to get our polyglot instance, and then pass it to the `I18n` provider. Thus these changes ;)

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -20,9 +20,10 @@ export class I18n extends Component {
   }
 
   init(props) {
-    const { lang, dictRequire, context, defaultLang } = props
+    const { polyglot, lang, dictRequire, context, defaultLang } = props
 
-    this.translation = initTranslation(lang, dictRequire, context, defaultLang)
+    this.translation =
+      polyglot || initTranslation(lang, dictRequire, context, defaultLang)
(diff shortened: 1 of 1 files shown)


react/I18n/index.jsx
```diff
@@ -20,9 +20,16 @@ export class I18n extends Component {
...
   init(props) {
-    const { lang, dictRequire, context, defaultLang } = props
-
-    this.translation = initTranslation(lang, dictRequire, context, defaultLang)
+    const {
+      polyglot,
+      lang,
+      dictRequire,
+      context,
+      defaultLang = DEFAULT_LANG
```
ptbrowne: Shouldn't the DEFAULT_LANG be configured as a defaultProp ?

goldoraf: Hmmm, not sure what we'd get by configuring it as a defaultProp instead of using a default value to be honest, what do you think?

ptbrowne: Other methods would have access to it.

goldoraf: Good point :+1:

Issue #545, opened 2018-06-15
😱 Close button is misplaced in intent modal


gregorylegarec:
Looks like a regression

![capture d ecran 2018-06-15 a 12 07 23](https://user-images.githubusercontent.com/776764/41462943-ee17e80e-7094-11e8-98b7-c9f06f8a3778.png)

GoOz:
really? I can't see it on styleguidist. :/

ptbrowne:
Can you give use more information on how to reproduce the problem ?

gregorylegarec:
@ptbrowne You can have a look at http://recette-interapp-sandbox.cozy.works

GoOz:
Ok I found the issue. It's related to #335. Giving more weight to the selector should fix this, temporarily at least.

ptbrowne:
Has been fixed by https://github.com/cozy/cozy-ui/pull/546

Issue #335, opened 2018-02-14
Find out why order of the CSS is not preserved


ptbrowne:
We have experienced problems with buttons that needed more specificity since order of the CSS rules was not preserved. Thanks @enguerran for the description below.

We need to find out why the order is not preserved in contacts. 

# Contexte

Dans `cozy/cozy-contacts`, on utilise un bouton (`import { Button } from "cozy-ui/react/button"`) dans un menu (`import { Menu } from "cozy-ui/react/Menu"`). À ce bouton, on lui ajoute la classe `narrow` via l'attrinut `extension` :

```
<Button theme="secondary" extension="narrow">
  <Icon icon="dots" />
</Button>
```

Voir la doc 
![](https://i.imgur.com/yHlyeT2.png)

Mais dans l'app contacts, le css embarqué n'est pas dans le bon ordre et la règle de la classe `c-bt--narrow` est écrasée par la version de `c-btn` comme le montre cette capture

![](https://i.imgur.com/p2zUoB9.png)

enguerran:
Related to #328

enguerran:
![](https://github.trello.services/images/mini-trello-icon.png) [(2) ETU je peux supprimer un contact depuis sa fiche](https://trello.com/c/eD8KjGbM/724-2-etu-je-peux-supprimer-un-contact-depuis-sa-fiche)

CPatchane:
Visiblement normalize.css est chargé en plein milieu de tous les autres styles. Le problème a aussi été rencontré dans cozy-store et normalize apparait à la ligne [77](https://github.com/cozy/cozy-store/blob/build/app.a67c9f3f3f08f0957d8f.min.css#L77). En dev, il semble apparaitre ligne 1527 et du coup aussi après le style de la bar.

GoOz:
This is also happening on the close modal button on Styleguidist.

Button styles are loaded before in [Modal](https://github.com/cozy/cozy-ui/blob/master/react/Modal/index.jsx#L5) but are still output after close modal button styles causing a failing override.

y-lohse:
tl;dr : I don't know what the problem is, but use the transpiled version of cozy-ui and you should be fine.

I've spent some time investigating this issue. It doesn't look like I'll get to the bottom of it this time, so leaving notes for the next person who digs into it.

[...]

PR #546, opened 2018-06-15
fix: IntentModal Cross position

GoOz:
Fixes issue #545 by giving more weight to the CSS selector.

Bug related to #335


gregorylegarec:
<3

PR #550, opened 2018-06-19
feat: add layout component

### LINKED ISSUE #549: iPhoneX layout
Some minor tweaks need to be done.

See https://github.com/cozy/cozy-drive/pull/992/files

GoOz:
Added a `<Layout />` component to ease app integration and enforce consistency between different apps.

Preview: https://gooz.github.io/cozy-ui/react/#layout-1

Thorough reviews are very welcome because I'm not 100% sure I dealt with every use case.

Fixes #549


GoOz:
[Ajouter un composant <Layout />](https://trello.com/c/PxhFODC5/19-ajouter-un-composant-layout)

stylus/objects/layouts.styl
```diff
@@ -91,7 +92,8 @@ $app
         display block
 
         main,
-        main > [role=contentinfo]
+        main > [role=contentinfo], // Deprecated
```
ptbrowne: ping @y-lohse @goldoraf seems cozy-photos is using this.

ptbrowne:
Does it support iPhoneX ?

react/Layout/Layout.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styles from './Layout.styl'
+import cx from 'classnames'
+
+export const Layout = ({ children, className, monoColumn, ...rest }) => {
+  return (
+    <div
+      className={cx(
+        monoColumn ? styles['o-layout'] : styles['o-layout--2panes'],
```
ptbrowne: I think o-layout--2panes should not be with `--` as it is not a variant. I would think a variant should be applied additionally to the normal class. In this case, it is one or the other.

react/Layout/Layout.md
```diff
@@ -0,0 +1,59 @@
...
+        </NavItem>
+      </Nav>
+    </Sidebar>
+    <Main>
+      <Content className='u-p-1'>
+        { content.ada.short }
+      </Content>
+    </Main>
+</Layout>
+```
```
ptbrowne: It is not possible to have it as a real example ?

stylus/objects/layouts.styl
GoOz: Actually almost every app uses this, that's why I'm just commenting the deprecated status and not removing it completely yet.

GoOz:
It does not. That wasn't the purpose of this PR but I guess I could add this too.

react/Layout/Layout.jsx
GoOz: true. It used to be a variant but I didn't change the naming. 👍

react/Layout/Layout.md
GoOz: I don't get it, what do you mean? Like real content like with one of our apps?

stylus/objects/layouts.styl
y-lohse: Is it enough to just remove the `[role=contentinfo]` ? If so that would be a quick change to do everywhere.

y-lohse:
I also think this would be a good place to integrate some iphone x changes :)

stylus/objects/layouts.styl
GoOz: It's more replacing than removing but basically yeah, that's it.

react/Layout/Layout.md
ptbrowne: sorry i was confused with the JSX code, I thought the example was not rendered, it's ok.

GoOz:
@y-lohse Does it look good now?

GoOz:
@y-lohse Time will tell… ;)

PR #555, opened 2018-06-28
Fix subtle button link

goldoraf:
This PR fixes the color on visited button links and adds a new "Cozy dark" (or negative) icon.


src/icons.jsx
```diff
@@ -14,6 +14,7 @@ const importIcons = function () {
     'icon-check-circleless',
     'icon-clock',
     'icon-cozy',
+    'icon-cozy-dark',
```
GoOz: May I suggest to call this icon `icon-cozy-full` or something like that?
`Dark` suggest a sense of tone which is not really the case. You could also call it `icon-cozy-negative` if you will, just something not "color" related.

goldoraf: I was referring to dark themes, but ok for `negative`

GoOz: Sure but it could be misunderstood by others :)

PR #562, opened 2018-07-06
fix: Allow capital letters in git messages

GoOz:
Small fix to allow capital letters anywhere in the git commit message. 
You can actually use lowercase, camelCase, or whatever case you want.


enguerran:
I don't know commitlint, but it seems fine to me.

PR #568, opened 2018-07-11
feat: add header Icon for Modal & IntentModal

### LINKED ISSUE #561: [Modal] Icon title effect
Add effect for modals with no title but a big icon instead. Scrolling within the modal pushes a smaller icon into the title area.

GoOz:
Added Header Icon on Modal & IntentModal
Also there's a new style for functionnal Intent Modal (app to app), the background is not grey anymore.

Preview: [Modal](https://gooz.github.io/cozy-ui/react/#modal) *(Header  Icon section)* & [IntentModal](https://gooz.github.io/cozy-ui/react/#intentopener)

⚠️ **2 things are missing, hence the WIP:**
~~**- I don't have the good icon to notify that it is a service yet**~~
**- I, apparently, need to do a change on Cozy-Client-JS so I can style `html` & `body` elements in iframe.** Although this should not prevent this PR from merging since this change in Cozy-Client-JS is an enhancement from the actual behaviour.

Fixes #561


react/Modal/index.jsx
```diff
@@ -9,9 +9,94 @@ import migrateProps from '../helpers/migrateProps'
...
+                  )}
+                </div>
+                <Icon
+                  icon="exchange"
+                  className={styles['c-header-exchange']}
+                  color={palette['coolGrey']}
+                />
+                <Icon icon={iconDest} className={styles['c-header-icon']} />
+              </h2>
+              <div
```
GoOz: If anyone has a better idea to duplicate an element without explicitly duplicate it, I'm all ears. I couldn't find one but maybe someone has a better idea.
NB: I need the `h2` to be duplicated so I can use it without interfering with the flow.

react/Modal/index.jsx
```diff
@@ -9,9 +9,94 @@ import migrateProps from '../helpers/migrateProps'
...
-  <div className={cx(styles['c-modal-content'], className)}>{children}</div>
-)
+class ModalContent extends Component {
+  constructor(props) {
+    super(props)
+    this.state = { displayGhostHeader: false }
+  }
+
+  componentDidMount() {
+    this.scrollingContent.addEventListener('scroll', this.handleScroll, {
```
ptbrowne: The event listener should be removed in `componentWillUnmount`.

GoOz: Indeed.

react/Modal/index.jsx
```diff
@@ -9,9 +9,94 @@ import migrateProps from '../helpers/migrateProps'
...
+        className={cx(styles['c-modal-content'], className)}
+        ref={div => {
+          this.scrollingContent = div
+        }}
+      >
+        {iconSrc &&
+          iconDest && (
+            <div>
+              <h2 className={cx(styles['c-modal-header--icon'], className)}>
+                <div>
```
ptbrowne: Why this extra div ?

GoOz: Because if I don't add it, I get the error `Syntax error: Adjacent JSX elements must be wrapped in an enclosing tag`

GoOz: ah no sorry, not this one, the other one inside. You're right, that's a leftover from a previous attempt. It is indeed useless.

PR #571, opened 2018-07-16
[Do not merge] fix: Only hide action menu on overlay click

y-lohse:
- What we expect: when single-taping the overlay, the menu disappears.
- What actually happens: tap any menu item — the event bubbles to the overlay and the menu is closed.

I consider this a bug fix, but it's possible that people were relying on this bug to close the menu, intentionally or not. Should be label this as a breaking change? 

In cozy-drive, we're good.

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -124,13 +124,15 @@ class ActionMenu extends Component {
   }
 
   handleMenuRef = menu => {
-    // FIXME
     this.menuNode = ReactDOM.findDOMNode(menu) // eslint-disable-line react/no-find-dom-node
   }
 
   handleWrapperRef = wrapper => {
-    // FIXME
     this.wrapperNode = ReactDOM.findDOMNode(wrapper) // eslint-disable-line react/no-find-dom-node
+  }
+
+  handleOverlayRef = overlay => {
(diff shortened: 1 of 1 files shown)


ptbrowne:
DO you mean that now you can click on a menu item and the menu does not close itself ?

y-lohse:
Yep, that's what it would mean. What option makes more sense:

1. Taping anything in the menu closes it — if you want to keep it open after a tap on an item, you need to intercept the event
2. Taping a menu item doesn't close the menu — you have to close it yourself when you want it

Number 2 makes more sense for me, but number 1 is practical if you have a menu with simple actions.

y-lohse:
Discussed together via zoom: we keep the current behavior, but if you don't want to close the menu, you should `preventDefault()` the event in your subcomponents. And in cozy-ui in the bar, we check `isDefaultPrevented` and only close thee menu it it hasn't been prevented. This will make the change non-breaking.

ptbrowne:
@y-lohse , I think there is still work to do on this PR ?

y-lohse:
Yes there is, i updated the title to reflect that.

ptbrowne:
I removed myself as a reviewer to keep my list of review clean. Please re-add me when it is ready.

y-lohse:
This hasn't come up again in a looong time, so probably not useful.

PR #578, opened 2018-07-31
fix: modal update

ptbrowne:
* this.props is old props in componentWillUpdate
* no need to use state for computed properties
* modified animated header example to use state (so that we can see the bug if it comes out again)

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -32,11 +32,7 @@ class ModalContent extends Component {
   constructor(props) {
     super(props)
     const { children } = this.props
-    const animatedHeader = _getAnimatedHeader(children)
-    this.animatedHeader = animatedHeader
-    this.childrenToRender = animatedHeader
-      ? _getChildrenToRender(children)
-      : children
+    this.refreshComputedParts(children)
     this.state = {
       displayGhostHeader: false
     }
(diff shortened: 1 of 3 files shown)


react/Modal/index.jsx
```diff
@@ -86,8 +83,8 @@ class ModalContent extends Component {
   render() {
     const { className } = this.props
 
-    const { displayGhostHeader, animatedHeader, childrenToRender } = this.state
```
CPatchane: By using `this` if we have an update in the `animatedModalHeader` component it won't update the content here :/. It's why I used the state finally.

react/Modal/index.jsx
```diff
@@ -42,8 +42,8 @@ class ModalContent extends Component {
     }
   }
 
-  componentWillUpdate() {
-    const { children } = this.props
+  componentWillUpdate(nextProps) {
+    const { children } = nextProps
```
CPatchane: Good shot! Thanks 👍

CPatchane:
[[STORE] La croix sur la modale est décallée avec le clic](https://trello.com/c/xyz2wVhE/1152-store-la-croix-sur-la-modale-est-d%C3%A9call%C3%A9e-avec-le-clic)

gregorylegarec:
[[STORE] La croix sur la modale est décallée avec le clic](https://trello.com/c/xyz2wVhE/1152-store-la-croix-sur-la-modale-est-d%C3%A9call%C3%A9e-avec-le-clic)

PR #579, opened 2018-07-31
fix(Button): Darken secondary subtle button for readability

### LINKED ISSUE #566: Change subtle button grey
Hi,

Here is the component : https://gooz.github.io/cozy-ui/react/#button

The grey is too pale for the secondary subtle button, it needs a --coolGrey color instead of --silver

GoOz:
A secondary single text on white background, so without the "box" of a button, seemed way too bright for a good readability, so I darkened it according to @Claiw's recommendation.

Fixes #566

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -521,7 +521,9 @@ $button-subtle--regular
     themedBtnSubtle(regularTheme)
 
 $button-subtle--secondary
-    themedBtnSubtle(secondaryTheme)
+    // Not using themedBtnSubtle(secondaryTheme) because
+    // silver color for a single text was too bright
+    color coolGrey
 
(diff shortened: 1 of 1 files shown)


stylus/components/button.styl
```diff
@@ -521,13 +521,15 @@ $button-subtle--regular
     themedBtnSubtle(regularTheme)
 
 $button-subtle--secondary
-    themedBtnSubtle(secondaryTheme)
+    // Not using themedBtnSubtle(secondaryTheme) because
```
GoOz: Is the comment too much? I wondered, since it's kind of an exception, it seemed legit.

CPatchane: I would keep that comment IMHO 👍

PR #581, opened 2018-08-02
Add t or nothing helper in I18n

CPatchane:
This feature is useful in the Store to know if there is some text to display or not since the `t()` will return the key by default if the local doesn't exist.

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -30,6 +30,11 @@ export class I18n extends Component {
   getChildContext() {
     return {
       t: this.translation.t.bind(this.translation),
+      tn: (key, options, fallback = null) => {
+        return this.translation.has(key)
+          ? this.translation.t(key, options)
+          : fallback || ''
+      },
       f: this.format,
       lang: this.props.lang
     }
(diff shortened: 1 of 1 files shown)


CPatchane:
The name can also be `tOr('key', options, 'default text')` instead of `tn`. The naming is open to discussion :)

CPatchane:
[Nettoyer la gestion des locales dans le Store](https://trello.com/c/A42CSX9K/1162-nettoyer-la-gestion-des-locales-dans-le-store)

gregorylegarec:
Thanks @CPatchane 💯

ptbrowne:
Je vois ca dans la doc de polyglot http://airbnb.io/polyglot.js/. Ca ne couvre pas le besoin ?

> If you like, you can provide a default value in case the phrase is missing. Use the special option key “_” to specify a default.

```
polyglot.t("i_like_to_write_in_language", {
  _: "I like to write in %{language}.",
  language: "JavaScript"
});
=> "I like to write in JavaScript."
```

CPatchane:
Hum, this `_` feature was missing the last time I looking for it, it was added [a month ago](https://github.com/airbnb/polyglot.js/blame/master/docs/polyglot.html#L544). We can use t("my.locale", { _: ""}) to get this same behavior indeed.
Should we revert this PR so? Sorry for not checking again before, I was looking for a specific method instead of the `t` itself :/

ptbrowne:
OK to revert

PR #593, opened 2018-08-08
fix: Import needs the filename

enguerran:
Or webpack needs a configuration to find the correct filename.

### DIFF
--- a/react/Text/index.jsx
+++ b/react/Text/index.jsx
@@ -3,13 +3,12 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
 
-const mkText = baseClass => props => {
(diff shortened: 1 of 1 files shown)


ptbrowne:
JSX are already imported, it is configured here :  https://github.com/cozy/cozy-ui/blob/master/docs/webpack.config.js#L9

I have not seen any bug during the tests of the Text docs on my machine. Have you seen any bug ?

ptbrowne:
`./index.jsx` could be transformed into `.` though

GoOz:
I have absolutely no opinion because of my lack of knowledge on this topic but may I suggest, before doing anything, to rebase from master because I just merge #592 and you may get conflicts.

PR #596, opened 2018-08-08
Add AppIcon component

gregorylegarec:
Since app icons are now provided as blob by the stack, we have to load them asynchronously.

`<AppIcon />` component uses Cozy Client in global context to fetch the icon.

It display a loading state, similar to what's already exists in Store and Drive, and handle errors by providing a default icon if something goes wrong.

`fetching` state
capture d ecran 2018-08-08 a 17 53 42

`failed` state
capture d ecran 2018-08-08 a 17 55 08

`done` state
capture d ecran 2018-08-08 a 17 52 38


react/AppIcon/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+    const cozyClient = client && client.link && client.link.client
+    if (!cozyClient || typeof cozyClient.fetch !== 'function') {
+      console.warn(
+        'AppIcon needs CozyClient and to be called into a <CozyClientProvider /> component.'
+      )
+      this.setState({ status: 'failed' })
+      return
+    }
+    const icon = await this.fetchIcon(app.links.icon, cozyClient)
+    this.setState({ icon, status: icon ? 'done' : 'failed' })
```
ptbrowne: IMHO, all this should be done in fetchIcon. It should not be the responsibility of `componentDidMount` to do that. 

```
componentDidMount() {
  this.fetchIcon()
}
```

This let the fetchIcon be the only responsible for updating the `status` state property.

react/AppIcon/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+    } catch (error) {
+      console.warn(error.message)
+      return
+    }
+    // check if MIME type is an image
+    if (!icon.type.match(/^image\/.*$/)) {
+      console.warn(`App icon ${url} is not an image.`)
+      return
+    }
+    return URL.createObjectURL(icon)
```
ptbrowne: IMHO all the functionality that is not connected to the component should be extracted in a helper method.

```
async fetchIcon() {
  this.setState({ status: 'fetching' })
  try {
      this.props.fetchIcon(this.props.app)
      this.setState({ status: 'done' })
  } catch (e) {
      this.setState({ status: 'error' })
  }
}
```

This would let the easier change of implementation of fetchIcon.

react/AppIcon/index.jsx
```diff
@@ -0,0 +1,88 @@
...
+    const { icon, status } = this.state
+    switch (status) {
+      case 'idle':
+      case 'fetching':
+        return (
+          <div
+            className={classNames(
+              styles['c-loading-placeholder'],
+              styles['c-app-icon'],
+              styles['c-app-icon-placeholder']
```
ptbrowne: I wonder if these 3 classes could be replaced by only 1 ? What happens if one is removed ? Do they need to be used together or have they orthogonal concerns ?

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+  it(`renders default icon when no Cozy Client`, () => {
+    const component = shallow(<AppIcon app={app} />).getElement()
+    expect(component).toMatchSnapshot()
+  })
+
+  it(`renders default when fetch error occurs`, () => {
+    context.client.link.client.fetch.mockImplementation(() => {
+      throw new Error('Mocked error')
+    })
```
ptbrowne: If the `fetchIcon` was passed as a property, it would be easier to mock and the tests would not have to know about the inner workings of the client.

IMHO there should be an

```
<DumpAppIcon />
class ClientAppIcon extends React.Component {
  constructor () {
    super(props, context)
    this.fetchIcon = this.fetchIcon.bind(this) 
  }

  fetchIcon () {
    const client = this.props.context
    // dealing with cozy.links, mime type etc... 
  }

  render () {
    return <DumbAppIcon fetchIcon={this.fetchIcon}/> 
  }
}
```

react/AppIcon/styles.styl
```diff
@@ -0,0 +1,36 @@
...
+    height 100%
+    width 100%
+}
+
+.c-app-icon-default {
+    color silver
+}
+
+.c-app-icon-placeholder {
+    animation-delay: 0.1s;
```
ptbrowne: Should this class be merged with `.c-loading-placeholder` ?

Maybe stylus composition features could be used : 

```
.c-app-icon-placeholder {
  @extends .c-loading-placeholder
  animation-delay: 0.1s;
}
```

This way `AppIcon` would only need to apply `.c-app-icon-placeholder` instead `.c-loading-placeholder` and `.c-app-icon-placeholder`.

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -0,0 +1,108 @@
+'use strict'
+/* eslint-env jest */
+
+import React from 'react'
+import { configure, shallow } from 'enzyme'
+import Adapter from 'enzyme-adapter-react-15'
+
+import AppIcon from '../'
+
+configure({ adapter: new Adapter() })
```
ptbrowne: Jest is configured for all the tests to use the React 15 adapter. 

https://github.com/cozy/cozy-ui/blob/98ed325f89ff04193df8cf6614a0e1250b28b55c/test/jestsetup.js#L2-L4

There is no need to `configure()` in each test ;)

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    ok: true,
+    blob: jest.fn().mockResolvedValue(
+      new Blob([iconSvg], {
+        type: 'text/plain'
+      })
+    )
+  }
+
+  beforeEach(() => {
+    jest.resetModules()
```
ptbrowne: Why do we need this here ?

react/AppIcon/index.jsx
gregorylegarec: I don't see the benefits here. Why extracting this method now ? If `this.props.fetchIcon` is already injected we can change its implementation.

ptbrowne: Yes, "helper" was perhaps not the right word, if it is extracted from the component, that's alright ;)

react/AppIcon/index.jsx
gregorylegarec: Actually, `c-app-icon` is used for all icon "states": `done`, `fetching` and `failed`, even if its content change. Both `(.)*-placeholder` classes are used for fetching animation. In the store all the loading animations have a common basis, which has been used here for `c-loading-placeholder`, and a specific `animation-delay` for different elements. This specific part has been kept here into `c-app-icon-placeholder`, but you're right we could actually merge both of theme.

Loading animation need to be integrated into Cozy UI, but it's not a simple work. When it will be done we will update this part.

For now, I merge those two `.*-placeholder` classes.

react/AppIcon/test/AppIcon.spec.js
gregorylegarec: Mmmmmh no need actually. Copy-pasted code. Thanks !

react/AppIcon/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+  async componentDidMount() {
+    this.fetchIcon()
+  }
+
+  async fetchIcon() {
+    const { app, fetchIcon } = this.props
+
+    if (typeof this.props.fetchIcon !== 'function') {
+      throw new Error('fetchIcon prop is required')
+    }
```
ptbrowne: MINOR

I think this is already checked by the PropTypes. 

`fetchIcon` is destructured above so we can remove `this.props.`

react/AppIcon/index.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+    this.setState({ status: 'fetching' })
+    let icon
+    try {
+      icon = await fetchIcon(app.links.icon)
```
ptbrowne: Minor: no need to `let icon` just above since it is not used outside the `try` block.

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -0,0 +1,45 @@
...
+import AppIcon from '../'
+
+describe('AppIcon component', () => {
+  const app = {
+    links: {
+      icon: 'apps/test/icon'
+    }
+  }
+
+  const successFetchIcon = jest.fn().mockResolvedValue('data://test-url')
```
ptbrowne: TIL mockResolvedValue

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+  it(`renders as loading`, () => {
+    const component = shallow(
+      <AppIcon app={app} fetchIcon={successFetchIcon} />
+    ).getElement()
+    expect(component).toMatchSnapshot()
+  })
+
+  it('renders app icon after fetching', done => {
+    const wrapper = shallow(<AppIcon app={app} fetchIcon={successFetchIcon} />)
```
ptbrowne: Minor: I think getting rid of the setTimeout would be better.

http://lea.verou.me/2016/12/resolve-promises-externally-with-this-one-weird-trick/

```js
function deferred() {
	let res, rej

	const promise = new Promise((resolve, reject) => {
	    res = resolve;
		rej = reject;
	});

	promise.resolve = res
    promise.reject = rej
	return promise;
}

it('renders app icon after fetching', done => {
	const def = deferred()
	const wrapper = shallow(<AppIcon app={app} fetchIcon={() => def} />)
	def.resolve()
	wrapper.update()
	const component = wrapper.getElement()
	expect(component).toMatchSnapshot()
})
```

gregorylegarec: I tested this kind of code when writing this test. I also tested your code but I have not been able to make it work. I guess it is too "quick" maybe ? The only way to make it work was using a `setTimeout`, and God knows I __HATE__ `setTimeout`.

By the way, if you manage to make it works, I am interested by the solution !

ptbrowne: It would not work with the deferred solution. The reason is that the callbacks for resolution/rejection of a promise are called "asynchronously".

Good article on the subject : https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/

gregorylegarec:
@ptbrowne Well done ! I tried something like this but never succeed. Test updated in #599

Issue #453, opened 2018-04-27
Modal don't show anymore after ActionMenu has been opened


drazik:
There is a very strange bug with Modal and ActionMenu. Here are the steps to reproduce : 

1. Open an ActionMenu
2. Close it
3. Open a Modal

The Overlay is shown, but the Modal is not present. Actually it is rendered but outside the viewport.

You can reproduce in the styleguide : https://cozy.github.io/cozy-ui/react/

ptbrowne:
It's no longer the case, may have been fixed in #602

PR #602, opened 2018-08-27
fix: Reset menu node style on unmount

y-lohse:
We've had a weird and hard to reproduce bug for a while, most recently described [here](https://github.com/cozy/cozy-drive/issues/1124).

Sometimes modals in drive would start with a `transitionY: 110%`. Commenting out the lines in the ActionMenu that apply the transition would solve the problem.

I suspect that preact is re-using the DOM node from the ActionMenu for the modal, and for some reason the style prop remains attached. Clearing it before the component is unmounted solves the problem.


y-lohse:
[🐛 supprimer 2 fichiers de suite frise l"app](https://trello.com/c/s8Jd4sDW/1119-%F0%9F%90%9B-supprimer-2-fichiers-de-suite-frise-lapp)

ptbrowne:
> and for some reason the style prop remains attached

Must be because the style prop is not tracked in the virtual dom of preact.

I guess the clean way of attaching the style prop would be to do it in JSX...

Could you add a comment explaining all this ? Or just pointing to this issue ?

GoOz:
Could it be related with #453 ?

y-lohse:
@ptbrowne Yeah sure, I'll do that.

@GoOz Yep, sounds like the same problem. cc @drazik

drazik:
Yes it seems related. Good catch @y-lohse

Issue #601, opened 2018-08-27
Enhance a11y


GoOz:
Several A11y issues have been spotted and Cozy-UI could benefit some basic audit at least.

Issues spotted so far:
- [x] Button or Links as button should have the `role="button"`
- [x] SVG icon should be readable when informative or hidden when decorative
- [x] Alert should have `role="alert"`
- [x] Accordion's header should have `role="button"` instead of `role="tab"` and other few tweaks
- [x] Check for accurate roles of landmark regions
- [ ] ~~Add roles on menu & menu item~~ Will be done when using M-UI's menu
- [x] Add `role` & `aria-label` on toolbar (Selection Bar)
- [x] Accessible Tabs
- [x] Accessible Modal

PR #604, opened 2018-08-28
feat: Make Button component more accessible

GoOz:
Refs  #601

**BREAKING CHANGE:**
- `label` props is now required
- `iconOnly` props is required if you want to get a button with only an icon

So allow me to explain a little bit.

### What we need

We need buttons to always, in any way, display a label explicitely describing what the actions of the button does.

### How to do it
There's two cases when you consider having an icon in your button, and either way the icon must be hidden with `aria-hidden` set to `true`, but there's some differences:
1. The Icon is with a text. Nothing special to do here. It's like before. 
2. The Icon is on its own. In this case, as the icon is hidden, we must add some label for screen readers by using the same DOM for a normal button except we're hiding it visually; then you should also have a `title` on your button so people that don't recognize the icon can read the tooltip by hovering it and understand what the action does. Also, as the `label` is always needed and therefore mandatory, we have no way of knowing when we want to display a button with only an icon, hence the addition of the `iconOnly` props.

I couldn't find another way to achieve that. If any of you has another suggestion, I'm all ears.

### DIFF
--- a/react/Button/index.jsx
+++ b/react/Button/index.jsx
@@ -42,6 +42,7 @@ const BaseButton = props => {
   const {
     children,
     icon,
+    iconOnly,
     label,
     subtle,
     className,
@@ -54,6 +55,9 @@ const BaseButton = props => {
   } = props
 
   const transformProps = tagToTransformProps[Tag] || identity
+  const tooltip = iconOnly ? label : null
+  const iconOnlyClass = iconOnly ? 'u-visuallyhidden' : null
+
   return (
     <Tag
       {...transformProps(restProps)}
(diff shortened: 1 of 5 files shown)


react/Button/index.jsx
```diff
@@ -65,10 +69,15 @@ const BaseButton = props => {
         className,
         variant: subtle && 'subtle'
       })}
+      title={tooltip}
     >
       <span>
-        {Icon.isProperIcon(icon) ? <Icon icon={icon} /> : icon}
-        {label && <span>{label}</span>}
+        {Icon.isProperIcon(icon) ? (
+          <Icon icon={icon} aria-hidden focusable="false" />
```
GoOz: The `focusable="false"` attribute is for telling IE to never focus the SVG.

CPatchane:
@GoOz why `label` props should be required? We could use the button add pass the content/text using the React childrend, couldn't we?

GoOz:
> why label props should be required? We could use the button add pass the content/text using the React children, couldn't we?

Well I figured that since using `children` is deprecated it should be all about `label` but I guess we can check if either `children` or `label` is here and use it but in that case I don't know how to "require" one or the other.

CPatchane:
Using `children` is deprecated? I think I missed out this information :/
How should we use a custom icon in the button then without children?

GoOz:
You can't like this? 
`<Button theme="danger" icon={ <Icon icon='delete' color='yellow' /> } label='delete' />`
but instead of `icon="delete"` you pass the object of the SVG.
No?

CPatchane:
I will test it out for this case, I don't use children anymore in Store so I won't blocking this PR

PR #606, opened 2018-08-29
fix: Better a11y on Accordion

GoOz:
Previously, Accordion was using Tabs logic for accessibility.
This PR cleans this up.

Refs #601

### DIFF
--- a/react/Accordion/index.jsx
+++ b/react/Accordion/index.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+import uniqueId from 'lodash/uniqueId'
 import cx from 'classnames'
 import Icon from '../Icon'
 import styles from './styles.styl'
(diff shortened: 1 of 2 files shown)


react/Accordion/index.jsx
```diff
@@ -15,9 +16,9 @@ class AccordionItem extends Component {
 
     const { selected } = this.state
 
-    const removeSpaceCharacters = str => str.replace(/\s+/g, '')
+    const headingID = uniqueId('accordion_header_')
 
-    const bodyName = removeSpaceCharacters(label)
+    const contentID = uniqueId('accordion_content_')
```
ptbrowne: I just realized that the uniqueId is going to change at each render, it would be better to compute it at mount time and reuse it.

```
componentDidMount() {
  this.contentID = uniqueId('accordion_content')
}

render() {
   ...
   <div id={this.contentId} />
    ....
}
```

The same technique should be applied to the other PRs.

PR #607, opened 2018-08-30
fix: Add role="navigation" on Nav

GoOz:
Refs #601


ptbrowne:
Snapshot must be updated. LGTM sinon ! Merci

PR #608, opened 2018-09-03
fix: Made Modal accessible

GoOz:
Refs: #601

### DIFF
--- a/react/Modal/Readme.md
+++ b/react/Modal/Readme.md
@@ -20,7 +20,7 @@ initialState = { modalDisplayed: false};
   <button onClick={()=>setState({ modalDisplayed: !state.modalDisplayed })}>
     Toggle modal
   </button>
-  {state.modalDisplayed && <Modal  description={content.ada.short} dismissAction={() => setState({ modalDisplayed: false })} /> }
+  {state.modalDisplayed && <Modal description={content.ada.short} aria-label="Accessible title" dismissAction={() => setState({ modalDisplayed: false })} /> }
 </div>
 ```
 
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -8,6 +8,7 @@ import Icon from '../Icon'
 import migrateProps from '../helpers/migrateProps'
 import palette from '../../stylus/settings/palette.json'
 import Portal from 'preact-portal'
+import uniqueId from 'lodash/uniqueId'
 
 class AnimatedContentHeader extends Component {
   render() {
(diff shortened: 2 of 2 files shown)


react/Modal/index.jsx
```diff
@@ -158,7 +161,12 @@ const ModalHeader = ({
 }) => {
   const isTitle = typeof children === 'string'
   return (
-    <div className={cx(styles['c-modal-header'], className)} style={style}>
+    <div
+      className={cx(styles['c-modal-header'], className)}
+      style={style}
+      id={titleId}
+      tabIndex="-1"
```
GoOz: `tabIndex="-1"` is to make sure that when the modal takes the focus, reading began at the beginning of the modal and not the next focusable element (inside the modal) which is the behaviour by default.

ptbrowne: Why do we have to put an id ?

GoOz: In order to assign a title to the modal, having a `<hn>` isn't always enough apparently.

drazik:
Is there plans to trap the focus inside the `Modal` when the user opens it ? Currently we can tab behind it.

GoOz:
@drazik according to what I read, the fact that the modal has `role="dialog"` and `aria-modal=true` is enough to give the focus and ignoring the background content. Before `aria-modal=true` from WAI 1.1, you had to add `aria-hidden` on content behind to ensure this behaviour, but not anymore.

react/Modal/index.jsx
ptbrowne: Assigning a title, what for ?

GoOz: It's a label, to be read, so the reader knows what the content of the modal is all about.

GoOz: There's `aria-labelledBy={titleId}` on the Modal, L320
I didn't add `aria-describedby` as it is optional, contextual and in many case, not relevant.
As for the unique ID being generated outside the modal, if I generate it inside the Modal, how can `ModalHeader` can use it too then?

ptbrowne: I did not understand that aria-labelledBy was a reference and not the text to be read.

`titleId` should be generated in the Modal render and passed to ModalHeader though.

drazik:
Oh, OK :) Thanks @GoOz

react/Modal/Readme.md
```diff
@@ -20,7 +20,7 @@ initialState = { modalDisplayed: false};
   <button onClick={()=>setState({ modalDisplayed: !state.modalDisplayed })}>
     Toggle modal
   </button>
-  {state.modalDisplayed && <Modal  description={content.ada.short} dismissAction={() => setState({ modalDisplayed: false })} /> }
+  {state.modalDisplayed && <Modal description={content.ada.short} arial-title="Accessible title" dismissAction={() => setState({ modalDisplayed: false })} /> }
```
ptbrowne: arial -> aria

GoOz: Wow that's a 100% pure *typo* 😆
Nice catch, thanks.

PR #610, opened 2018-09-04
Catch context locales missing error

CPatchane:
Since `cozy-home` is using context with the `I18n` component, a missing locales file was not caught correctly and was breaking the whole application (blank page).
It should be fixed by this PR.

### DIFF
--- a/react/I18n/translation.jsx
+++ b/react/I18n/translation.jsx
@@ -27,8 +27,12 @@ export const initTranslation = (
 
   // Load context locales
   if (context) {
-    const dict = dictRequire(lang, context)
-    _polyglot.extend(dict)
+    try {
+      const dict = dictRequire(lang, context)
+      _polyglot.extend(dict)
+    } catch (e) {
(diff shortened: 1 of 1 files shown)


CPatchane:
[[Home] Page blanche avec certaines locales](https://trello.com/c/r5oVShqu/888-home-page-blanche-avec-certaines-locales)

react/I18n/translation.jsx
```diff
@@ -27,8 +27,12 @@ export const initTranslation = (
 
   // Load context locales
   if (context) {
-    const dict = dictRequire(lang, context)
-    _polyglot.extend(dict)
+    try {
+      const dict = dictRequire(lang, context)
+      _polyglot.extend(dict)
+    } catch (e) {
+      console.warn(`The context phrases for "${lang}" can't be loaded`)
```
gregorylegarec: Sorry for the late comment but what about a more explicit message, containing the `context` information ?

I suggest :
> ```js
`The context ${context} cannot be loaded for lang ${lang}
```

PR #612, opened 2018-09-10
Upgrade svg-sprite-loader for webpack 4

CPatchane:
Required to handle svg icons with Webpack 4
~~Should not break the webpack 3 compatibility. It was the version used by the Store since now.~~

Finally it requires that cozy-ui use Webpack 4 too, I think __it can be considered as a breaking change__ so

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "style-loader": "^0.18.2",
     "stylint": "^1.5.9",
     "stylus-loader": "^3.0.1",
-    "svg-sprite-loader": "^3.2.5",
+    "svg-sprite-loader": "^3.9.2",
     "travis-deploy-once": "^5.0.0",
     "webpack": "^3.5.6"
   },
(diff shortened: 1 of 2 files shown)


ptbrowne:
Does it mean that every app that want to use CozyUI needs to have webpack 4 ?

CPatchane:
@ptbrowne During my tests, the svg loader in cozy-ui seems to use the webpack from 'cozy-ui' instead of the webpack from the application. 
That being said, I linked the webpack 4 cozy-ui in the current version of the Store and the Home, I don't have any issues about icons from cozy-ui. I think things are sufficiently separated and work well except that Webpack 4 needs loaders using webpack 4 :/

CPatchane:
[Utiliser Webpack 4 dans cozy-ui](https://trello.com/c/wfRXskej/1227-utiliser-webpack-4-dans-cozy-ui)

PR #615, opened 2018-09-18
chore(Modal): 🎨 Renaming props for clarity

### LINKED ISSUE #494: Modal semantic
Some little semantic concerns about the Modal, no emergency at all:

* As we _close_ the modal, in the README we should try to always name our flag `modalOpened` instead of `modalDisplayed`. It could also be a good practice to do so in our apps codebases.
* The close button is still called `cross` in the comment and README, I think it should be updated to `close button` (same remark for properties `crossClassName` and `crossColor`, those names make me remind me concepts like _cross-platform_).

GoOz:
Fixes #494

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -276,8 +276,8 @@ class Modal extends Component {
       dismissAction,
       overflowHidden,
       className,
-      crossClassName,
-      crossColor,
+      closeBtnClassName,
+      closeBtnColor,
       into,
       size,
       height,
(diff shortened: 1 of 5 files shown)


react/Modal/index.jsx
```diff
@@ -436,6 +436,8 @@ ModalContent.propTypes = {
 
 const EnhancedModal = migrateProps([
   { src: 'withCross', dest: 'closable' }, // withCross -> closable
+  { src: 'crossClassName', dest: 'closeBtnClassName' }, // withCross -> closable
+  { src: 'crossColor', dest: 'closeBtnColor' }, // withCross -> closable
```
gregorylegarec: Wrong comment due to copy/paste.

GoOz: HAHA I did that yesterday but forgot to push :o

PR #616, opened 2018-09-19
fix(Modal): 💄 Modal width 100% on small screens

### LINKED ISSUE #490: Modal width coherence between different media width
When we choose a xxlarge size, this means we need almost the full width, especialy if the screen is narrow, But here the width percentage of the screen is smaller for a small screen than for a wider screen, what is I think the opposite of the need of someone specifiying `size='xxlarge'`.
in the exemple bellow, the second window is a bit smaller and the modal is much smaller what leads to a width percentage much smaller than on a wider window... 
image
I suggest the the width of "xxlarge" modal is a fix percentage of the width whatever the media width and bellow a certain width it turns into a fullscreen modal.

GoOz:
Even if the content of a Modal is not long enough, a Modal will take 100% of the available space until it reaches its own max width

Fixes #490

### DIFF
--- a/stylus/components/modals.styl
+++ b/stylus/components/modals.styl
@@ -73,7 +73,7 @@ for size in 'small' 'medium' 'large' 'xlarge' 'xxlarge'
 
         $max-width=lookup(size + '-width') + lookup(size + '-margin') * 2
         @media (max-width: $max-width)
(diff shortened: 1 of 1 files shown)


ptbrowne:
Thanks !

PR #619, opened 2018-09-26
Avoid conflict with c-app-icon

gregorylegarec:
`c-app-icon` class used in Modal was causing conflict issue and visual regression on cozy-home and bar.

Surprisingly, this issue just appeared recently, and everything was working well at the date of the different commits related to this bug.

capture d ecran 2018-09-26 a 18 39 51

### DIFF
--- a/react/Modal/styles.styl
+++ b/react/Modal/styles.styl
@@ -59,7 +59,7 @@
 .c-app-editor
     @extend $modal-header-app-editor
 
-.c-app-icon
+.c-modal-app-icon
     @extend $modal-header-app-icon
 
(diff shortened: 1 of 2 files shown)


react/Modal/styles.styl
```diff
@@ -59,7 +59,7 @@
 .c-app-editor
```
CPatchane: Maybe we could also change here for more consistency?

gregorylegarec: Yes, but it's not in the scope of this PR and should be done in a cleaning/refacto :)

PR #621, opened 2018-10-03
feat: Added ContextHeader element

y-lohse:
Adds a new component like the header here : https://marvelapp.com/2d65897/screen/39342136

Building the docs is currently broken so I'll update them as soon as they're back working!

### DIFF
--- a/react/ContextHeader/index.jsx
+++ b/react/ContextHeader/index.jsx
@@ -1,0 +1,46 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '../Button'
+import { Title, Text } from '../Text'
+import Icon from '../Icon'
+import styles from './styles.styl'
+import palette from '../../stylus/settings/palette.json'
+import cx from 'classnames'
+
+const ContextHeader = ({ icon, title, text, onClose, className }) => {
+  return (
+    <div className={cx(styles['context-header'], className)}>
+      {icon && <Icon className={styles['context-header-icon']} icon={icon} />}
+      <div className={styles['context-header-content']}>
+        <Title className={false}>{title}</Title>
+        {text && <Text className={false}>{text}</Text>}
+      </div>
+      {onClose && (
+        <Button
+          theme="close"
+          className={styles['c-modal-close']}
+          onClick={onClose}
+          extension="narrow"
+          type="button"
+        >
+          <Icon
+            icon="cross"
+            width="14"
+            height="14"
+            color={palette['coolGrey']}
+          />
+        </Button>
+      )}
+    </div>
+  )
+}
+
+ContextHeader.PropTypes = {
+  icon: PropTypes.node,
+  title: PropTypes.string.isRequired,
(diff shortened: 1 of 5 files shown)


react/ContextHeader/styles.styl
```diff
@@ -0,0 +1,31 @@
...
+  padding 1rem
+  border-radius .3rem
+  display flex
+  flex-grow 1
+  flex-shrink 0
+  flex-direction row
+  justify-content flex-start
+  align-items center
+
+.context-header--content
```
GoOz: Not sure you're using BEM quite right. It's kinda `block-element--modifier` and `content`, `title`, `text` & `icon` are really element, not modifiers.
I would go for something like this `context-header-content` or simply `contextheader-content` which is closer to BEM style but not as easily readable as the first one.

react/ContextHeader/styles.styl
```diff
@@ -0,0 +1,31 @@
+@require 'settings/palette.styl'
+
+.context-header
+  margin 1rem 2rem
```
GoOz: There's a mixin now, to simplify your math. you can just write `margin rem(16 32)`, the unit is implicitely pixel.

react/ContextHeader/index.jsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { Button } from '../Button'
+import Icon from '../Icon'
+import styles from './styles.styl'
+import palette from '../../stylus/settings/palette.json'
+
+const ContextHeader = ({ icon, title, text, onClose }) => {
+  return (
+    <header className={styles['context-header']}>
```
GoOz: `<header>` element is most likely to wrap headings, like titles and subtitles. Not quite sure this component can be related to a header per se. 
I would use a simple `<div>` for this one. It's not because it's the first element of the container that is should be a `<header>`, actually `<header>` is not a required element.

GoOz:
Yes, Webpack 4 broke Styleguidist. Currently trying to fix this.

y-lohse:
[(8) Déplacer un fichier ou un dossier](https://trello.com/c/OSfXpP1q/847-8-d%C3%A9placer-un-fichier-ou-un-dossier)

GoOz:
Styleguidist should be repaired when [this PR](https://github.com/cozy/cozy-ui/pull/622) is merged. You then can rebase your branch from it. :)

y-lohse:
Awesome, thanks! I'll get back to this PR soon.

GoOz:
The code looks good but I'd like to see the updated docs, to test it out, please :3

y-lohse:
Here you go : https://yannick-lohse.fr/cozy-ui/react/#contextheader

GoOz:
Neat! 
Although I think it would be nice to use the `<Text />` component for the title & text of this component (Sorry, I just thought about it). You can give it a `className` prop as well if still needed.
It should go smoothly but if you encounter some issues, just ping me.

PR #623, opened 2018-10-04
feat: ✨ Add Body Scroll Lock for Modals

GoOz:
Added the dependency [Body Scroll Lock](https://github.com/willmcpo/body-scroll-lock) to prevent background scrolling when Modal is displayed and scrollable. 
Essential especially on iOS but useful on other mobile OS and desktop.

For now, it's only used with Modal but could be extended to other components if needed.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -9,6 +9,7 @@ import migrateProps from '../helpers/migrateProps'
 import palette from '../../stylus/settings/palette.json'
 import Portal from 'preact-portal'
 import uniqueId from 'lodash/uniqueId'
+import { disableBodyScroll, clearAllBodyScrollLocks } from 'body-scroll-lock'
 
 class AnimatedContentHeader extends Component {
   render() {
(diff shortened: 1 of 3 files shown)


react/Modal/index.jsx
```diff
@@ -56,12 +58,12 @@ class ModalContent extends Component {
     this.scrollingContent.addEventListener('scroll', this.handleScroll, {
       passive: true
     })
-    document.body.classList.add('has-modal')
+    disableBodyScroll(document.querySelector(this.contentID))
```
Crash--: You don't need to add a new ID. You can use the already defined ref.

Something like : 
```
disableBodyScroll(this.scrollingContent)
```

To quote React documentation: 
```
 Refs are guaranteed to be up-to-date before componentDidMount or componentDidUpdate fires
```

y-lohse:
I think removing the `has-modal` was an accident? It had been introduced [here](https://github.com/cozy/cozy-ui/pull/588) and well, alerts in modals on mobile are now broken :(

y-lohse:
Ok :D Should I make a PR that re-adds the classname, or do we want to do this a different way?

GoOz:
@y-lohse My mistake, I thought this classname had another purpose, something old that I already removed in the past. This should get back as it was before.
I was gonna fix/revert this this morning but if you're on it then go for it. :)

PR #626, opened 2018-10-10
feat(ModalFooter): Option for button alignment

y-lohse:
I'm using a `ModalFooter` with buttons inside of it, but I want to be able to disable one of them, and generally have better control over them — but I also want the usual alignment.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -217,7 +217,11 @@ const ModalCross = ({ onClick, color, className }) => (
   </Button>
 )
 
-const ModalFooter = ({
+const ModalFooter = ({ children, className }) => (
+  <div className={cx(styles['c-modal-footer'], className)}>{children}</div>
+)
+
+const ModalButtons = ({
   secondaryText,
   secondaryAction,
   secondaryType,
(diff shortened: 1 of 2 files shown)


ptbrowne:
I am sorry I did not understand what is the usage and the goal here. I would be reluctant to expand the API of the Modal for alignment purposes.

y-lohse:
No problem, I'll explain in more details.

I have a modal with a footer, with 2 buttons in it — primary and secondary, the usual. But I need to sometimes disable the primary button. This is not something the API currently allows, as we can only control the `theme`, the `onClick` and the `label` of the button.

So I have 2 options

1. I extend the API to add control over the buttons `disabled` property
2. I don't use the props to control my buttons, and instead I pass `children` to the `ModalFooter` and control my buttons myself

I opted for the second solution because I think we just want superficial control over the buttons through props, for more fine grained controlled the `children` option seems more appropriate.

So far so good, but at the moment the content of the `ModalFooter` will only be properly aligned [if we use the `props` to control the buttons](https://github.com/cozy/cozy-ui/pull/626/files#diff-c3e57fd2150089dc0703cdf857587297L229). I think it's a valid use case to have buttons you control yourself, but still benefit from the normal layout.

Is that more clear?

ptbrowne:
OK I understand better, thanks.

I find that the original class implementation is not so great. I think having a ModalButtons component would make the code easier without adding this hasButtonChildren which is here for alignemnt purposes.

In the Modal we would display the buttons inside <ModalButtons /> and in your case, you would wrap the buttons inside <ModalButtons />. This way, the user does not care about the internal of ModalButtons and what it does with respect to alignment, it is just a matter of "semantic" components.

y-lohse:
Ok, I updated the code! It shouldn't really change anything, but it's still labeled as breaking because we had an old, deprecated `ModalButtons` component.

ptbrowne:
Thx !

CPatchane:
@y-lohse it could be useful to have an example in the styleguidist?

PR #631, opened 2018-10-11
Pass correct props to api hack

CPatchane:
Sorry, I didn't see that

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -37,13 +37,13 @@ export class I18n extends Component {
 
   // for preact
   // eslint-disable-next-line react/no-deprecated
-  componentWillReceiveProps() {
-    this.UNSAFE_componentWillReceiveProps()
+  componentWillReceiveProps(nextProps) {
+    this.UNSAFE_componentWillReceiveProps(nextProps)
   }
 
-  UNSAFE_componentWillReceiveProps = newProps => {
-    if (newProps.lang !== this.props.lang) {
-      this.init(newProps)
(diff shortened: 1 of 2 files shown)


react/Modal/index.jsx
```diff
@@ -42,8 +42,8 @@ class ModalContent extends Component {
 
   // for preact
   // eslint-disable-next-line react/no-deprecated
-  componentWillUpdate() {
-    this.UNSAFE_componentWillUpdate()
+  componentWillUpdate(nextProps) {
+    this.UNSAFE_componentWillUpdate(nextProps)
```
Crash--: UNSAFE_componentWillUpdate(nextProps, nextState)

react/I18n/index.jsx
```diff
@@ -37,13 +37,13 @@ export class I18n extends Component {
 
   // for preact
   // eslint-disable-next-line react/no-deprecated
-  componentWillReceiveProps() {
-    this.UNSAFE_componentWillReceiveProps()
+  componentWillReceiveProps(nextProps) {
+    this.UNSAFE_componentWillReceiveProps(nextProps)
   }
```
enguerran: To avoid forgetting to pass the parameters or having to update it if the signature changes, you can just make a variable assignment.

```
class PreactCompatibleComponent extends React.Component {

  componentWillReceiveProps = this.UNSAFE_componentWillReceiveProps;
  componentWillUpdate = this.UNSAGE_componentWillUpdate;

}
```

PR #632, opened 2018-10-11
fix(ContextHeader): Responsive content

y-lohse:
Adjusts the components behavior for small screen sizes and/or long texts.


react/ContextHeader/styles.styl
```diff
@@ -14,10 +15,20 @@
...
 
+  +small-screen()
+      margin rem(8)
+
 .context-header-content
   flex-grow 1
+  flex-shrink 1
+  overflow hidden
+
+.context-header-text
```
GoOz: the `<Text/>` component already has a prop `ellipsis` to allow that, it shouldn't need extra styles for this.

GoOz: https://docs.cozy.io/cozy-ui/react/#text -> Props & Methods section

y-lohse: Brillant, thanks! Missed that.

PR #633, opened 2018-10-12
fix: Alerter not compatible with React

atomiix:
this.base is specific to Preact and was undefined with React


CPatchane:
Thanks @atomiix for opening this PR :)
Indeed, I confirm that React alerts don't disappear correctly without this `ref` fix, good catch 👍 
It doesn't break anything for Preact usage.

Thanks for your contribution! 🌮

PR #635, opened 2018-10-15
fix(MenuButton): Hover background

y-lohse:
Reseting the complete background property means you also loose any background images, positions, etc. Which isn't what we want.


y-lohse:
[(0.5) [Partage] UI marouflage 2](https://trello.com/c/bh2mPBxZ/1212-05-partage-ui-marouflage-2)

PR #636, opened 2018-10-15
Action menu react

ptbrowne:
Tweaks to make the action menu work with React.

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -1,5 +1,4 @@
 import React, { Component } from 'react'
-import ReactDOM from 'react-dom'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 import Hammer from 'hammerjs'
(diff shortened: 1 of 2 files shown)


react/ActionMenu/index.jsx
```diff
@@ -125,13 +129,11 @@ class ActionMenu extends Component {
...
   handleMenuRef = menu => {
-    // FIXME
-    this.menuNode = ReactDOM.findDOMNode(menu) // eslint-disable-line react/no-find-dom-node
+    this.menuNode = menu
   }
 
   handleWrapperRef = wrapper => {
-    // FIXME
-    this.wrapperNode = ReactDOM.findDOMNode(wrapper) // eslint-disable-line react/no-find-dom-node
+    this.wrapperNode = wrapper
```
CPatchane: I think we can now put them directly in the render code

ptbrowne: I prefer not to create functions in the render method. It's not based on scientific benchmarks.

CPatchane: Ok

ptbrowne:
Is it OK ?

PR #637, opened 2018-10-15
feat: Add position fixed to menu

Crash--:
I create this PR just to get some feedbacks about this API. (don't worry, I'll re-add the removed code ;))

Goal: `<Menu>` should be able to position itself either or top or bottom depending on the available  space. 

`Menu` has `position` props `{left|right}`. I have added one : `fixed` (but can be renamed to `autoVertical` or something like else). 

To detect if I have enough space to display menu at the bottom, I use `window.innerHeight` to get the screen height and `domElement.bottom` to get the bottom of my MenuItem. It works great except for one use case :

If my menu is inside a scrollable div and I've scrolled, `domElement.bottom` should be balanced with the scrollTop of my container. 

I can imagine at least two ways of doing it: 
- adding a `containerClassName` props to `Menu`.  This class would refer to the scrollable div. 
- creating a `<MenuContainer>` component to wrap menu & the scrollable div.

 In this PR I've selected the first one, but I would like your advice. I don't really like doing [this](https://github.com/cozy/cozy-ui/pull/637/files#diff-ea25ad7ecae9b5d0ebfaa8f34b3d24d8R54)

Exemple: 
```jsx
const myMenu = () =>  <Menu
            text={
              status === 'ready' && type
                ? t(`Share.type.${type}`)
                : t(`Share.status.${status}`)
            }
            className={styles['recipient-menu']}
            buttonClassName={styles['recipient-menu-btn']}
            position={'fixed'}
            itemsStyle={{
              width: '272px'
            }}
            containerClass={'mySelectContainer'}
          >
            <MenuItem icon={<Icon icon={this.getStatusIcon(status)} />}>
              {status === 'ready' && type
                ? t(`Share.type.${type}`)
                : t(`Share.status.${status}`)}
            </MenuItem>
            <hr />
            <MenuItem onSelect={this.onRevoke} icon={<Icon icon={IconTrash} />}>
              <div className={styles['action-unshare']}>
                {isOwner
                  ? t(`${documentType}.share.revoke.title`)
                  : t(`${documentType}.share.revokeSelf.title`)}
              </div>
              <p className={styles['action-unshare-desc']}>
                {isOwner
                  ? t(`${documentType}.share.revoke.desc`)
                  : t(`${documentType}.share.revokeSelf.desc`)}
              </p>
            </MenuItem>
          </Menu>
```
```jsx 
<div
   className={classnames(
              'mySelectContainer',
              styles['share-modal-secondary']
            )}
          >
<MyMenu />
<OthersComponent />
</div>
```


edas:
Shouldn't you pass a ref to the container as prop instead of a className ?
Or even a function that will give you the numbers you expect (and leave the container compute them) ?

edas:
Alternatively, I expect you _could_ interactively look for each parent in the tree until you find the body or one which is scrollable.

ptbrowne:
> until you find the body or one which is scrollable.

I think you would have to go up the tree, and sum for each element that is scrollable, since you can have several nested scrollable elements.

Crash--:
Thank you for the suggestion. After taking a look at Material-UI, they also go up all the tree. Will go this way. 
https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Popover/Popover.js#L53

Crash--:
Can be reviewed :)

react/Menu/index.jsx
```diff
@@ -132,16 +162,91 @@ class Menu extends Component {
...
+    this.state = {
+      style: { display: 'block', visibility: 'hidden', position: 'fixed' }
+    }
+  }
+  //When the component is ready, than let's do some Maths
+  componentDidMount() {
+    this.handleDisplay()
+  }
+
+  handleDisplay() {
```
ptbrowne: I find handleDisplay not very telling, something like `computeCoordinates` would be clearer IMHO

react/Menu/index.jsx
```diff
@@ -132,16 +162,91 @@ class Menu extends Component {
...
+      <div
+        className={cx(styles['c-menu__inner'], styles['c-menu__inner--opened'])}
+        ref={element => {
+          if (element !== null && element !== this.element) {
+            this.element = element
+          }
+        }}
+        style={({ ...this.props.itemsStyle }, { ...this.state.style })}
+      >
+        {this.props.renderItems()}
```
ptbrowne: Why not render the items in the parent and simply render `this.props.children` here ?

react/Menu/index.jsx
```diff
@@ -42,24 +43,28 @@ class Menu extends Component {
...
-        this.close()
-      }
-    } else if (!isDisabled) {
-      this.close()
+    while (el && el.parentNode !== document) {
+      scrollTop += el.scrollTop
+      el = el.parentNode
     }
+    return scrollTop
   }
```
ptbrowne: I'm not sure but I think it can be possible to have the coordinates/scroll logic only in one component. Here, it is spread between two components and I find the responsibilities of each one hard to grasp.

react/Menu/index.jsx
```diff
@@ -132,16 +162,91 @@ class Menu extends Component {
...
   onSelect: null,
-  onSelectDisabled: null
+  onSelectDisabled: null,
+  itemsStyle: {}
 }
 
 Menu.propTypes = {
-  /** Use position='right' to display the menu to the right */
+  /** Use position='right' to display the menu to the right. Use left to display it to the left.
+   * Use position='fixed' to let it choose if it has enough space to be displayed at bottom.
```
ptbrowne: I think `position` should not be used here. The thing we are trying to achieve here is an "adjustment" of the position based on viewport estate.

I think this adjustment should be done automatically without interaction from the user, no additional property is needed IMHO.

The position attribute is ortogonal to this since you first choose how you would like the popover to appear, maybe with a preference on right or left, given that it has all the space available, and then you adjust.

react/Menu/index.jsx
```diff
@@ -132,16 +162,91 @@ class Menu extends Component {
...
+  }
+
+  render() {
+    return (
+      <div
+        className={cx(styles['c-menu__inner'], styles['c-menu__inner--opened'])}
+        ref={element => {
+          if (element !== null && element !== this.element) {
+            this.element = element
+          }
```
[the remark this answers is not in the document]
Crash--: At the beginning, I wanted to do the work inside the ref callback. But DOM is not ready yet. Also, this kind of check is my react native deformation. Will remote it.

react/Menu/index.jsx
Crash--: It was done that way. I didn't want to change it to much. But will give a look.

react/Menu/index.jsx
Crash--: The menu is used every where in our code base. I don't want to break anything, specially I don't deal with scrolling / resizing issue. But you're right, I shouldn't use `position`.

I'll add a new property : `popover` {true | false}, and let `position`. I can also add a `console.warn` if popover is set to default / false ?

react/Menu/index.jsx
```diff
@@ -1,9 +1,10 @@
 import React, { Component } from 'react'
+
 import styles from './styles.styl'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { Media, Bd, Img } from '../Media'
-
+import PopOver from '../PopOver'
```
ptbrowne: Popover without capital O please, otherwise we will always make a mistake, and those on macOS will not see the mistake, and will see it in Travis.

react/Popover/Popover.jsx
```diff
@@ -0,0 +1,78 @@
+import React, { Component } from 'react'
+import ReactDOM from 'react-dom'
+import PropTypes from 'prop-types'
+export default class Popover extends Component {
+  constructor(props) {
+    super(props)
+    this.element = ''
+    this.state = {
+      style: { display: 'block', visibility: 'hidden', position: 'fixed' }
```
ptbrowne: Position fixed has some problems in Safari mobile when scrolling, is it necessary ?

react/Popover/Popover.jsx
```diff
@@ -0,0 +1,78 @@
...
+  computeCoordinates() {
+    const { refParent, style } = this.props
+    // eslint-disable-next-line
+    const menuElement = ReactDOM.findDOMNode(refParent)
+    const scrollTop = this.getScrollParent(menuElement)
+    const coordinates = menuElement.getBoundingClientRect()
+
+    //Find me & give me my coordinates
+    // eslint-disable-next-line
+    const domItem = ReactDOM.findDOMNode(this.element)
```
ptbrowne: I think there is no need for the ref `this.element`. `ReactDOM.findDOMNode(this)` would work fine IMHO.

PR #641, opened 2018-10-15
feat: ✨ Add custom properties colors

GoOz:
A lot of changes but none perceptible.

Stylus color vars point to CSS custom properties now instead of hexadecimals.
Those custom properties are now declared by default at a `:root` level.

Had to change some css values using Stylus functions such as `alpha()`, `darken()` or `rgba()`. Some of which were in fact not legit, so that's a good cleaning job as well.

I only had to add one "color", it's `overlay`, the background color for the overlay of a Modal, because it's the only one which explicitely requires transparency.

The only downside is that you can't use those colors with transparency anymore, like `rgba` or `hsla`. `opacity` is all you got but it would mean that the children would inherit the opacity as well so be careful.

I can think of a way to bypass that side effect but it would mean a lot more custom properties… :/


stylus/settings/palette.json
```diff
@@ -1,32 +1,33 @@
...
-  "melon": "#FD7461",
-  "blazeOrange": "#FC6D00",
-  "mango": "#FF962F",
-  "pumpkinOrange": "#FF7F1B",
-  "darkPeriwinkle": "#6984CE",
-  "purpley": "#7F6BEE",
-  "lightishPurple": "#B449E7",
-  "barney": "#922BC2",
-  "weirdGreen": "#40DE8E"
+  "black": "var(--black)",
```
y-lohse: Since this file is now referencing css variables, can we still use it in js?

GoOz: I guess so yes. Usually you just pick the color with JS and add it with styles, like inline-style. If so then there's no problem, but is there other usecases you can think of?

y-lohse: We're using them for [`fill`](https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L39) values too, but I think that will also work with css variables? I think that's actually [the only use case](https://github.com/search?p=1&q=org%3Acozy+settings%2Fpalette.json&type=Code) 👍 .

GoOz: `fill` attribute or CSS property ?

GoOz:
[Utiliser custom properties sur UI pour la personnalisation](https://trello.com/c/oHPjhtmr/336-utiliser-custom-properties-sur-ui-pour-la-personnalisation)

gregorylegarec:
I am maybe missing something, but if the goal is to be able to redefine and change colors from Cozy-UI, shouldn't we use more semantic names ? I thinking of a partners who'd like to change the white background with orange, all the codebase will have the `white` value were everything will be orange.

Btw how shall we release this feature ? Several options:
* Keep it as a feature branch an let every team make test on applications (spoiler, none of them will)
* Merge into master and release it as a beta, but it could block future releases
* Merge into master and release it as stable, and release patch as soon as we need in case of regressions

ptbrowne:
Please add a description.

*Automatically posted*

PR #642, opened 2018-10-15
fix(Portal): Import the main module

y-lohse:
@CPatchane I wouldn't mind a second check for this PR — for me, `require('preact-portal')` gives me the module right away, whereas `require('preact-portal').default` returns `undefined`.

(The alerts still worked but weren't portal anymore.)


CPatchane:
Thanks for this check 👍

PR #648, opened 2018-10-17
[WIP] feat: Add Tooltip component

Crash--:
Added `<Tooltip>` component based on `Popover` component.

I'm using it this way in Drive : 
```jsx 
export const Avatar = props => {
  return (
    <Tooltip title={props.textId}>
      <CozyUIAvatar {...props} className={styles['recipient-avatar']} />
    </Tooltip>
  )
}
```

### DIFF
--- a/react/Popover/Popover.jsx
+++ b/react/Popover/Popover.jsx
@@ -8,7 +8,8 @@ export default class Popover extends Component {
     this.defaultStyle = {
       display: 'block',
       visibility: 'hidden',
-      position: 'fixed'
+      position: 'fixed',
+      zIndex: 100
     }
   }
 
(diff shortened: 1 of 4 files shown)


react/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,24 @@
...
+    dataTip,
+    dataFor
+  })
+  return (
+    <React.Fragment>
+      {newChild}
+      <ReactTooltip
+        place="bottom"
+        effect="solid"
+        className={styles['shared-tooltip']}
```
Crash--: I know I've to open the API to place / effect etc.

react/Tooltip/tooltip.styl
```diff
@@ -0,0 +1,8 @@
+.shared-tooltip
```
ptbrowne: I think "shared" can be removed

react/Tooltip/index.js
```diff
@@ -0,0 +1,2 @@
+import Tooltip from './Tooltip'
+export default Tooltip
```
ptbrowne: Nit: `export default from './Tooltip'` ?

react/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import ReactTooltip from 'react-tooltip'
+import styles from './tooltip.styl'
+//Only needed for Preact?
```
ptbrowne: If it is sure, let's remove the question mark :)

ptbrowne:
Some small remarks, overall LGTM ! Thanks !

ptbrowne:
It would be good to have an example in the styleguide.

There is the styleguide.config.js to change and a Readme.md file to add.

react/Tooltip/Tooltip.jsx
ptbrowne: I did not understand ^

react/Tooltip/tooltip.styl
```diff
@@ -0,0 +1,8 @@
...
+    padding     .5rem
```
GoOz:
```suggestion
    padding     rem(8)
```

ptbrowne: Why ?

GoOz: Consistency, there's `rem()` everywhere else

Crash--:
I just pushed a new implementation of `Tooltip` based on our new `Popover` component.  No more external dependencies and no more Fragment hack. 

Added a render props to Popover to know if the component has pop bellow or not. 

@GoOz If you want to style it, feel free to do it ;)

GoOz:
What is it supposed to look like? Can we see an example in styleguidist?

Crash--:
Will reopen one day

PR #649, opened 2018-10-17
fix(Overlay): Use lib to lock scroll

y-lohse:
Using the Overlay component to open a modal would result in the page being scroll-locked, because `body-scroll-lock` would restore the `overflow` property to `hidden` when the Modal closes, as set in the Overlay.
Using the lib in both cases fixes the issue.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -9,7 +9,6 @@ import migrateProps from '../helpers/migrateProps'
 import palette from '../../stylus/settings/palette.json'
 import Portal from '../Portal'
 import uniqueId from 'lodash/uniqueId'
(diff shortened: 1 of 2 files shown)


gregorylegarec:
Right what we need also in the Store. Thanks @y-lohse !

react/Modal/index.jsx
```diff
@@ -63,13 +62,11 @@ class ModalContent extends Component {
     this.scrollingContent.addEventListener('scroll', this.handleScroll, {
```
y-lohse: @GoOz Does this line still make sense without the scroll lock? I think so, but not quite sure.

GoOz: Yes, it has another purpose.

PR #650, opened 2018-10-18
feat: Add Avatar new sizes

Crash--:


stylus/components/avatar.styl
```diff
@@ -16,6 +16,12 @@ $avatar--small
     height rem(32)
     font-size rem(16)
 
+$avatar--small-plus
+    width rem(48)
+    height rem(48)
+    font-size rem(24)
+
+
```
ptbrowne:
```suggestion
```

ptbrowne:
I don't like `small-plus`, I don't know if it mean smaller or larger.
There are already conventions for this kind of sizes

> CSS font-size has these predefined values:
xx-large
x-large
large
medium
small
x-small
xx-small

Maybe `small-plus` should be `medium`, and `medium` be `large` here ? What are the different sizes used by the design team (@Claiw) ?

LGTM for this, but I think this should be temporary.

GoOz:
@ptbrowne I agree but it would mean that this is a breaking change. Somehow it bothers me to have a breaking change for that but I must say it's the most logical way.

y-lohse:
I checked with @Claiw, and we have 5 sizes:

- 20px (not used yet but will be added)
- 32px (mostly in the contact list in the contact app)
- 40px (in a lot of places)
- 48px (recipient sharing modal in drive)
- 64px (contact card in the contact app)

which would translate to

- xsmall
- small
- medium
- large
- xlarge

Medium seems to be the most often used, so it makes sense to keep it as a default. It would be a breaking change since the previous `medium` would now be `xlarge`. AFAIK the only place where that size is used is in the contacts app.

👍 from me to do a breaking change on this.

edas:
Ok, that's the sizes we have. I am not requesting to remove some but maybe we could choose 2 or 3 in the list and consider the others deprecated. Do we really need 5 sizes which are not a multiple one of another ?

Like : 
* 32 + 64 (32 + 32x2)
* 32 + 48 + 64 (standard icon sizes)
* 20 + 40 + 60
* 16 + 32 + 64

We definitly shoud have a style guide and have strict guidelines to remove variants

Claiw:
well I suppose we will talk about it on wednesday :)

ptbrowne:
> well I suppose we will talk about it on wednesday :)

Have you talked about it ? How can we move on this issue ?

Claiw:
no, sadly Eric was not available, I look if we could talk about it this week

Claiw:
Hi,

Here is my proposal for the avatar sizes
- 16px (not used yet but will be added for sharing previsualization views)
- 32px (in the contact list in the contact app, in mobile contextual menu for a shared element)
- 40px (in a lot of places -> I explain the reason for this size below)
- 48px (recipient sharing modal in drive)
- 64px (contact card in the contact app)

concerning the exception for 40px

this size is a compromise between 32px and 48px for several reasons :
- our medium (and default) button is at height :  
40px, and it allows to have a visual line with same height for avatars and buttons next to each others. I want the same size to limit size differences between elements of the same group, and because I consider those elements have an equal importance
below, you can see the comparison between 40px and 32px : 
screen shot 2018-11-21 at 15 49 57

- list of contacts for a shared element (dropdown or contacts who have access to a shared element)
it could be interesting to take the 48px version in the future, but I think it's more interesting when we'll have some pictures for the avatars. 
40px is a good compromise compared to 32px, the 32px version "floats" and seems more like an UI bug; it's visually better to have the big avatar element with a height embracing the height of the two combined texts.
40px have a good balance, it floats less and it's not too big compared to the fact we can't personalize the avatars yet.
screen shot 2018-11-21 at 15 40 41

same opinion for the list of contacts for a shared element :
screen shot 2018-11-21 at 15 40 29

to conclude, here is a visual synthesis for all the sizes : 

**x small - 16px**
screen shot 2018-11-21 at 16 08 27

**small - 32px**
screen shot 2018-11-21 at 16 16 26

**medium - 40px**
screen shot 2018-11-21 at 15 49 57
screen shot 2018-11-21 at 15 40 29

**large - 48px**
screen shot 2018-11-21 at 16 17 27

**x large - 64px**
screen shot 2018-11-21 at 16 18 00

nono:
Ping @Crash-- @y-lohse @GoOz @ptbrowne: what should we do about this PR?

GoOz:
Well we were waiting for design specs I guess. If it's settled, I'm ok to move forward on this.

Claiw:
aren't the specs I wrote here sufficient in terms of design specs ? 
point me here what you need to complete in there are some point I need to specify

GoOz:
@Claiw yes it is. Sorry, i wasn't explicit enough, I meant now that you gave us the specs (considering that everybody's ok with it), we can move forward. 😅
That's the only reason this PR was held.

Claiw:
great ! 👍

GoOz:
[(S) New avatar sizes](https://trello.com/c/srHhcdTL/719-s-new-avatar-sizes)

GoOz:
Should be ok to review and merge now. ping @Crash--

ptbrowne:
Travis is not happy :/ https://travis-ci.org/cozy/cozy-ui/builds/474932512#L835

GoOz:
@ptbrowne Yeah, trying to figure tou why because jest is ok locally… :/

GoOz:
Well the medium size exists actually, it's just the default size for the Avatar component (with no prop size given). If you want a smaller or a bigger size then you give the props the appropriate value.

As for the Circle, it could be done in a new PR I think, but I don't really get where or when it is used, since the example pretty much show the same thing than Avatar does.

y-lohse:
>  it's just the default size for the Avatar component 

Ok !

> but I don't really get where or when it is used, since the example pretty much show the same thing than Avatar does.

Yep, visually it's the same thing, but it's used when what we're displaying is *not* a placeholder Avatar. There's one example in cozy-drive, on the "move files" screen where we write a number inside the circle.

GoOz:
Oh ok. I got it. Yeah another PR for that seems better I think. I'll do the issue.

react/Avatar/index.jsx
```diff
@@ -60,14 +64,14 @@ export const Avatar = ({ text, textId, image, size, className, style }) => {
 Avatar.propTypes = {
   text: PropTypes.string,
   image: PropTypes.string,
-  size: PropTypes.oneOf(['small', 'medium']),
+  size: PropTypes.oneOf(['xsmall', 'small', 'large', 'xlarge']),
```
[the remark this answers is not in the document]
GoOz: I don't know, the more I think about it the more I think it is useless. 
What would be the expected outcome or benefits?

stylus/cozy-ui/build.styl
```diff
@@ -609,19 +609,18 @@ NB: You'll have to deal the behaviour with javascript which is basically adding/
 Display an avatar of a user or contact with initials or an image
 
  Markup:
-<div class="c-avatar c-avatar--small">
+<div class="c-avatar {{modifier_class}}">
```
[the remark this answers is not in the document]
GoOz: Yes it is, it is KSS syntax for looping in the modifier classes list.

react/Avatar/index.jsx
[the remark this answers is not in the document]
edas: I would prefer it to be added as it costs about nothing 
It allows users to have some conditionnal value and be explicit when needed.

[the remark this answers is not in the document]
GoOz: then what does it do? nothing? it just doesn't return an error?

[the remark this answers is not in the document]
edas: the same as when you give no parameter, like any default value

GoOz:
Is it ok now?

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -8,27 +8,34 @@ exports[`ActionMenu should render examples: ActionMenu 1`] = `
...
 "<div>
-  <div class=\\"styles__c-avatar___PpDI- styles__c-avatar--medium___hjYh_\\" style=\\"color: white;\\"><span class=\\"styles__c-avatar-initials___310qC\\">CD</span></div>
+  <div class=\\"styles__c-avatar___PpDI- undefined\\" style=\\"color: white;\\"><span class=\\"styles__c-avatar-initials___310qC\\">CD</span></div>
 </div>"
 `;
 
 exports[`Avatar should render examples: Avatar 3`] = `
 "<div>
-  <div class=\\"styles__c-avatar___PpDI- styles__c-avatar--medium___hjYh_\\"><img src=\\"https://cozy.io/fr/images/cozy-logo_white.png\\" class=\\"styles__c-avatar-image___3yJnb\\" alt=\\"\\"></div>
+  <div class=\\"styles__c-avatar___PpDI- undefined\\"><img src=\\"https://cozy.io/fr/images/cozy-logo_white.png\\" class=\\"styles__c-avatar-image___3yJnb\\" alt=\\"\\"></div>
```
Crash--: do we really want `undefined` as a valid classname ?

GoOz: Isn't CSS-Modules or webpack or whatever supposed to ignore called classnames that don't exist? That's what it used to happen.

Crash--: I don't know but I'm pretty sure that I want to see `styles__c-avatar--medium___hjYh_` in my HTML instead of `undefined` or even nothing. No?

GoOz: Should be better.

gregorylegarec: It looks ok now. Shall we merge ?

PR #657, opened 2018-10-23
feat: Counter component

y-lohse:
A new component to show a simple counter. [Here's a link to the demo and the docs!](https://yannick-lohse.fr/cozy-ui/react/#counter)

There are potentially more things that could be done, like coupling the `max` and the `size` properties, an easier way to change the font color, etc. I don't really know what the needs are at the moment, so I think we can expand it later, but le me know if you think something essential is missing.

There's only one size for now — I think we'll be using the same ones as in #650 once it's settled.

### DIFF
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -32,7 +32,8 @@ module.exports = {
       components: () => [
         '../react/Layout/Layout.jsx',
         '../react/Hero/index.jsx',
-        '../react/Sidebar/index.jsx'
+        '../react/Sidebar/index.jsx',
+        '../react/Circle/index.jsx'
       ]
     },
     {
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -11,7 +11,8 @@ module.exports = {
         '../react/ButtonAction/index.jsx',
         '../react/Chip/index.jsx',
         '../react/Icon/index.jsx',
-        '../react/Spinner/index.jsx'
+        '../react/Spinner/index.jsx',
+        '../react/Counter/index.jsx'
       ]
     },
     {
--- a/react/Circle/Readme.md
+++ b/react/Circle/Readme.md
@@ -1,0 +1,11 @@
+Displays a simple circle with some text inside of it.
+
+### Default
+
+```
+<div >
+  <Circle>
+    yo
+  </Circle>
+</div>
+```
--- a/react/Circle/index.jsx
+++ b/react/Circle/index.jsx
@@ -1,0 +1,34 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import styles from './styles.styl'
+
+const Circle = ({ children, backgroundColor, size, className }) => {
+  return (
+    <div
+      className={cx(
+        styles['c-circle'],
+        {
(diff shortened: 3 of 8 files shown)


y-lohse:
[(2) Déplacer plusieurs fichiers/dossiers](https://trello.com/c/zKuIBwvp/1322-2-d%C3%A9placer-plusieurs-fichiers-dossiers)

ptbrowne:
Apart from the color, LGTM.

I think styling should be decoupled from the original component. As we are moving towards themes, I think to support themes, we should have some way of wrapping a component to provide it with `primaryColor`, `secondaryColor` etc...

```jsx
const Counter = props => ...

const counterStyle = theme => ({
  backgroundColor: primaryBackgroundColor,
  color: primaryContrastColor
})
const ThemedCounter = props => (
  <ThemeProvider>{ theme => (
     <Counter {...props} style={counterStyle(theme)} />
  }</ThemeProvider>
)
```

y-lohse:
@ptbrowne Can you look at the latest commit and see if it's what you had in mind?

I agree that tying the palette to the component is not such a good idea after all, it's convenient but might cause problems.

On the other hand I'm less comfortable with a `style` property, especially along other props like `size` and `classNames` which are other ways to influence the look of the component. 
I think component props should be explicit about what they change, so `size` is good for example, and `color` as a way to define the main color of the component also looks good to me. Blanket props like `style` are ok for extreme cases where we need to override everything, but shouldn't be the norm.

Regarding the theming, I think a better approach might be that the component maps the theme colors to it's own props? Something like

```
const Counter =({ color, size, … }) => ...

const ThemedCounter = props => (
  <ThemeProvider>{ theme => (
     <Counter {...props} color={theme.primaryBackgroundColor} />
  }</ThemeProvider>
)
```

Let me know what you think! I'm ok with the PR as it is but I still feel like a `color` prop makes more sense than a generic `style` prop.

ptbrowne:
My suggestion was mainly about not tying palette and naming the property `backgroundColor`, to echo the CSS properties. Otherwise I think we will mix it up. OK for not having styles, it will push for utility classes.

ptbrowne:
Looking twice, I don't think the circle is tied to the Counter. What if we want the `max` (99+) behavior without the circle ?

stylus/components/counter.styl
```diff
@@ -0,0 +1,20 @@
...
+    justify-content center
+    border-radius 50%
+    overflow hidden
+    background-color dodgerBlue
+    color white
+
+$counter--medium
+    width rem(40)
+    height rem(40)
+    font-size rem(20)
```
ptbrowne: It seems that for perfect alignment, a negative top margin should be applied.

image image

y-lohse:
> Looking twice, I don't think the circle is tied to the Counter. What if we want the max (99+) behavior without the circle ?

Yeah that's a good point. Do you think it would make sense to have a `Circle` component that would be used in conjunction with the `Counter` and `Avatar` components?

ptbrowne:
Yes I would be for Circle, responsible for size, and Counter inside. It's true that Avatar could be done with this but I would not do that now.

y-lohse:
Ok done! The counter and the circle are now separate components.

PR #658, opened 2018-10-23
feat(ContextHeader): Allow any illustration

y-lohse:
An evolution for the [`ContextHeader`](https://docs.cozy.io/cozy-ui/react/#contextheader) — instead of only supporting an icon for the visual clue, we can now pass it any component.

Technically this is a breaking change, since if you want an icon you now have to pass an `<Icon />` component to the new prop. However this component is very recent and AFAIK only used in one place, so using a new major version just for this seems like it would cause trouble than not. But let me know if you'd rather bump a major version anyway!

### DIFF
--- a/react/ContextHeader/Readme.md
+++ b/react/ContextHeader/Readme.md
@@ -12,7 +12,7 @@ A little inset with text and optional icon that can be used to provide some cont
 
 ```
 <div >
-  <ContextHeader title="Pick a book" text="What will you read today?" icon="album" />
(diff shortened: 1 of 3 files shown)


y-lohse:
[(2) Déplacer plusieurs fichiers/dossiers](https://trello.com/c/zKuIBwvp/1322-2-d%C3%A9placer-plusieurs-fichiers-dossiers)

y-lohse:
Replaced by #659

PR #663, opened 2018-10-29
fix: 🐛 Scrolling table when icon button inside

GoOz:
The `position: absolute` of `.u-visuallyhidden` class in a button was messing with the height of the scrollable container because the absolute had no origin inside the button.

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -54,6 +54,7 @@ themedBtn(theme)
 \*------------------------------------*/
 
 $button
+    position         relative
     box-sizing       border-box
     display          inline-flex
     margin           0 rem(4)


ptbrowne:
This means that each time we use `u-visuallyhidden` we have to make sure we are in a relative container ?

GoOz:
Usually it's not an issue. 
As this element is hidden it doesn't matter which element is its reference but in this case, probably because its a fluid height scrollable flex container, webkit doesn't deal with it properly. Firefox is ok though.
To answer your question, you shouldn't have to but in some odd cases like this one you'll have to.

GoOz:
[[Drive] Folder - infinite scroll](https://trello.com/c/RsXzT3Zi/415-drive-folder-infinite-scroll)

ptbrowne:
OK. Re-reading the original article (because I could not remember what was the use), I find that this class could be renamed `u-screenreader-only`.

GoOz:
both `sr-only` & `visuallyhidden` are commonly used accross a lot of CSS frameworks. It's kind of common knowledge. I don't really see any benefit to a breaking change like that.

ptbrowne:
:shipit:

GoOz:
Thanks for merging, I was on something else and forgot about it. 👍

Issue #664, opened 2018-10-30
Styleguidist errors


Crash--:
I was looking at the Styleguidist this morning and found several errors: 
- [ ] https://docs.cozy.io/cozy-ui/react/#layout 
`The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX.
`
- [ ] https://docs.cozy.io/cozy-ui/react/#empty 
`The `style` prop expects a mapping from style properties to values, not a string. For example, style={{marginRight: spacing + 'em'}} when using JSX.
`

- [ ] https://docs.cozy.io/cozy-ui/react/#alerter 
`TypeError: Cannot call a class as a function`

PR #665, opened 2018-10-30
Fix styleguidist

ptbrowne:
- Fix warnings in Empty/Layout
- Fix Alerter (need to pass process.env.USE_REACT to cozy-ui)
- Fix Chip which used props from render


ptbrowne:
Fix https://github.com/cozy/cozy-ui/issues/664

PR #672, opened 2018-11-06
Use portal from preact-compat

CPatchane:
Since we use `preact-compat` as alias for `react-dom` there already a `createPortal` compatibility so we don't need to use `preact-portal` anymore. And same thing for the `USE_REACT` variable.
See 
https://github.com/developit/preact-compat/pull/486


ptbrowne:
:tada

Crash--:
:+1:

PR #678, opened 2018-11-09
AppIcon: Prefer img element over fetchIcon method

gregorylegarec:
This PR ~~deprecates the fetchIcon prop for AppIcon~~ use directly icon sources
an `` tag if no `fetchIcon` prop is passed, now that all icon served by the stack are accessible
without permissions on doctype `io.cozy.apps`.

### DIFF
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -1,0 +1,40 @@
+export const preload = async (app, domain, secure) => {
+  const source = _getAppIconURL(app, domain, secure)
+
+  if (!source) {
+    throw new Error(`Cannot get icon source for app ${app.name}`)
+  }
+
+  return new Promise((resolve, reject) => {
+    const loader = document.createElement('img')
+    loader.onload = () => {
+      resolve(source)
+    }
+    loader.onerror = () => {
+      const loadError = new Error(`Error while preloading ${source}`)
+      console.error(loadError.message)
+      reject(loadError)
+    }
+    loader.src = source
+  })
+}
+
+const _getProtocol = (secure = true) => {
+  return `http${secure ? 's' : ''}:`
+}
+
+const _getAppIconURL = (app, domain, secure) => {
+  if (!domain) throw new Error('Cannot fetch icon: missing domain')
+
+  return (
+    app.links &&
+    app.links.icon &&
+    `${_getProtocol(secure)}//${domain}${app.links.icon}`
+  )
+}
+
+export const Preloader = {
+  preload
+}
+
+export default Preloader
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -1,3 +1,5 @@
+const _loaded = {}
+
 export const preload = async (app, domain, secure) => {
   const source = _getAppIconURL(app, domain, secure)
 
@@ -8,6 +10,7 @@ export const preload = async (app, domain, secure) => {
   return new Promise((resolve, reject) => {
     const loader = document.createElement('img')
     loader.onload = () => {
+      _loaded[source] = true
       resolve(source)
     }
     loader.onerror = () => {
@@ -17,6 +20,11 @@ export const preload = async (app, domain, secure) => {
     }
     loader.src = source
   })
+}
+
+export const getPreloaded = (app, domain, secure) => {
+  const source = _getAppIconURL(app, domain, secure)
+  return source && _loaded[source] ? source : null
 }
 
 const _getProtocol = (secure = true) => {
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -34,10 +34,20 @@ const _getProtocol = (secure = true) => {
 const _getAppIconURL = (app, domain, secure) => {
   if (!domain) throw new Error('Cannot fetch icon: missing domain')
 
+  let path = _getInstalledIconPath(app)
+  path = path || _getRegistryIconPath(app)
+  return path ? `${_getProtocol(secure)}//${domain}${path}` : null
+}
+
+const _getInstalledIconPath = app => {
+  return app.links && app.links.icon
+}
+
+const _getRegistryIconPath = app => {
   return (
-    app.links &&
-    app.links.icon &&
-    `${_getProtocol(secure)}//${domain}${app.links.icon}`
+    app.latest_version &&
+    app.latest_version.version &&
+    `/registry/${app.slug}/${app.latest_version.version}/icon`
   )
 }
 
--- a/react/AppIcon/Readme.md
+++ b/react/AppIcon/Readme.md
@@ -1,6 +1,22 @@
-`<AppIcon />` uses a `fetchIcon` prop to load asynchronously an app icon.
+`<AppIcon />` renders an app icon.
 
-## Example with a fake instance of cozyClient
+It contains business logic related to apps documents.
+
+Apps documents provided by the stack are containing a `links.icon` property which is public and can be directly used as `src` value for an `<img />` tag.
+
+### Example
+```
+  <div>
+    <p>
+      <AppIcon app={app} />
+    </p>
+  </div>
+```
+
+## Using fetchIcon
+Is it also possible to provide a custom asynchronous `fetchIcon` which takes an app document as argument and resolves with an icon URL (see [`URL.createObjectURL`](https://developer.mozilla.org/en/docs/Web/API/URL/createObjectURL)). `fetchIcon` may also throw errors.
+
+### Example with fetchIcon
 ```
   <div>
     <p>
@@ -9,5 +25,5 @@
   </div>
 ```
 
-The `fetchIcon` method is given `app.links.icon` as argument and must return an URL. This URL may be created with [`URL.createObjectURL`](https://developer.mozilla.org/en/docs/Web/API/URL/createObjectURL).
-`fetchIcon` may also throw errors.
+## `domain` and `secure` props
+If the `fetchIcon` is missing, the `<AppIcon />` component needs the `domain` prop (litteraly the cozy domain) and the secure prop, to know which protocol use between `http` or `https`.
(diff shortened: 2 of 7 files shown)


react/AppIcon/test/AppIcon.spec.js
```diff
@@ -7,39 +7,200 @@ import { shallow } from 'enzyme'
...
+        app={app}
+        fetchIcon={async () => {
+          const icon = await successFetchIcon()
+          return icon
+        }}
+      />
     )
-    await successFetchPromise
+
+    await failureImageLoad
```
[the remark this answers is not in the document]
gregorylegarec: See the comment before the `beforeEach` method.

It's pretty impossible to detect or wait for the end of `componentDidMount` into the test. The only way I found after reading stuff online is to wait one promise per call in the tested component.

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
+
+const _getRegistryIconPath = app => {
+  return (
+    app.latest_version &&
+    app.latest_version.version &&
+    `/registry/${app.slug}/${app.latest_version.version}/icon`
+  )
+}
+
+const _loaded = []
```
ptbrowne: I think `_loaded` should be managed by the `_preload` function. This way, AppIcon does not have to do it. A cache abstraction could maybe help to divide responsibilities more cleanly.

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
+    throw new Error('Cannot find [role=application] element')
+  }
+  domain = root.dataset.cozyDomain
+  if (!domain) {
+    throw new Error(
+      'Application element does not contain `data-cozy-domain` attribute'
+    )
+  }
+  return domain
+}
```
ptbrowne: After a mobile logout, this variable will not be cleared and thus domain will have the wrong value after a second login.

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
   constructor(props, context) {
     super(props, context)
+
+    const { app } = props
+    let source
+    try {
+      source = _getAppIconURL(app)
+    } catch (error) {} // eslint-disable-line no-empty
+
+    const loaded = source && _loaded.includes(source)
```
ptbrowne: Hashes are a better fit to check for inclusions (no need to go through the whole array).

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
+        await this.fetchIcon()
+      } else {
+        console.error(error.message)
+        this.setState({ status: ERRORED, error: error })
+      }
+    }
+  }
+
+  async load() {
+    this.setState({ status: FETCHING })
```
ptbrowne: The FETCHING/DONE state is managed at two places in this class, it would be clearer it was managed only at one place.

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
+    const { app } = this.props
+
+    let source
+    try {
+      source = _getAppIconURL(app)
+    } catch (error) {} // eslint-disable-line no-empty
+
+    if (!source) {
+      throw new Error(`Cannot get icon source for app ${app.name}`)
+    }
```
ptbrowne: Why catch and do nothing and then put this if ?

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,129 @@ import Icon from '../Icon'
...
+
+    let loadedUrl
+    try {
+      loadedUrl = await _preload(source)
+    } catch (error) {
+      console.error(error.message)
+      throw error
+    }
```
ptbrowne: Could this console.error be done in the _preload method ? The code would be cleaner in the component.

react/AppIcon/index.jsx
gregorylegarec: The only flaw I see is that the AppIcon, in the constructor, needs to know if a icon has already been loaded or not, to renders it immediately without a temporary `fetching` state.

react/AppIcon/test/AppIcon.spec.js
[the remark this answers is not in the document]
gregorylegarec: Fixed with the use of the `onReady`

react/AppIcon/index.jsx
gregorylegarec: The only solution I see is to always perform the `document.querySelector`. Do you see any other option ?

ptbrowne: This won't work in mobile as we do not have the cozyDomain on the DOM node. To me, AppIcon should work in the context of a client. And the domain should be taken from the client.

react/AppIcon/index.jsx
```diff
@@ -7,16 +7,134 @@ import Icon from '../Icon'
...
+    )
+  }
+  return domain
+}
+
+let protocol
+const _getProtocol = () => {
+  if (protocol) return protocol
+  protocol = window.location.protocol
+  return protocol
```
ptbrowne: This will not work on mobile as the protocol is `file:///`

react/AppIcon/index.jsx
gregorylegarec: Actually, we still can use in the constructor.

react/AppIcon/index.jsx
gregorylegarec: I prefer hashes too, I don't even remember why I used `.includes`, exhaustion maybe. 👍

react/AppIcon/index.jsx
gregorylegarec: The thing is, it's managed at the beginning when the component starts to load the icon, and updated again when it's done. The thing is ti should be done directly in the `componentDidMount` method I think.

gregorylegarec:
Thanks for your feedbacks @ptbrowne and @CPatchane, it should be ok now (at least I like the way it looks). Tests are cleaner (they pointed out errors in code) and concerns are separated.

Tell me if you see other things to improve, thanks !

react/AppIcon/test/AppIcon.spec.js
```diff
@@ -7,39 +7,128 @@ import { shallow } from 'enzyme'
...
+          const wrapper = shallow(
+            <AppIcon app={app} domain={domain} secure={secure} />
+          )
+          const component = wrapper.getElement()
+          expect(component).toMatchSnapshot()
+          expect(Preloader.getPreloaded).toHaveBeenCalledWith(
+            app,
+            domain,
+            secure
+          )
```
ptbrowne: I am not sure if this should be tested as the preloader is an implementation detail.

react/AppIcon/Preloader.js
```diff
@@ -0,0 +1,58 @@
+const _loaded = {}
```
ptbrowne: I am not a fan of the Preloader name, for me it is just a cache.

`Preloader.getPreloaded` -> `cache.get`
`Preloader.preload` -> `cache.fetch`

react/AppIcon/Readme.md
```diff
@@ -1,6 +1,24 @@
...
+```
+  <div>
+    <p>
+      <AppIcon app={app} />
+    </p>
+  </div>
+```
+
+## Using fetchIcon
+Is it also possible to provide a custom asynchronous `fetchIcon` which takes an app document as argument and resolves with an icon URL (see [`URL.createObjectURL`](https://developer.mozilla.org/en/docs/Web/API/URL/createObjectURL)). `fetchIcon` may also throw errors.
```
ptbrowne: I would not document this if it is deprecated.

gregorylegarec: Actually I not sure anymore we should deprecate it.

react/AppIcon/Preloader.js
gregorylegarec: Actually the pattern is an image preloading.

react/AppIcon/Readme.md
```diff
@@ -9,5 +27,5 @@
   </div>
 ```
 
-The `fetchIcon` method is given `app.links.icon` as argument and must return an URL. This URL may be created with [`URL.createObjectURL`](https://developer.mozilla.org/en/docs/Web/API/URL/createObjectURL).
-`fetchIcon` may also throw errors.
+## `domain` and `secure` props
+If the `fetchIcon` is missing, the `<AppIcon />` component needs the `domain` prop (litteraly the cozy domain) and the secure prop, to know which protocol use between `http` or `https`.
```
ptbrowne: Since the AppIcon are going to be rendered in the context of a client, those props could be provided automatically.

```
const AppIcon = withClient(props => {
  return <DumbAppIcon domain={props.client.domain} secure={props.client.url.indexOf('https') === 0} />
})
```

or something like this ?

ptbrowne:
LGTM. Thanks @gregorylegarec !

react/AppIcon/Readme.md
gregorylegarec: Yep, it looks clever, I like it. (By the way I am still not a fan of always testing the protocol in client.url. IMHO the client should expose a `secure` attribute, but it's a subject for another time :))

react/AppIcon/test/AppIcon.spec.js
gregorylegarec: We target to use the image preloading as default behaviour, now that icons are accessible without permissions, so I think it is important to test it.

react/AppIcon/index.jsx
```diff
@@ -6,43 +6,55 @@ import styles from './styles.styl'
...
+    } catch (error) {
+      loadError = error
     }
 
-    this.setState({ status: 'fetching' })
+    this.setState({
+      error: loadError,
+      icon: loadedUrl,
+      status: loadError ? ERRORED : DONE
+    })
```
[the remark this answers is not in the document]
gregorylegarec: Let's do this

react/AppIcon/Preloader.js
ptbrowne: Yes I knew the name :) I still prefer the names cache / get and fetch, I find `getPreloaded` a bit ugly :)

PR #679, opened 2018-11-09
feat: Upgrade to react-select 2.1.1

Crash--:
ATM we're on a beta... So I upgraded to the latest stable. 

I didn't have any issue but regarding the changelog we can have one issue but I don't know if we are concerned
 
`[BREAKING] innerRef assignments in custom components must now be made from the root of the prop object, as opposed to reaching into innerProps. This is part of the work to normalize the behaviour circa custom components. #2824`

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "preact": "^8.3.1",
     "pretty": "^2.0.0",
     "react-hot-loader": "^4.3.11",
-    "react-select": "^2.0.0-beta.6",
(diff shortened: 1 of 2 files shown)


GoOz:
ping @y-lohse I think he's the one with most background on this component.

y-lohse:
I don't think we're using `innerRef` at the moment, so we're not affected. That said there might be places where we should use it and we don't.

PR #682, opened 2018-11-12
fix: 🐛 Changed Modal CSS order to avoid conflicts

GoOz:
Button styles were loaded after the Modal styles, making some unreasonable conflicts.


ptbrowne:
LGTM. What were the conflicts ?

GoOz:
Button classes like `.c-btn` and others were declared **after** the `.c-modal-close` class from the Modal when this class is supposed to overwrite some properties from `.c-btn` class. 
It's hard to overwrite anything when you come first in CSS.

PR #689, opened 2018-11-19
fix: 🐛 IntentModal can have has-modal class

GoOz:
`has-modal` class was only added when `ModalContent` was used and
IntentModal don't use this component so I moved the `has-modal` toggle
to `Modal` component instead.


CPatchane:
Thanks @GoOz 👍

PR #694, opened 2018-11-19
fix(Select): Added innerRef prop

y-lohse:
Regarding #679 , [react-select has an `innerRef` prop that needs to be passed down for some components](https://react-select.com/components#inner-ref).

### DIFF
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -14,6 +14,16 @@ const options = [
 <SelectBox options={options} />
 ```
 
+For example if you want to force the menu to be open:
+
+```
+const options = [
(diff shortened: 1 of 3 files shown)


ptbrowne:
> Should the snapshot test fail because of this change?

The SelectBox is not opened during the snapshot. This is why the snapshot does not fail.

https://github.com/cozy/cozy-ui/blob/master/react/__snapshots__/examples.spec.jsx.snap#L920

I think opening the SelectBox automatically during tests would be a good ida.

Crash--:
> I think opening the SelectBox automatically during tests would be a good idea.

+1111111

Crash--:
<3

PR #700, opened 2018-11-20
feat(Select): Extra render prop in Option

y-lohse:
I needed to add some buttons *inside* an `Option`. I doubt anyone will need the same buttons, so I added an option that allows us to render extra elements inside the Option. Please [see the new documentation for more details](https://yannick-lohse.fr/cozy-ui/react/#selectbox) ("Further control over options" section).

### DIFF
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -37,7 +37,7 @@ const options = [
   { value: 'vanilla', label: 'Vanilla' }
 ];
 
-<SelectBox 
+<SelectBox
   options={options}
   components={{
     Control: reactSelectControl(MyControl)
@@ -57,13 +57,35 @@ const options = [
   { value: 'vanilla', label: 'Vanilla' }
 ];
 
-<SelectBox 
+<SelectBox
   options={options}
   isMulti
   components={{
     Option: CheckboxOption
   }}
 />
+```
+
+### Options with actions
+
+You can display additional actions inside an Option with the ActionsOption component.
+
+```
+const ActionsOption = require('../SelectBox').ActionsOption;
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' }
+];
+
+const CustomOption = (props) => (<ActionsOption {...props} actions={[
+    { icon: 'delete', handler: () => alert('deleting') },
+    { icon: 'rename', handler: ({ data }) => alert(data.value) }
+  ]} />);
+
+<SelectBox options={options} components={{
+  Option: CustomOption
+}} />
 ```
 
 ### Fixed options
@@ -75,7 +97,7 @@ const options = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('').map(x => ({ value:x, labe
 options[0].fixed = true
 options[options.length - 1].fixed = true;
 
-<SelectBoxWithFixedOptions 
+<SelectBoxWithFixedOptions
   menuIsOpen={isTesting ? true : undefined}
   options={options} />
 ```
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -79,8 +79,8 @@ const options = [
 ];
 
 const CustomOption = (props) => (<ActionsOption {...props} actions={[
-    { icon: 'delete', handler: () => alert('deleting') },
-    { icon: 'rename', handler: ({ data }) => alert(data.value) }
+    { icon: 'delete', onClick: () => alert('deleting') },
+    { icon: 'rename', onClick: ({ data }) => alert(data.value) }
   ]} />);
 
 <SelectBox options={options} components={{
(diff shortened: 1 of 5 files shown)


y-lohse:
[(3) Groupes - supprimer](https://trello.com/c/AUW3POQY/1404-3-groupes-supprimer)

Crash--:
I agree with you that we will not have the same buttons in our apps. But having a common wrapper for theses extra (with the right margin-right, flex alignment...) can be a good idea?

y-lohse:
I'm going to add a pre-made wrapper around `renderExtra`, specifically to render a bunch of button-icon with proper styling. Hang tight.

y-lohse:
Ok, I rewrote this PR. The original `Option` component is unchanged, and instead we have a new `ActionsComponent` that can be used to inject… well, actions into an option. Documentation and example is still here : https://yannick-lohse.fr/cozy-ui/react/#selectbox

react/SelectBox/Readme.md
```diff
@@ -66,6 +66,28 @@ const options = [
...
+const ActionsOption = require('../SelectBox').ActionsOption;
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' }
+];
+
+const CustomOption = (props) => (<ActionsOption {...props} actions={{
+  delete: () => alert('deleting'),
+  rename: ({ data }) => alert(data.value)
```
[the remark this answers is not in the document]
y-lohse: Yeah i wondered about that. The `SelectionBar` component also has silent icons so I wanted to keep it consistent, but I think i also like the more verbose option better.

react/SelectBox/SelectBox.jsx
```diff
@@ -109,6 +113,34 @@ const CheckboxOption = ({ ...props }) => <Option {...props} withCheckbox />
...
+            action.handler(props)
+          }}
+        />
+      ))}
+    </span>
+  </Option>
+)
+
+ActionsOption.propTypes = {
+  actions: PropTypes.objectOf(PropTypes.func)
```
ptbrowne: The propTypes is no longer OK.

PropTypes.shape({
  icon: PropTypes.string,
  handler: PropTypes.func
})

react/SelectBox/SelectBox.jsx
```diff
@@ -109,6 +113,34 @@ const CheckboxOption = ({ ...props }) => <Option {...props} withCheckbox />
...
+}
+
+ActionsOption.defaultProps = {
+  actions: {}
```
ptbrowne:
```suggestion
  actions: []
```

ptbrowne:
LGTM after the small remarks. Thanks @y-lohse !

ptbrowne:
One thing that bothers me a bit is the style. When nothing is selected, the action buttons float in the middle of nowhere.

image

I have trouble also understanding the specific usecase here. If I've understood well, these would be used in the context of a SelectBox to add/remove groups to a contact. In this case, why are there action buttons on the group ? As a user, I don't think my primary intention when opening this select box is to rename/delete a group. It clutter the UI a bit. @Claiw @joel-costa ?

react/SelectBox/Readme.md
```diff
@@ -66,6 +66,28 @@ const options = [
...
+```
+const ActionsOption = require('../SelectBox').ActionsOption;
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' }
+];
+
+const CustomOption = (props) => (<ActionsOption {...props} actions={[
+    { icon: 'delete', handler: () => alert('deleting') },
```
Crash--: `handler`? Since this is designed to be executed during the `onClick`, why not naming it `onClick` ?

y-lohse: That would work too — but it's also the handler for the action, or the event. I don't know, I don't think one is much better than the other ?

react/SelectBox/SelectBox.jsx
```diff
@@ -109,6 +113,39 @@ const CheckboxOption = ({ ...props }) => <Option {...props} withCheckbox />
...
+      icon: PropTypes.string,
+      handler: PropTypes.func
+    })
+  )
+}
+
+ActionsOption.defaultProps = {
+  actions: []
+}
+
```
Crash--: New file? :-°

Claiw:
hi, the screenshot I see doesn't fit with the specifications.
Here is the expected behaviour you can find in our zeplin spec artboard :
screen shot 2018-11-23 at 18 12 03
on the hover of an item the actions appears, there are not supposed to be displayed by default.

ptbrowne:
I think the actions are displayed by default for the menu to work on mobile without any extra work. I think they could be showed by default on mobile and hidden on desktop, while we wait to display a proper action menu on mobile. @y-lohse ?

y-lohse:
Yep, could work. Lemme try that.

Claiw:
💯

ptbrowne:
I created an issue [here](https://github.com/cozy/cozy-ui/issues/719) since this PR is closed :)

PR #702, opened 2018-11-21
AppIcon: Allow undefined domain

gregorylegarec:
AppIcon was not usable without a domain prop,
making fetchIcon prop impossible to use.


CPatchane:
Thanks @gregorylegarec 🌮

PR #704, opened 2018-11-21
fix(Icon): Handle palette in <Icon />

CPatchane:
Be able to use the cozy-ui palette in this component like:

```jsx
<Icon icon="help" color="dodgerBlue" />
```

### DIFF
--- a/react/Icon/Readme.md
+++ b/react/Icon/Readme.md
@@ -10,7 +10,7 @@ your applicaiton.
 
 ```
 const Sprite = require('./Sprite').default;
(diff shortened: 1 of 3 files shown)


ptbrowne:
Have you considered how would this work with themes ?

CPatchane:
Thanks @nono 
For the theme I don't really know how our current palette mecanism will work with the theme so if a component use the palette it will be the same work/question for me.

ptbrowne:
I am not sure this PR is necessary since one can `import` the palette in the app and use the color from the palette. I do not see the gain here.

CPatchane:
Ok then let's close this PR and see if this need will come back. Thanks @ptbrowne 👍

PR #706, opened 2018-11-21
chore: Remove generated palette

ptbrowne:
The JS palette file is generated when releasing (prepublishOnly),
it is not useful in the repo. I added a docs/dev.md to indicate
how to develop with a linked cozy-ui.


CPatchane:
What about doing it on `postinstall`?

ptbrowne:
> What about doing it on postinstall?

What's the advantage ?

CPatchane:
> > What about doing it on postinstall?
> 
> What's the advantage ?

Not having to remember doing it each time, instead it will be done after installing cozy-ui dependencies when we want to work on it.

ptbrowne:
Thanks nono, it took some time because of a hard to understand problem with Travis. It was reporting "obsolete snapshots" when in local, this was not reported. Eventually, adding `--ci` locally showed the problem in local; it was a problem of transpilation since specs were transpiled. Removing specs from transpilation solved the problem.

PR #707, opened 2018-11-21
docs: Set focus programmatically

Crash--:
Added an exemple to show how to focus programmatically an `<Input />`

### DIFF
--- a/react/Input/Readme.md
+++ b/react/Input/Readme.md
@@ -34,6 +34,33 @@
 <Input placeholder="I'm full width" fullwidth />
 ```
 
+### inputRef ( set focus / blur / ... programmatically)
+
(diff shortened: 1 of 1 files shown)


react/Input/Readme.md
```diff
@@ -34,6 +34,30 @@
 <Input placeholder="I'm full width" fullwidth />
 ```
 
+### inputRef ( set focus / blur / ... programmatically)
+
+```
```
ptbrowne:
```suggestion
```If you need to programmatically access the underlying `<input />` for example to give focus or move the caret, you can use the `innerRef` prop, that is passed to the `ref` property of the `<input />`.
```

Crash--: is it better now ?

PR #708, opened 2018-11-21
feat: ✨ Add platform wizard styles

GoOz:
Took the styles for App login from Drive to UI so it can be used by the Cloudery and the Stack in the near future.


GoOz:
[(𝐒) Amélioration de la page d'accueil - Drive (mobile)](https://trello.com/c/4YTthk9q/360-%F0%9D%90%92-am%C3%A9lioration-de-la-page-daccueil-drive-mobile)

GoOz:
[(𝐌) [LOGIN] Amélioration de la page de saisi d'url (mobile)](https://trello.com/c/lrJOqAeY/362-%F0%9D%90%8C-login-am%C3%A9lioration-de-la-page-de-saisi-durl-mobile)

stylus/components/wizard.styl
```diff
@@ -0,0 +1,168 @@
...
+    left 0
+    height 100%
+    display flex
+    justify-content center
+    align-items center
+    width 100%
+    background-color white
+    color charcoalGrey
+    text-align center
+    transition height ease 300ms
```
ptbrowne: This is surprising to have a transition on height since it will trigger a reflow/repaint each time. I think using transition only on properties that are managed by the composer is a better practice.

GoOz: This is needed for iOS (when we'll resize the wrapper on keyboard open). 
What do you call "properties that are managed by the composer" exactly?

stylus/components/wizard.styl
```diff
@@ -0,0 +1,168 @@
...
+
+$wizard-notice--lost
+    font-size rem(16)
+
+    a
+        color dodgerBlue
+        text-decoration none
+
+        &:hover
+            color scienceBlue
```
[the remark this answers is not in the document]
GoOz: Clearly this is something to work on but kinda redundant with future addition of utilities. Until it's not added and so until we know how it will be done I think this got the job done.

stylus/components/wizard.styl
```diff
@@ -0,0 +1,168 @@
...
+$wizard-logo-img
+    width 100%
+
+$wizard-logo-badge
+    position absolute
+    bottom rem(6)
+    right rem(16)
+    width rem(32)
+    height rem(32)
+    background dodgerBlue embedurl('./assets/icons/icon-cozy.svg') center no-repeat
```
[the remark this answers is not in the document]
GoOz: I honestly didn't refactor this part, I guess it could be done indeed.

stylus/components/wizard.styl
ptbrowne: I meant css properties whose changes do not affect layout or repainting and can be completely carried oud by the compositor thread on the GPU.

> Animations made with JavaScript are sometimes called imperative animations, and those made with CSS are called declarative animations. CSS animations are handled by the browser’s compositor thread rather than the main thread responsible for painting and styling. Consequently, such animations are unaffected by the main thread’s more expensive tasks. *Of course, animations that trigger the paint or layout events will require work from the main thread, which negates the benefits of using CSS animations.*

Emphasis mine.

The height property will cause layout changes, and these mean animations that are not very fluid.

https://www.keycdn.com/blog/animation-performance
https://csstriggers.com/transform

GoOz: Ok but still, I want to animate the height so I don't see what else I can do.

ptbrowne: Sometimes it is not necessary to animate the height, the opacity and/or scaleX/scaleY can be used. Maybe it is not necessary here.

Crash--:
Lot of CSS, but LGTM for now ;)

PR #711, opened 2018-11-22
fix(dep): Use react-dom as dev dep

CPatchane:
This was breaking the react-select installation.
It was missing from installing with yarn.


Crash--:
Is it because of the upgrade ? or because of the move from `dep` to `devDep` ?

CPatchane:
@Crash-- I just tried it again and it's clearly the move from `dep` to `devDep` the fix here.

PR #712, opened 2018-11-22
Make react-dom a peer-dependency

CPatchane:
We only need it when a React application wants to use React components.


Crash--:
Nop. I know that I use react dom in the Popover component :(

CPatchane:
Ok then we can close this PR 👍

Issue #705, opened 2018-11-21
Remove delete-grey08, upload-grey08, share-08 icons


CPatchane:
These icons are available to be used by the component `<Icon />`, I would suggest to remove them from the `icons/ui` folder since those are icons with colors. At least, we can remove them from the `<IconSprite />` to not use them through `<Icon />`.

https://github.com/cozy/cozy-ui/commits/master/assets/icons/ui/delete-grey08.svg
https://github.com/cozy/cozy-ui/blob/master/assets/icons/ui/share-grey08.svg 
https://github.com/cozy/cozy-ui/blob/master/assets/icons/ui/upload-grey08.svg

ptbrowne:
Go :)

kosssi:
I propose that we list the places where we use its icons to delete them if it is no longer used?

- Cozy Banks don't use it

y-lohse:
They have been removed.

Issue #674, opened 2018-11-07
The place of the IconSprite


CPatchane:
With the last major cozy-ui version, the `<IconSprite />` must be included in the application in order to have it in the DOM and to use it.
The problem is that the `cozy-bar` also uses `cozy-ui` so needs that icon sprite but the `cozy-bar` must not rely on the application since it must be independent (must work for an application which doesn't use cozy-ui).
So I see many solutions to this issue (temporary or not):
- My preference will be to let the bar handle the cozy-ui sprite inclusion into the DOM tree. On the one hand the `cozy-bar` will have to be with the last version of cozy-ui to have the last sprite but on the other hand the sprite doesn't change a lot to be often blocking
- The bar have to include its own namespaced sprite, but each time we will have to use a new icon (or a component from cozy-ui needs a new icon) we will have to add it into this namespaced sprite manually (or add all cozy-ui icons in one shot)
- Let the stack handle the icon sprite as a placeholder `{{.UISprite}}` but it seems heavy and add more dependency to the stack
- Add also the sprite in the bar, make things twice but the priority in the app code will be the first sprite found which will be the bar one since it's before the application in the DOM tree.

CPatchane:
For now, the bar will require the needed icons by itself in the sprite to be independant, see https://github.com/cozy/cozy-bar/pull/332/commits/d58bd1e5432b738bc5393936f8c354ee0fab7d0e

ptbrowne:
I think it is possible to import the icons directly from cozy-ui : 

`import topIcon from 'cozy-ui/assets/icons/ui/top.svg'`

y-lohse:
I actually think the current approach is the correct one.

I'd rather not let the apps rely on the bar to get their icons, it's not intuitive and would require bar updates to get new icons, which isn't always possible.

Is there a major problem with how things are done at the moment? I'm guessing you're not importing new icons very often in the bar?

y-lohse:
Closing this since everything seems ok.

PR #715, opened 2018-11-24
Proposal: Use SVGR components for UI icons

CPatchane:
This PR brings SVG components using [svgr](https://github.com/smooth-code/svgr) to the `<Icon />` instead of using the sprite `<IconSprite/>`. 

Results/Preview available here: https://cpatchane.github.io/cozy-ui/react/#icon

### Why ?
For me, the application should not worry about adding a sprite to use a component from a UI framework. Using this way, the application just have to use `<Icon />` from cozy-ui and the only used icons will be added to the app code.

### Breaking changes ?

I don't think so. The `<Icon />` will use these components by default when you will pass a string which do not start by `#` as the `icon` props. If it find a component matching this name, it will use the related SVG component. If not it will use the id as before (#name).
If we really want to force using the sprite from the application, a props `forceSprite` can be passed to force using sprite in all cases.

### Anything else ?

This PR also remove all these icons with colour from `<Icon />` usage (not from cozy-ui): `delete-grey08`, `share-grey08`, `upload-grey08`, `dots-white`, `cross-white`, `spinner-red`, `spinner-blue`, `spinner-grey`, `spinner-white`.
They shouldn't be usedby `<Icon />` since there is a colour option and you can't use this option with those.

Fix https://github.com/cozy/cozy-ui/issues/705 
Fix https://github.com/cozy/cozy-ui/issues/674


package.json
```diff
@@ -34,8 +35,9 @@
     "test:jest": "env BABEL_ENV=transpilation jest --verbose --coverage",
     "transpile": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --copy-files --ignore '**/test/**,**/__mocks__/**'",
     "sprite": "scripts/make-icon-sprite.sh transpiled/react/Icon/icons-sprite.js; cp transpiled/react/Icon/icons-sprite.js react/Icon/icons-sprite.js",
+    "svgComp": "svgr --icon --no-dimensions --prettier-config .prettierrc.json -d 'react/_icons' 'assets/icons/ui'",
     "copy-files": "scripts/copy-files.sh",
-    "release": "yarn transpile; yarn sprite; yarn copy-files",
+    "release": "npm-run-all --sequential transpile sprite copy-files",
```
ptbrowne: What is the advantage of doing this ?

ptbrowne: npm-run-all exposes the run-s shortcut.

react/Icon/index.jsx
```diff
@@ -2,6 +2,8 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
 import cx from 'classnames'
+import toPascalCase from 'to-pascal-case'
+import * as iconsIndex from './icons'
```
ptbrowne: >  and the only used icons will be added to the app code.

With this, all icons are added.

ptbrowne:
LGTM, thanks Cedric !

Crash--:
Thank you @CPatchane for this work! I like the idea of not having to include any sprite in our app. 

To resume: 

- We remove the need of using a sprite by using an IconComponent. But if we use 10 times the same icon, we'll have ten times the icon into the DOM. 
- Having an icon component means that we can't override it anymore. Is it something we really want for UI's icons @edas ? 

Some questions / remarks :
- It seems that it will be retro compatible. So we'll be able to use Icon sprite and icon component at the same time? 
- Can we edit the doc to explain how to add a new icon to cozy ui with this new system?

CPatchane:
> It seems that it will be retro compatible. So we'll be able to use Icon sprite and icon component at the same time?

Yes, you still can import a sprite in your application and use the it via the `forceSprite` props here

edas:
>     * Having an icon component means that we can't override it anymore. Is it something we really want for UI's icons @edas ?

No we don't
What we are trying to achieve, is a way to be able to override all icons and colors. We probably should put in a sprite or external file all SVG which are now in base64, not the opposite.

Icons that respect a theme (black & white themable by CSS) may be different. The question is : do we really want them to be inserted multiple times in the DOM if we use them more than once ?

CPatchane:
@edas So this PR is not a solution that we want about SVG handling here. It may be better to see what we can do with our sprite tools instead. Thanks!

PR #716, opened 2018-11-26
fix(Select): Transparent action buttons on desktop

y-lohse:
As suggested in #700 , this PR makes the action buttons transparent on larger screens — this the demo [here](https://yannick-lohse.fr/cozy-ui/react/#selectbox).


react/SelectBox/styles.styl
```diff
@@ -44,6 +48,11 @@
 
 .select-option__actions
     float right
+    opacity 0
+    transition all .2s ease-out
```
ptbrowne:
```suggestion
    transition opacity .2s ease-out
```

PR #721, opened 2018-11-28
feat(Select): Use ellipsis for text overflow

y-lohse:
This PR turns long labels in the select box into ellipsis-truncated texts. You can see the demo here : https://yannick-lohse.fr/cozy-ui/react/#selectbox


react/SelectBox/SelectBox.jsx
```diff
@@ -86,7 +87,10 @@ const Option = ({
         className={styles['select-option__checkbox']}
       />
     )}
-    <span className={styles['select-option__label']}>{children}</span>
+    <span className={styles['select-option__label']}>
+      <span className="u-ellipsis">{customValue ? customValue : children}</span>
+      {customValue ? children : false}
+    </span>
```
y-lohse: A note on this bit as it's maybe not intuitive.

For regular situations, we just put the option label in the ellipsis container, no big deal.

But when appending actions to an option with the `ActionsOption` component, the normal behaviour would mean the buttons end up in the ellipsis container, which we don't want. So in that case we can explicitly provide the option's label, and the rest of the option is rendered outside the ellipsis container.

y-lohse:
[(2) Groupes - longueur des noms (dropdowns)](https://trello.com/c/8lFuNJAL/1431-2-groupes-longueur-des-noms-dropdowns)

y-lohse:
[We're using it](https://github.com/cozy/cozy-ui/pull/721/files#diff-04e31d606b9893f78f10f735f8f8b8dbR91) ;)

Crash--:
It's somehow a breaking change since we're changing the current behavior, right? 

I don't know if we have somewhere else long label for selectBox, but maybe we can add a props to tell if we want to break / no-break? 

@Claiw This is something we always want? Or can we have some situation with multiples lines?

y-lohse:
My understanding is that's what we always want, so this is more a fix than a breaking change (and I should change the commit type accordingly). It's not *breaking* but it will change how things look, so I don't mind labeling it as breaking change if it helps :)

react/SelectBox/SelectBox.jsx
ptbrowne: `customValue` dans ce cas représente le label c'est bien ça ? Dans ce cas là on peut renommer customValue en label ?

y-lohse: C'est bien ca, mais la prop `label` est déja prise — c'est littéralement le `label` sous forme de `string`, alors que `customValue` est le label sous forme de composant (c'est un `ValueContainer ` de react-select ou un truc dans le genre.)

`value` est déja pris aussi — mais on peut mettre `customLabel` si c'est plus clair?

ptbrowne: > alors que customValue est le label sous forme de composant

labelComponent ? labelChildren ?

y-lohse: I like `labelComponent` ! Let's go for that 👍

PR #722, opened 2018-11-28
fix: SelectBoxWithFixedOptions works with React/Preact

ptbrowne:
Use Group features of SelectBox to implement the fixed behavior. Much less hacky and works with React and Preact.

Visible at: https://ptbrowne.github.io/cozy-ui/react/#!/SelectBox/11


Crash--:
👍 tip top

ptbrowne:
No problem, you're right drazik :)

PR #724, opened 2018-11-28
Fix/wizard styles

GoOz:
Some fixes after some feedbacks.

The removal of the linter rule is due to a bug from Stylint, that sees @block as duplicate.
Since this is a way of doing we're gonna use a lot in the near future and since Stylint development is kinda dropped, I guess disabling the rule was better than adding `// Stylint ignore` everytime.

### DIFF
--- a/.stylintrc
+++ b/.stylintrc
@@ -23,10 +23,7 @@
   "cssLiteral": "never",
   "customProperties": [],
   "depthLimit": 4,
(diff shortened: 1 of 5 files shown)


ptbrowne:
LGTM. Thanks !

Crash--:
Cool, something new to share on our next sharing meeting? ;)

PR #728, opened 2018-11-30
fix: 🐛 table divider z-index & color

GoOz:
Removed the problematic pseudo-element `::before`

### DIFF
--- a/stylus/components/table.styl
+++ b/stylus/components/table.styl
@@ -1,5 +1,6 @@
 @require '../settings/breakpoints'
 @require '../settings/palette'
+@require '../settings/z-index'
 @require '../tools/mixins'
 
 $table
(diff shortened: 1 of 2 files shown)


y-lohse:
The pseudo-element was being displayed on top of the divider, which can contain text (as it does in contacts for example). Not sure why we were using a pseudo-element in the first place.

stylus/components/table.styl
```diff
@@ -93,23 +93,15 @@ $table-cell--primary
 
 $table-divider
     position sticky
+    z-index 1
```
y-lohse: Do we need the `z-index`? And if so shouldn't we use a variable if we do?

GoOz: Not really. It's just to make sure that it goes over relative positionned elements such as the avatar in Contacts. 
Without it any element positionned, whether its relative, absolute, fixed or sticky, and with a `z-index` of 0 that is later in the DOM will go over the divider.

It's not really a layer context like alerts, modals, etc can be.
I guess I could give it the `$alert-index-mobile` which is the lower before 0 but It seems a little bit confusing to me. Or maybe I could add a new `z-index` value of 1 in the settings for cases like this.
What do you think?

PR #729, opened 2018-11-30
[WIP] refactor: Use a JSON a source for all colors

edas:
Theme.json as source for our colors. It uses an array for RGBA
instead of an #RRGGBB because:
- this allows us to have an alpha transparency (#RRGGBBAA is
not supported by some browsers)
- this allows apps to get the color components without having
to manually decode a string

For everyone else, compatibility is garanteed in react/palette.js
and stylus/settings/palette.styl, so this should have no impact.

```
yarn build:doc:react
```


ptbrowne:
I don't understand why we use the array for colors.

Can we have an example of something that was hard to do before and that is made possible with this ?

edas:
> I don't understand why we use the array for colors.
> 
> Can we have an example of something that was hard to do before and that is made possible with this ?

Colors are used : 
1. by `style/settings/palette.styl` to create CSS variables
2. by some react files outside of cozy-ui when they need to modify color codes instead of reusing them directly

We could let everyone in the second cas decode themself a #RRGGBBAA or a rgba(x,x,x,x) string to extract the r/g/b/a components before manipulation but it seems easier to me to directly having the raw data at source.

Exemple https://github.com/cozy/cozy-drive/pull/1267/files where they use direct color codes to add alpha transparency.

kosssi:
Why you don't use directly `rgba(x, x, x, o)` directly on `theme.json`?

ptbrowne:
> We could let everyone in the second cas decode themself a #RRGGBBAA or a rgba(x,x,x,x) string to extract the r/g/b/a components before manipulation

In stylus, the red(), blue(), and green() can be used to extract easily the individual components.

>  but it seems easier to me to directly having the raw data at source.

Easier than what ? I am sorry, there are no comparison between two codes here, I have trouble seeing what is being compared to what.

edas:
Ok, I was expecting that some JS code did get the color code, extract the components and recreate a new modified color code afterwards. I did not find one after a quick search so I may be wrong.

Do you prefer #RRGGBBAA or rgba(x,x,x,x) ? the color will be translated to rgba(x,x,x,x) by stylus anyway in the resulting CSS

edas:
Updated with rgb() and rgba() color codes, based on the new master

I had to write again a palette.js. I don't think it's a problem but maybe @ptbrowne should let me know if we should avoir it.

nono:

PR #733, opened 2018-12-03
refactor around className

kosssi:
~~When I would like update `MainTitle` with my className it's override by default style.~~

~~I propose to put className always at the end~~

It's a refactoring to use cx everywhere and className on the end...


CPatchane:
Should we consider this as a breking change? We could see some changes in the styles due to the overwriting order changer here.

ptbrowne:
LGTM. Yes this is is a breaking change.

PR #736, opened 2018-12-04
Handle AppIcon update

gregorylegarec:
We faced a case when AppIcons were nested in lists and keeping
their state, i.e. displaying the wrong icon.

See capture.

capture_du_2018-11-30_12-10-45

This PR fixes this behavior and add a related test.

### DIFF
--- a/react/AppIcon/index.jsx
+++ b/react/AppIcon/index.jsx
@@ -17,19 +17,37 @@ export class AppIcon extends Component {
     super(props, context)
     const { app, domain, secure } = props
     const preloaded = getPreloaded(app, domain, secure)
-    this.state = {
-      error: null,
-      icon: preloaded,
-      status: preloaded ? DONE : FETCHING
-    }
+    this.state = this.buildState(preloaded)
   }
 
-  componentDidMount() {
-    this.load()
+  async componentDidMount() {
+    const { onReady } = this.props
+    const { status } = this.state
+    if (status !== DONE) {
+      await this.load()
+    }
+    if (typeof onReady === 'function') onReady()
+  }
+
+  // When nested into a list of component, an AppIcon may update,
+  // and we must then update the state
+  async componentDidUpdate(prevProps) {
+    const { app, domain, onUpdate, secure } = this.props
+
+    if (app.slug === prevProps.app.slug) return
+
+    const preloaded = getPreloaded(app, domain, secure)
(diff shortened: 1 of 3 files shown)


react/AppIcon/index.jsx
```diff
@@ -25,7 +25,28 @@ export class AppIcon extends Component {
   }
 
   componentDidMount() {
-    this.load()
+    const { status } = this.state
+    if (status !== 'DONE') {
```
CPatchane: Is it not supposed to be `status !== DONE` since you have variable above?

react/AppIcon/index.jsx
```diff
@@ -25,7 +25,28 @@ export class AppIcon extends Component {
...
+
+    const preloaded = getPreloaded(app, domain, secure)
+
+    if (preloaded) {
+      this.setState({
+        icon: preloaded
+      })
+    } else {
+      this.load()
+    }
```
ptbrowne: The management of icon/status/error is now done in 3 methods : `componentDidMount`, `componentDidUpdate` and the `constructor`. I think we should try to have less methods touching this shared state.

One problem I see here is that when the icon is set, the `error` is not reset. Also since `load()` can take time to set the state to the proper icon, the wrong icon can be shown for a while. I think the 3 methods should call something like `setIcon` that would be the only manager for error/icon and status.

ptbrowne:
Another thing worth to try IMHO: React does not reuse components when their `key` changes. I see [here](https://github.com/cozy/cozy-home/blob/332fbaba97bc0dab16df8b4cc6b9c900de94c498/src/ducks/connections/components/queue/queue.jsx#L147) that the key is set to the index of the item. Maybe it could work to set the key to the slug of the konnector ?

gregorylegarec:
@ptbrowne The first thing I tried was actually to use keys, but as the Queue it was a little bit more complex than passing an id. Eventually I came with using `triggerId` as key for AppIcon and it does the job, as seen [in this PR](https://github.com/cozy/cozy-home/pull/1016/commits/393d81a991e107caf9490fe705b557a7610fe232)

nono:

gregorylegarec:
Sorry for the delay.

As https://github.com/cozy/cozy-home/pull/1016 fixed the issue in cozy-home, I think we should not commit this fix.

PR #750, opened 2018-12-07
Proposal: Don't use a single stylesheet for all transpiled components

CPatchane:
Using a single stylesheet doesn't give any choice about which component styles are imported and need a custom require in the application or the webpack configuration which to much to use cozy-ui IMHO.

Since we juste want to transpile our components, I think that we should keep the application handle the styles files required by the components on its side.

With this transpiling configuration, all components will have a single css files required in its JS code.
When the app will import the components to use it, it will automatically import the good css-transformed stylesheet related file:
screenshot 2018-12-07 at 13 36 40

It may be considered as a breaking change if some applications are still looking for a single `transpiled/stylesheet.css`

### DIFF
--- a/scripts/importPathFormatter.js
+++ b/scripts/importPathFormatter.js
@@ -1,0 +1,1 @@
+module.exports = path => path.replace(/\.styl$/, '.css')
--- a/.babelrc
+++ b/.babelrc
@@ -9,10 +9,16 @@
     "transpilation": {
       "plugins": [
         ["css-modules-transform", {
-          extensions: [".styl"],
-          preprocessCss: "./preprocess",
-          extractCss: "./transpiled/stylesheet.css",
-          generateScopedName: "[name]__[local]___[hash:base64:5]",
+          "extensions": [".styl"],
+          "preprocessCss": "./preprocess",
+          "importPathFormatter": "./scripts/importPathFormatter",
+          "extractCss": {
+              "dir": "./transpiled/react/",
+              "relativeRoot": "./react/",
+              "filename": "[path]/[name].css"
+          },
+          "generateScopedName": "[name]__[local]___[hash:base64:5]",
+          "keepImport": true
         }]
       ]
     }
--- a/.babelrc
+++ b/.babelrc
@@ -21,6 +21,15 @@
           "keepImport": true
         }]
       ]
+    },
+    "test": {
+      "plugins": [
+        ["css-modules-transform", {
+          "extensions": [".styl"],
+          "preprocessCss": "./preprocess",
+          "generateScopedName": "[name]__[local]___[hash:base64:5]"
+        }]
+      ]
     }
   },
   "ignore": ["examples/**/*"]
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "sprite": "scripts/make-icon-sprite.sh",
     "test": "yarn test:jest",
     "test:jest": "env BABEL_ENV=transpilation jest --verbose --coverage",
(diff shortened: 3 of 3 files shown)


scripts/importPathFormatter.js
```diff
@@ -0,0 +1 @@
+module.exports = path => path.replace(/\.styl$/, '.css')
```
ptbrowne: missing m ?

CPatchane: The multiline flag? Why? It should get only a pathname string, shoudn't it?

ptbrowne:
I like the idea, ideally it would mean the final application CSS would be smaller since some components would not be imported. My main worry is duplication across transpiled CSS files, for example of global styles.

>  doesn't give any choice about which component styles are imported and need a custom require in the application or the webpack configuration which to much to use cozy-ui IMHO.

I don't understand, what choice can the application have ? What do you mean with custom require ?

Can you provide some size benchmarks / to see if there is duplication ?

CPatchane:
> I don't understand, what choice can the application have ? What do you mean with custom require ?

I was thinking about importing just the styles of the needed components and not all the styles from all components. By custom require I was meaning that the application must have to import somewhere the `transpiled/stylesheet.css` file in order to be handled by webpack.

About the duplications nothing change for both solutions:
- For now, the `transpiled/stylesheet.css` contains a lot of duplicated styles code like the variables definition using the `:root` selector which is duplicated 34 times (see https://unpkg.com/cozy-ui@17.5.0/transpiled/stylesheet.css).
- Here, each styles files will be handled by webpack to be extracted and bundled into the application css file.

Both solutions will uses, thanks to [our webpack configuration](https://github.com/CPatchane/create-cozy-app/blob/cozy-scripts%401.3.0/packages/cozy-scripts/config/webpack.config.base.js#L70-L85), [PostCSS assets plugin](https://github.com/klimashkin/postcss-assets-webpack-plugin) and [`postcss-discard-duplicates`](postcss-discard-duplicates) to remove all duplicates from the application extracted css.

ptbrowne:
Mmh ok, I am still uncomfortable with shipping duplicated code on npm (I understand it is already the case now). Do you think there would be a way not to have the duplicated :root ? Users of cozy-ui should not have to de-duplicate the CSS via a plugin IMHO.

scripts/importPathFormatter.js
ptbrowne: Sorry, there must a bug in GitHub.

image

CPatchane: Oh :o

CPatchane:
@ptbrowne I don't see how to do this with 1 style per component but we have to choice between:
* Using 1 stylesheet, the application have to hande on its side the cozy-ui css sheet in order to use cozy-ui components
* Using 1 sheet/components, cozy-ui will works without doing anything with the styles but to optimize the code it will have to handle a postcss to remove duplicates

I would personaly prefer the second one since cozy-ui will works anyway compared to the first one. In all cases, cozy-ui is built to be used in a Cozy application with bundling configurations.

@Crash-- Yes it will for now (or we may transpile twice)
And this is a kind related to https://github.com/CPatchane/create-cozy-app/pull/960 to make cozy-scripts handling this automatic stylesheet import. We won't need to do that with 1 style by component here.

edas:
As Patrick, I fear a lot a duplication. Did someone try and have real figures ?

Also, I don't think we have time today to introduce breaking changes that requires other teams to change the way they build their apps

CPatchane:
Not relevant anymore since https://github.com/cozy/cozy-ui/pull/839

PR #753, opened 2018-12-11
fix: Remove postinstall and use prepare

ptbrowne:
Prepare runs the task before publishing and after running `yarn install` (when developing on cozy-ui).

> it runs before the package tarball is packed and uploaded to the registry during the publishing process, as well as when you run npm install (with no package name) after cloning a package, to prepare it for use (i.e. by transpiling source).

Postinstall would run `release` when people ran `yarn install cozy-ui` and this is not what we want (since cozy-ui is already published with transpiled versions on npm).


Crash--:
Thk!

Issue #688, opened 2018-11-19
Enhance grey colors for secondary button and link


Claiw:
Hi,

The issue concerns two components in cozy-ui https://docs.cozy.io/cozy-ui/styleguide/section-components.html#kssref-components-buttons-icons

We have this aspect : <img width="560" alt="screen shot 2018-11-16 at 18 18 04" src="https://user-images.githubusercontent.com/31621610/48698667-f4183500-ebe7-11e8-842f-dba33ef6e18b.png">

I would like to have something like that : 
it will be more coherent with for example what we have in the cozy-bar for parameters, and in the sidebar for the tabs

<img width="616" alt="screen shot 2018-11-19 at 10 47 27" src="https://user-images.githubusercontent.com/31621610/48698906-8d474b80-ebe8-11e8-8c1e-bad2437dd749.png">

**secondary button colors :**
- icon : --slateGrey _ie_ (hex)5D6165 
- label : --charcoalGrey _ie_ (hex)32363F

**secondary link colors :**  
- icon : --coolGrey _ie_ (hex)95999D
- label : --slateGrey _ie_ (hex)5D6165 

Thanks!

y-lohse:
Ooops, just saw you reopened it before @Crash-- . Was there anything left to do on this issue?

PR #768, opened 2018-12-26
feat: Change secondary subtle button color

Crash--:
I think it's a kind of breaking change since it'll change the current color. 

It's a first step to close https://github.com/cozy/cozy-ui/issues/688 

Here is the result : 
capture d ecran 2018-12-26 a 09 44 41


Crash--:
[(1) Contacts - écran d'accueil 0 contact - revue UI](https://trello.com/c/jPGXd2SS/1425-1-contacts-%C3%A9cran-daccueil-0-contact-revue-ui)

Crash--:
I tend to think that if we change a current behavior, even if it's only a few colors, we're causing a breaking change but hey if you both think not, good, I'm merging it

PR #769, opened 2018-12-26
feat: Add Info component

Crash--:
Added an info component used to display persistant information. 

capture d ecran 2018-12-26 a 09 49 47


Crash--:
[(1) Contacts - écran d'accueil 0 contact - revue UI](https://trello.com/c/jPGXd2SS/1425-1-contacts-%C3%A9cran-daccueil-0-contact-revue-ui)

react/Infos/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+        </Text>
+      </div>
+    </div>
+  )
+}
+
+Infos.defaultProps = {
+  icon: null
+}
+Infos.prototype = {
```
drazik:
```suggestion
Infos.propTypes = {
```

react/Infos/index.jsx
```diff
@@ -0,0 +1,31 @@
...
+  icon: PropTypes.any,
```
drazik: Is there a reason for not using the same propTypes as in `Icon` here ? Cf. https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L69 (minus `isRequired` obviously ;))

Crash--: You're right. I added this any at the beginning and didn't change it. 

`isRequired` is not required since we can use this component without icon see https://zpl.io/V0JGPwx

PR #770, opened 2018-12-26
Select option with prefix

kosssi:
Add classNamePrefix on Option class.

On Banks app we add `classNamePrefix='needsclick cz'` but the Option component don't use it, it's a bug we can display only on iOS ;)

https://kosssi.github.io/cozy-ui/react/#!/SelectBox

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -5,7 +5,7 @@ import styles from './styles.styl'
 import Icon from '../Icon'
 import { dodgerBlue, silver, coolGrey } from '../palette'
 import withBreakpoints from '../helpers/withBreakpoints'
-import cx from 'classnames'
+import classNames from 'classnames'
 
 const customStyles = {
   container: base => ({
@@ -73,7 +73,7 @@ const Option = ({
   <div
     {...innerProps}
     ref={innerRef}
-    className={cx(styles['select-option'], {
+    className={classNames(styles['select-option'], {
       [styles['select-option--selected']]: isSelected && !withCheckbox,
       [styles['select-option--focused']]: isFocused,
       [styles['select-option--disabled']]: isDisabled
@@ -179,7 +179,7 @@ class SelectBox extends Component {
         onMenuOpen={this.handleOpen}
         onMenuClose={this.handleClose}
         {...props}
-        className={cx(
+        className={classNames(
           {
             [styles['select__overlay']]: showOverlay
           },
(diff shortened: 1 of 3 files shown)


react/SelectBox/SelectBox.jsx
```diff
@@ -60,6 +60,11 @@ const reactSelectControl = CustomControl => ({
   </div>
 )
 
+export const withPrefix = (cx, className) => {
+  const classNameWithPrefix = cx(null, { [` ${className}`]: true })
```
drazik: Why `null` ?

react/SelectBox/SelectBox.jsx
```diff
@@ -60,6 +60,11 @@ const reactSelectControl = CustomControl => ({
...
+  return classNameWithPrefix === '' ? className : classNameWithPrefix
```
drazik: This condition is `true` only when `className` is an empty string, no ? If yes, I don't understand its purpose.

drazik: Actually no, since there's a space in ` ${className}`. I don't understand how this condition can be `true` :thinking:

kosssi: classNameWithPrefix is empty when we use `react-select` without `classNamePrefix`

react/SelectBox/SelectBox.jsx
kosssi: Here is the cx function : https://github.com/JedWatson/react-select/blob/bd3705b/src/utils.js#L42-L55
```
export function classNames(
  prefix?: string | null,
  cssKey?: string | null,
  state?: ClassNamesState,
  className?: string,
) {
  const arr = [cssKey, className];
  if (state && prefix) {
    for (let key in state) {
      if (state.hasOwnProperty(key) && state[key]) {
        arr.push(`${applyPrefixToName(prefix, key)}`);
      }
    }
  }

  return arr.filter(i => i).map(i => String(i).trim()).join(' ');
}
```
The first parameter is bind by react-select [here](https://github.com/JedWatson/react-select/blob/9690b28dd7713b68728e56075d6819e6898ec430/src/Select.js#L692), so in this place `null` is `cssKey`, it's not necessary.

react/SelectBox/SelectBox.jsx
```diff
@@ -60,6 +60,20 @@ const reactSelectControl = CustomControl => ({
   </div>
 )
 
+export const withPrefix = (cx, className) => {
+  // react-select implement a classnames function https://git.io/fhT8S
+  // it's not the same of classnames library
+  // 1st parameter is bind by react-select with prefix https://git.io/fhkIj
+  // 2nd parameter is cssKey we don't need so null is send
+  // 3rd parameter is used to add prefix but we don't want to stick with
+  //   webpack className so we add space ' '
```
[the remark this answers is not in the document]
kosssi: I would like to say "coller" in french :D

PR #775, opened 2019-01-02
feat(InputPassword): Add InputPassword component

sizvix:
Add InputPassword witch purpose a button to show/hide the string in the input.
The button could be disabled with the props "askShowHide" to "false".

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -1,15 +1,59 @@
 import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
+
 import styles from './styles.styl'
 import Label from '../Label'
+import { default as LabelStyles } from '../Label/styles.styl'
 import Input from '../Input'
 import SelectBox from '../SelectBox'
 import Textarea from '../Textarea'
 
+class InputPassword extends React.Component {
+  state = {
+    visible: false
+  }
+
+  toggleVisibility() {
+    this.setState({ visible: !this.state.visible })
+  }
+
+  render() {
+    const { hideLabel, showLabel, showVisibilityButton } = this.props
+    const { visible } = this.state
+    return (
+      <div className={styles['o-field-input']}>
+        {showVisibilityButton && (
+          <div
+            className={cx(
+              LabelStyles['c-label'],
+              styles['o-field-input-action']
+            )}
+            onClick={() => this.toggleVisibility()}
+          >
+            {visible ? hideLabel : showLabel}
+          </div>
+        )}
+        <Input {...this.props} type={visible ? 'text' : 'password'} />
+      </div>
+    )
+  }
+}
+
+InputPassword.propTypes = {
+  hideLabel: PropTypes.string,
+  showLabel: PropTypes.string,
+  showVisibilityButton: PropTypes.bool
+}
+
+InputPassword.defaultProps = {
+  showVisibilityButton: true
+}
+
 const Field = props => {
   const {
     className,
+    fieldProps,
     fullwidth,
     label,
     id,
@@ -19,6 +63,7 @@ const Field = props => {
     error,
     onChange,
     readOnly,
+    secondaryLabels,
     size
   } = props
 
@@ -39,10 +84,23 @@ const Field = props => {
             readOnly={readOnly}
           />
         )
-      case 'text':
       case 'password':
+        return (
+          <InputPassword
+            id={id}
+            type={type}
+            placeholder={placeholder}
+            value={value}
+            error={error}
+            onChange={onChange}
+            readOnly={readOnly}
+            {...fieldProps}
+            {...secondaryLabels}
+          />
+        )
       case 'email':
       case 'url':
+      case 'text':
         return (
           <Input
             fullwidth={fullwidth}
@@ -72,6 +130,7 @@ const Field = props => {
 }
 
 Field.PropTypes = {
+  fieldProps: PropTypes.object,
   fullwidth: PropTypes.bool,
   label: PropTypes.string.isRequired,
   id: PropTypes.string,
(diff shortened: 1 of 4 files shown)


Crash--:
Can you add : 
- snapshot 
- an exemple

GoOz:
There's the how to in the PR template 

```
yarn build:doc:react
```

sizvix:
@GoOz : That was one of our reflexions, but it need to transform Field witch is a simple function to a class extending React component , just for one face of Field ... ( but that's simple ... )
Or just wrap the field in an element witch extends React component.
The way I prefer could be have a place after the label for generic "inputInteractionElement" , witch here is "show/hide buton" , on a `type="date"` it could be "today" , etc ...

react/InputPassword/styles.styl
```diff
@@ -0,0 +1,8 @@
+.wrapInputPassword
```
gregorylegarec: We use to use dash-case for or CSS class names. We are also prefixing or component related class names with `c-`.

Actually it was not in our guidelines, I juste made a PR to fix that. https://github.com/cozy/cozy-guidelines/pull/57

react/InputPassword/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+import cx from 'classnames'
+import { default as LabelStyles } from '../Label/styles.styl'
+import Input from '../Input'
+import styles from './styles.styl'
+
+class InputPassword extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { visible: false }
+    this.askShowHide = props.askShowHide !== false
```
gregorylegarec: I think `visibilityButton` (`true`/`false`) is a better name for this prop. What do you think ?

By the way it should be only taken into account in the render method.

react/InputPassword/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  }
+  render() {
+    return (
+      <div className={styles['wrapInputPassword']}>
+        {this.askShowHide && (
+          <div
+            className={cx(LabelStyles['c-label'], styles['actionOnInput'])}
+            onClick={() => this.toggleVisibility()}
+          >
+            {this.state.visible ? 'HIDE' : 'SHOW'}
```
gregorylegarec: We cannot let untranslated text anywhere in our code.

Also, this line ask a question : we need to translate component from Cozy-UI and it has never been done before. A quick solution could be to pass the label as props (maybe `buttonHideLabel` and `buttonShowLabel`), waiting for a more global solution provided by Cozy-UI.

react/InputPassword/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+          </div>
+        )}
+        <Input
+          {...this.props}
+          type={this.state.visible ? 'text' : 'password'}
```
gregorylegarec: As `this.state.visible` is used two times in this methond, maybe we can retrieve it at the beginning of the method with 
```js
const { visible } = this.state
```

react/InputPassword/styles.styl
GoOz: https://docs.cozy.io/cozy-ui/styleguide/ there is guidelines but not in cozy-guidelines I admit.

react/InputPassword/index.jsx
ptbrowne: Finally the `translate` signature could serve to something :) Default translating strings could be passed directly as options

```
export default translate({
  en: { show: 'Show', hide: 'Hide' },
  fr: { show: 'Montrer', hide: 'Cacher' }, 
})(InputPassword)
```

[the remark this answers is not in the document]
sizvix: temporary with props ...

react/InputPassword/styles.styl
sizvix: @GoOz is it good now ?

GoOz: where is the link?

sizvix: [new styles.styl](https://github.com/cozy/cozy-ui/blob/fc7206c7b78148dffc48ad519e77f8a042717513/react/Field/styles.styl)
I add the classes : `.wrap-input-password` and `.action-on-input`  :)

react/Field/index.jsx
```diff
@@ -67,25 +124,29 @@ const Field = props => {
...
   value: PropTypes.string,
   placeholder: PropTypes.string,
   error: PropTypes.bool,
-  size: PropTypes.oneOf(['tiny', 'medium', 'large'])
+  size: PropTypes.oneOf(['tiny', 'medium', 'large']),
+  secondaryLabels: PropTypes.object
 }
 
 Field.defaultProps = {
+  enableExtraFunction: true,
```
gregorylegarec: This name is not very explicit and can be confusing as we can expect a function here.

After discussing with @ptbrowne, the best way to achieve this seems to pass `showVisibilityButton` inside a `fieldProps` prop.

gregorylegarec:
As there were only a few remark remaining I made the change directly. Does @GoOz and @Crash--  may approve changes ? thanks !

react/Field/index.jsx
```diff
@@ -34,10 +79,23 @@ const Field = props => {
...
+          <InputPassword
+            id={id}
+            type={type}
+            placeholder={placeholder}
+            value={value}
+            error={error}
+            onChange={onChange}
+            readOnly={readOnly}
+            {...fieldProps}
+            {...secondaryLabels}
```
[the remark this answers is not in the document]
gregorylegarec: I think that it shows a bad pattern at some place. Maybe InputPassword should be passed as prop to Field. It implies a little rework of our Field component.

gregorylegarec:
Thanks @sizvix 👍

PR #777, opened 2019-01-02
Wrap jest command

gregorylegarec:
By doing this we ensure that running 'yarn jest' will use the expected
transpilation environment variable, and we avoid differences in jest
snapshots.


package.json
```diff
@@ -37,6 +37,7 @@
     "release": "npm-run-all --parallel sprite palette && yarn transpile",
     "semantic-release": "semantic-release",
     "sprite": "scripts/make-icon-sprite.sh",
+    "jest": "yarn test",
```
ptbrowne:
```suggestion
    "jest": "yarn test:jest",
```

PR #780, opened 2019-01-03
fix: Use the correct Info SVG and fix Info > SVG css

Crash--:
I was not using the right SVG file for the Info icon

I also fixed an issue with the current CSS of `<Infos` component


react/Infos/styles.styl
```diff
@@ -12,7 +12,8 @@
 .infos--icon
   width rem(16)
   height rem(16)
-  padding-top: 0.2em
+  padding-top 0.2em
```
GoOz: I know it's not really what you changed in this PR but I prefer to have `.2rem` (removing the useless 0 and using `rem` instead of `em` it really isn't the same thing) or use the mixin `rem(3)` (implicit pixel unit).
Mostly for consistency but also because `em` is not what we want for sizing our elements.

Crash--: Done

PR #782, opened 2019-01-04
fix(Subtle button): Fix hover color

Crash--:
After a discussion with @Claiw : 
- change hover color for txt & svg


Claiw:
i am OK

PR #785, opened 2019-01-07
Improve SelectBox

gregorylegarec:
* Make it usable as Field by handling the type `dropdown`.
* Handle fullwidth and size  props to be compatible with Field props

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import styles from './styles.styl'
 import Label from '../Label'
 import Input from '../Input'
+import SelectBox from '../SelectBox'
 import Textarea from '../Textarea'
 
 const Field = props => {
@@ -23,6 +24,10 @@ const Field = props => {
 
   const inputType = type => {
     switch (type) {
+      case 'select':
+        // If value prop is falsy, the SelectBox never displays any selected option.
+        // Only passing `undefined` make the SelectBox work properly.
+        return <SelectBox {...props} value={props.value || undefined} />
       case 'textarea':
         return (
           <Textarea
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -25,6 +25,23 @@ const options = [
 <SelectBox options={options} menuIsOpen />
 ```
 
+### Props
+
+#### `fullwidth`
+
+Set the select to spread to 100% of the available width (default: `false`).
+
+```
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' },
+  { value: 'long', label: 'Salt and vinegar crisps with vegamite and brussel sprouts, double chai latte sauce' },
+];
+
+<SelectBox options={options} fullwidth />
+```
+
 ### Custom control component
 
 You can use the `reactSelectControl` HOC to turn an existing, custom component into the control button for the `Select`:
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -42,6 +42,36 @@ const options = [
 <SelectBox options={options} fullwidth />
 ```
 
+#### `size`
+
+Set the text size between `tiny`, `medium` and `large` (default: `large`).
+
+```
+const tinyOptions = [
+  { value: '1', label: 'I am a tiny SelectBox' },
+  { value: '2', label: 'I am not a medium SelectBox' },
+  { value: '3', label: 'I am not a large SelectBox' }
+];
+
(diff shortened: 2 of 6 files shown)


react/Field/index.jsx
```diff
@@ -23,6 +24,8 @@ const Field = props => {
 
   const inputType = type => {
     switch (type) {
+      case 'dropdown':
```
y-lohse: Should we call this `select` instead of `dropdown`? The other field names all seem to use names from the HTML spec.

react/SelectBox/SelectBox.jsx
```diff
@@ -7,11 +7,15 @@ import { dodgerBlue, silver, coolGrey } from '../palette'
...
 
-const customStyles = {
-  container: base => ({
-    ...base,
-    maxWidth: '30rem'
-  }),
+// Heights compatible with Field component heights. We need to add an extra
+// pixel for each top and bottom border. Final heights will be two pixels less.
+const heights = {
+  tiny: '34px',
```
GoOz: why not `rem` values ?

react/SelectBox/styles.styl
```diff
@@ -7,6 +8,13 @@
     height 0
     overflow hidden
 
+.select--autowidth
```
GoOz: Why not adding this `max-width` by default and removing it when you want fullwidth ?

react/Field/index.jsx
gregorylegarec: Actually we are using `dropdown` in `manifest.konnector` files, and it really. So I passed directly the field from manifest to our form.

But let's go for a select, we will perform the mapping directly into Cozy-Harvest-Lib 👍

react/SelectBox/SelectBox.jsx
gregorylegarec: Let's try them !

gregorylegarec: Do we agree on `2.125rem` ?

react/SelectBox/styles.styl
gregorylegarec: Actually I need to use a CSS class. Otherwise it's the customStyle from SelectBox which always have priority, I don't figured out why.

react/SelectBox/SelectBox.jsx
GoOz: Sizes are supposed to be:
- 32px
- 40px
- 48px
All those are Including the paddings and borders.

If the `box-sizing` is, as I think it is, `border-box` then those size in rem should be:
- 2rem
- 2.5rem
- 3rem

gregorylegarec: Yep, put react-select seems to remove 2px for free, so I added them in the code. Want a demo ?

GoOz: yes please :/ that's an odd behaviour

gregorylegarec:
@y-lohse Of course https://gregorylegarec.github.io/cozy-ui/react/#selectbox

gregorylegarec:
I updated the first commit with an example, which conduced me to fix the value passed as prop.

react/Field/index.jsx
```diff
@@ -23,6 +24,9 @@ const Field = props => {
 
   const inputType = type => {
     switch (type) {
+      case 'select':
+        // Falsy value other than `undefined` cannot let the select have a value
```
ptbrowne: I don't understand this sentence ^

react/SelectBox/Readme.md
```diff
@@ -25,6 +25,53 @@ const options = [
...
+```
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' },
+  { value: 'long', label: 'Salt and vinegar crisps with vegamite and brussel sprouts, double chai latte sauce' },
+];
+
+<SelectBox options={options} fullwidth />
+```
```
ptbrowne: Since it deals with `size`, could we simplify the API by having `size='full'` instead of `fullwidth ?

react/SelectBox/SelectBox.jsx
GoOz: In your styleguide you set 2.125rem and it has a height of 2.125rem. I don't see where those 2px are removed :/

react/SelectBox/Readme.md
gregorylegarec: The idea here is to use the same API as Field component. In Field, size is used for component's height.

react/Field/index.jsx
gregorylegarec: I was so tired. I will clarify it 😅

react/SelectBox/Readme.md
ptbrowne: OK thanks.

PR #786, opened 2019-01-07
feat(AppIcon): Add possibility to get the icon from an app slug

drazik:
Currently we can only get an app icon by giving an object with slug and latest_version properties to `AppIcon`. In Banks, we will need to get konnectors icons from their slug.

The idea here is to make `AppIcon` accept a string as its `app` prop. If we pass a string, it will always fetch the icon url from the registry via the [`/registry/:app/icon` route](https://github.com/cozy/cozy-stack/blob/master/docs/registry.md#get-registryappicon).

### DIFF
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -46,6 +46,10 @@ const _getInstalledIconPath = app => {
 }
 
 const _getRegistryIconPath = app => {
+  if (typeof app === 'string') {
+    return `/registry/${app}/icon`
+  }
+
   return (
     app.latest_version &&
     app.latest_version.version &&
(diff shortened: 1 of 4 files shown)


react/AppIcon/Preloader.js
```diff
@@ -46,6 +46,10 @@ const _getInstalledIconPath = app => {
 }
 
 const _getRegistryIconPath = app => {
+  if (typeof app === 'string') {
```
gregorylegarec: I am not a huge fan of having a prop which can have several types but it seems the better thing to do here to avoid a huge refactoring.

drazik: Yep. I think it's better to have a single prop that can be of multiple types than to have several props that would make impossible states possibles (or more complicated to validate).

With a single prop, we just have to check its type. With multiple props, we may end with `app={{ slug: 'stuff', latest_version: '1.2.3' }}` and `slug="thing"` and the same time. In that case we have to decide which one to take. It can be complicated IMHO.

PR #790, opened 2019-01-10
Fix Empty component icon proptype

cedricmessiant:
The icon prop type should match the type of Icon component's icon prop

### DIFF
--- a/react/ContextHeader/index.jsx
+++ b/react/ContextHeader/index.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Button } from '../Button'
 import { Title, Text } from '../Text'
-import Icon from '../Icon'
+import Icon, { iconPropType } from '../Icon'
 import styles from './styles.styl'
 import palette from '../palette'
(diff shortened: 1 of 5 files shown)


cedricmessiant:
@ptbrowne done!

cedricmessiant:
Now it's much better, thanks :-)

ptbrowne:
:shipit:

react/ContextHeader/index.jsx
```diff
@@ -36,7 +36,7 @@ const ContextHeader = ({ icon, title, text, onClose, className }) => {
 }
 
 ContextHeader.propTypes = {
-  icon: PropTypes.node,
+  icon: iconPropType,
```
Crash--: Hmm `icon: PropTypes.oneOfType([PropTypes.node, iconPropType]),` ?

Crash--: Or we add node to the iconPropType ;)

cedricmessiant: as `icon` prop is directly passed to `Icon` component, I think it shouldn't be a node.

PR #794, opened 2019-01-14
feat: Ui theme for MUIMenu

GoOz:
[Styleguidist preview](https://gooz.github.io/cozy-ui/react/#menus)

Still glitching and/or missing:
- ~~The `Grow` transition only works when opening but not when closing.~~
- Z-index issue on the `Popper`, I'm still trying to figure out if there's a way to get a higher z-index without using styles like the legacy `MuiMenu` that "uses" Modal styles… ¯\_(ツ)_/¯
- Getting the drawer on mobile
- Put styles in `theme.json` only

### DIFF
--- a/react/MuiCozyTheme/Menus/index.js
+++ b/react/MuiCozyTheme/Menus/index.js
@@ -1,3 +1,7 @@
+// This component is inspired by https://material-ui.com/demos/menus/#menulist-composition
+// Since the MuiMenu component doesn't allow another Popover positioning,
+// we have to recompose the Menu component ourselves with basic Mui component
+
 import React, { Component } from 'react'
 
 import MenuButton from '../../Button'
--- a/react/MuiCozyTheme/Menus/index.js
+++ b/react/MuiCozyTheme/Menus/index.js
@@ -1,3 +1,122 @@
-import Menu from '@material-ui/core/Menu'
+import React, { Component } from 'react'
 
-export default Menu
+import MenuButton from '../../Button'
+import PropTypes from 'prop-types'
+
+import ClickAwayListener from '@material-ui/core/ClickAwayListener'
+import Grow from '@material-ui/core/Grow'
+import Popper from '@material-ui/core/Popper'
+import Paper from '@material-ui/core/Paper'
+import MenuList from '@material-ui/core/MenuList'
+import { withStyles } from '@material-ui/core/styles'
+
+const styles = {
+  root: {
+    position: 'relative'
+  },
+  paper: {
+    '& hr': {
+      margin: '8px 0',
+      border: 0,
+      borderTop: '1px solid var(--silver)'
+    }
+  }
+}
+
+class Menu extends Component {
+  state = { open: false, anchorEl: null }
+
+  toggle = event => {
+    const { currentTarget } = event
+    this.setState(state => ({
+      anchorEl: currentTarget,
+      open: !state.open
+    }))
+  }
+
+  close = () => {
+    this.setState({
+      open: false
+    })
+  }
+
+  render() {
+    const {
+      label,
+      disabled,
+      buttonClassName,
+      component,
+      children,
+      placement,
+      classes
+    } = this.props
+    const { open, anchorEl } = this.state
+    const offset = {
+      offset: {
+        offset: '0, 4'
+      }
+    }
+
+    return (
+      <ClickAwayListener onClickAway={this.close}>
+        <div className={classes.root}>
+          {!component ? (
+            <MenuButton
+              disabled={disabled}
+              onClick={this.toggle}
+              label={label}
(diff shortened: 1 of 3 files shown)


react/MuiCozyTheme/Menus/index.js
```diff
@@ -1,3 +1,121 @@
...
+import Paper from '@material-ui/core/Paper'
+import MenuList from '@material-ui/core/MenuList'
+import { withStyles } from '@material-ui/core/styles'
+
+const styles = {
+  paper: {
+    '& hr': {
+      margin: '8px 0',
+      border: 0,
+      borderTop: '1px solid #D6D8DA'
```
kosssi:
```suggestion
      borderTop: '1px solid var(--silver)'
```

react/MuiCozyTheme/Menus/index.js
```diff
@@ -1,3 +1,121 @@
...
+            React.cloneElement(component, {
+              disabled,
+              onClick: this.toggle
+            })
+          )}
+          {
+            <Popper
+              open={open}
+              anchorEl={anchorEl}
+              transition={open}
```
ptbrowne: I think this is why the transition does not work when closing. Since `transition={false}` in this case, transitions are not activated and Popper cannot support a Transition component like `Grow`.

> transition: Help supporting a react-transition-group/Transition component.

GoOz:
[Theme Cozy pour menu M-UI](https://trello.com/c/GDaNLILf/340-theme-cozy-pour-menu-m-ui)

react/MuiCozyTheme/Menus/index.js
GoOz: Yeah I didn't really cleaned the CSS yet but I changed that anyway.

kosssi:
Il y a un problème de scroll sur le composant https://gyazo.com/a21559cf4d66680eb116a69132845467

react/MuiCozyTheme/Menus/index.js
```diff
@@ -1,3 +1,122 @@
...
+              label={label}
+              buttonClassName={buttonClassName}
+            />
+          ) : (
+            React.cloneElement(component, {
+              disabled,
+              onClick: this.toggle
+            })
+          )}
+          {
```
kosssi: J'ai l'impression que `{` n'est pas indispensable ici ?

react/MuiCozyTheme/Menus/index.js
```diff
@@ -1,3 +1,122 @@
...
+  placement: 'bottom-end'
+}
+
+Menu.propTypes = {
+  /** Disables the menu */
+  disabled: PropTypes.bool,
+  /** Specifies a custom component for the opener */
+  component: PropTypes.element,
+  /** placement for the popover */
+  placement: PropTypes.oneOf(['bottom-start', 'bottom-end'])
```
ptbrowne: I think we should keep `transformOrigin` syntax, it is known by everybody doing CSS and it is more versatile. It also maps directly to the Grow interface.

react/MuiCozyTheme/Menus/index.js
```diff
@@ -1,3 +1,122 @@
...
+
+  render() {
+    const {
+      label,
+      disabled,
+      buttonClassName,
+      component,
+      children,
+      placement,
+      classes
```
ptbrowne: `s/classes/className` ?

GoOz: Mmmmh nope, `classes` refers to classes set in the `styles` const. 
Something M-UI proposes to override classes.

GoOz:
Should it be merged as is? It is functional and not breaking.

ptbrowne:
What about the other tasks in the description ? Are they done or do we choose that they are not that important ? Kosssi reported a problem with scroll also, is it solved ? On the material UI site, it is not possible to scroll when a menu is opened, it should fix the problem easily what do you think ?

GoOz:
@ptbrowne 

> It seems this is inspired by https://material-ui.com/demos/menus/#menulist-composition , I think having this URL inside the code would help to understand why we've done that and did not use the original Menu from material-ui.

I can do that :)

> What about the other tasks in the description ? Are they done or do we choose that they are not that important ? 

I think the other tasks would be nice to have but not required so far. It can be added later if really needed.

> Kosssi reported a problem with scroll also, is it solved ? 

Yes.

> On the material UI site, it is not possible to scroll when a menu is opened, it should fix the problem easily what do you think ?

The thing is the `<Menu>` component is somehow using Modal behavior applied to the popover, not sure how to do that. The Menulist composition demo doesn't really copy eveything the Menu does. There may be some further investigation to do to make this component better but I'm having a hard time taming Mui.

ptbrowne:
Merci pour les éclaircissements, LGTM !

PR #795, opened 2019-01-15
SelectBox: numLines prop

gregorylegarec:
This PR is a proposal to add a `numLines` prop to SelectBox.

In Harvest we need to limit the size of the menulist to avoid overflow.

Another solution could be to use a Portal but Cozy-Home is still using preact and the Portal fails.

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -45,6 +45,12 @@ const customStyles = props => ({
   menu: base => ({
     ...base,
     zIndex: 10
+  }),
+  menuList: base => ({
+    ...base,
+    maxHeight: props.numLines
+      ? `${props.numLines * 2.125 + 0.25}rem`
+      : base.maxHeight
   })
 })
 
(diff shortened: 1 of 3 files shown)


react/SelectBox/SelectBox.jsx
```diff
@@ -45,6 +45,12 @@ const customStyles = props => ({
   menu: base => ({
     ...base,
     zIndex: 10
+  }),
+  menuList: base => ({
+    ...base,
+    maxHeight: props.numLines
+      ? `${props.numLines * 2.125 + 0.25}rem`
```
y-lohse: What is `2.125 + 0.25` ? Height of a line + padding ?

gregorylegarec: It misses a comment, my bad (but you're right).

gregorylegarec:
Thanks for your feedback @y-lohse.

Actually I was not a huge fan of this solution, that was I write _proposal_  in the description.

I think #803 is a better solution to address our problem (but harder and longer to code). So let's close this PR and focus on the other solution.

PR #799, opened 2019-01-15
feat: Add ExpansionPanel component

drazik:
This adds the customisation of the MUI's `ExpansionPanel`, `ExpansionPanelSummary` and `ExpansionPanelDetails`.

Styleguide demo: https://drazik.github.io/cozy-ui/react/#expansionpanel


react/MuiCozyTheme/ExpansionPanel/Readme.md
```diff
@@ -0,0 +1,33 @@
+```
+const MuiCozyTheme = require('..').default
```
edas: Why are you using `require()` instead of `import`?

react/MuiCozyTheme/ExpansionPanel/theme.js
```diff
@@ -0,0 +1,23 @@
...
+const borderRadius = {
+  borderTopLeftRadius: radiusValue,
+  borderTopRightRadius: radiusValue,
+  borderBottomLeftRadius: radiusValue,
+  borderBottomRightRadius: radiusValue
+}
+
+const MuiExpansionPanel = {
+  root: {
+    ...borderRadius,
```
edas: Shouldn't this be in the "shape/borderRadius" at the root of the materialUI theme to be the default for us instead of overriding individual components ?

react/MuiCozyTheme/ExpansionPanel/theme.js
```diff
@@ -0,0 +1,23 @@
...
+    boxShadow: '0 4px 12px 0 rgba(0, 0, 0, 0.08)',
```
edas: Same here. Shouldn't we define our list of shadows in the appropriate property at the root of the theme instead of individual propoerties in each components?

react/MuiCozyTheme/ExpansionPanel/Readme.md
drazik: Actually just because I copy/pasted the `Button` readme, and it uses `require`. I just tested with `import` and got an error:

```
CompileError: import is not supported (1:0)
1 : import MuiCozyTheme from '../index.js'
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

It seems styleguidist doesn't support `import` by default.

react/MuiCozyTheme/ExpansionPanel/Readme.md
```diff
@@ -0,0 +1,33 @@
...
+const ExpansionPanel = require('@material-ui/core/ExpansionPanel').default
+const ExpansionPanelSummary = require('@material-ui/core/ExpansionPanelSummary').default
+const ExpansionPanelDetails = require('@material-ui/core/ExpansionPanelDetails').default;
+
+<MuiCozyTheme>
+  <ExpansionPanel>
```
kosssi: can you add an example with same button on Cozy Banks?

kosssi:
I thinks there is a mistake marge 
capture d ecran 2019-01-15 a 16 59 37

react/MuiCozyTheme/ExpansionPanel/Readme.md
edas: well, there an import later in `react/MuiCozyTheme/ExpansionPanel/index.js`. Is it because you looked for index.js and not index.jsx ?

react/MuiCozyTheme/ExpansionPanel/theme.js
drazik: Actually I didn't know this `theme.shape.borderRadius` thing. I think you're right.

react/MuiCozyTheme/ExpansionPanel/theme.js
drazik: Definitely yes.

react/MuiCozyTheme/ExpansionPanelSummary/theme.js
```diff
@@ -0,0 +1,36 @@
+const MuiExpansionPanelSummary = {
+  expanded: {},
+  root: {
+    backgroundColor: 'var(--paleGrey)',
+    textTransform: 'uppercase',
```
kosssi: I'm not sure this is a default value.

drazik: You may be right, but since we don't have another example with another design for now, I think we can keep it.

drazik:
Added a customized example in the doc and fixed the extra margin spotted by @kosssi.

drazik:
@GoOz It seems I added the example with an `expandIcon` at the same time you commented.

react/MuiCozyTheme/ExpansionPanel/Readme.md
ptbrowne: It does not support `import` inside the examples.

drazik:
Just updated the styleguide. You can see that there is a problem when we put the `borderRadius` in `theme.shape.borderRadius`. Even if https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/ExpansionPanel/ExpansionPanel.js#L45-L56 is using `theme.shape.borderRadius`, the result radius value is `2px` instead of the `6px` we provided in our theme. What's weird is that this value seems to come from nowhere since the value of the default theme is `4`.

react/MuiCozyTheme/ExpansionPanel/theme.js
drazik: Actually there are two problems. The first is that I can't access the `theme.shadows` variable from here.

By default, the `ExpansionPanel` is using the first elevation level (see https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/ExpansionPanel/ExpansionPanel.js#L163). That means that we have to either redefined the elevation shadows scale (that's a scale on 25 levels, see https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/styles/shadows.js) or not set the `boxShadow` generically and leave it to the app (I would use `withStyles` in Banks to provide the custom `boxShadow`, for example).

What do you think ?

drazik:
I'm merging it so we can use it in apps. But I'm going to try to figure out two things asap:

* Why does the border radius is not good in the styleguide?
* How to handle the box shadow properly.

PR #803, opened 2019-01-16
SelectBox: Add boundaries prop to avoid overflow

gregorylegarec:
In Cozy-Harvest-Lib we faced an overflow issue with the SelectBox component:

capture d ecran 2019-01-16 a 16 47 54

This issue could be resolved in several way:
* Limit the height
* Use a React Portal to show the menu list below all the other element in the DOM

There is already an height limitation in PR #795.

Portaling does not work with preact, and Cozy-Home is still in Preact.

In this PR we propose another solution: give an element as boundaries, and
not expand the menu list outside those boundaries.

With this PR, our first issue is resolved and our SelectBox in Cozy-Harvest-Lib
now looks like this:

capture d ecran 2019-01-16 a 18 11 55

You can have a preview [here](https://gregorylegarec.github.io/cozy-ui/react/#!/SelectBox/5).

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import ReactSelect, { components } from 'react-select'
+
 import styles from './styles.styl'
 import Icon from '../Icon'
 import { dodgerBlue, silver, coolGrey } from '../palette'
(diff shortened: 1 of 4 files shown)


react/SelectBox/SelectBox.jsx
```diff
@@ -204,30 +263,41 @@ class SelectBox extends Component {
...
-        )}
-        // react-select temporarily adds className to its innerComponents
-        // but this behavior will soon be removed. For the moment, we
-        // cancel it by setting it to empty string
-        classNamePrefix={classNamePrefix || ''}
-      />
+      <div
+        ref={element => {
+          this.element = element
+        }}
```
y-lohse: Why don't we put the ref on the `ReactSelect` directly?

CPatchane: We tried but ReactSelect doesn't forward the ref here, so you got the ReactSelect instance which is not a correct React node

gregorylegarec:
I updated this PR with a simpler way to determine space left, based only on `getBoundingClientRect()`, and changed the prop name from `boundaries` to `container`.

@ptbrowne Do you see a way to simplify even more ? How would you test this ?

react/SelectBox/SelectBox.jsx
```diff
@@ -48,6 +50,44 @@ const customStyles = props => ({
...
+const computedMenuListHeightStyles = (container, selectElement) => {
+  if (!(container && selectElement)) return {}
+
+  const isPreactRef = !container.current
+  const actualContainer = isPreactRef ? container : container.current
+
+  const containerPaddingTop =
+    get(actualContainer, 'style.paddingTop') || '0px'
+  const containerPaddingBottom =
+    get(actualContainer, 'style.paddingBottom') || '0px'
```
ptbrowne: Why is `get` used ? `style` is always defined on DOM nodes right ?

react/SelectBox/SelectBox.jsx
```diff
@@ -48,6 +50,44 @@ const customStyles = props => ({
...
+ * @param  {object} container React reference to an element containing
+ * the SelectBox
+ * @param  {object} element    React reference to the ReactSelect element
+ * @return {object}            A style object, with `menuList` property set.
+ */
+const computedMenuListHeightStyles = (container, selectElement) => {
+  if (!(container && selectElement)) return {}
+
+  const isPreactRef = !container.current
+  const actualContainer = isPreactRef ? container : container.current
```
ptbrowne: Nit: I personally prefer to shadow `container` in those cases. This makes it clear that we are not going to use `container` afterwards. And it makes it simpler to reason about this function since the number of variable is reduced.

```
container = container.current || container
```

Nit: Since this function's responsibility is about computing heights, I would not put it in charge of knowing the difference between preact and react.

ptbrowne:
To test this, I would pass mocked nodes to `computedMenuListHeightStyles` and check the result.

react/SelectBox/SelectBox.jsx
gregorylegarec: It is also bothering me. I will externalize this part, thanks.

gregorylegarec:
Updated with externalisation of Preact/React stuff, and some tests.

ptbrowne:
LGTM

PR #805, opened 2019-01-17
Input: Handle disabled attribute

gregorylegarec:
Base on @claiw 's mockup for harvest apps.

capture d ecran 2019-01-17 a 15 33 57

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -55,6 +55,7 @@ InputPassword.defaultProps = {
 const Field = props => {
   const {
     className,
+    disabled,
     fieldProps,
     fullwidth,
     label,
@@ -78,6 +79,7 @@ const Field = props => {
       case 'textarea':
         return (
           <Textarea
+            disabled={disabled}
             id={id}
             placeholder={placeholder}
             value={value}
@@ -89,6 +91,7 @@ const Field = props => {
       case 'password':
         return (
           <InputPassword
+            disabled={disabled}
             fullwidth={fullwidth}
             id={id}
             type={type}
@@ -108,6 +111,7 @@ const Field = props => {
       case 'text':
         return (
           <Input
+            disabled={disabled}
             fullwidth={fullwidth}
             id={id}
             type={type}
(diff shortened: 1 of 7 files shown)


stylus/components/forms.styl
```diff
@@ -269,6 +269,16 @@ $input-text
     &:invalid
         border rem(1) solid pomegranate
 
+    &[aria-disabled]
```
GoOz:
```suggestion
    &[aria-disabled]
    &[disabled]
```
For uses outside of react components, if there's no `aria-disabled`, which is ok beacuse it's kinda redundant since most screen readers understand the `disabled` attribute just as well.
But I'll keep both, it doesn't cost much and covers all cases.

react/Field/index.jsx
```diff
@@ -78,6 +79,7 @@ const Field = props => {
       case 'textarea':
         return (
           <Textarea
+            disabled={disabled}
```
GoOz: Does this component can handle `disabled` as well?

react/Field/index.jsx
```diff
@@ -89,6 +91,7 @@ const Field = props => {
       case 'password':
         return (
           <InputPassword
```
GoOz: Same question as Textarea.

gregorylegarec: It does.

react/Field/index.jsx
gregorylegarec: It doesn't.

I suggest to remove this line.

stylus/components/forms.styl
gregorylegarec: Thanks !

react/Field/index.jsx
gregorylegarec: Spoiler: I've just removed it.

GoOz: It would have been nice to add this behavior to Textarea component instead of removing the line. 😜
For consistency ✊

gregorylegarec: I agree, it would have been nice. 😏

gregorylegarec: Done.

capture d ecran 2019-01-18 a 12 55 43

GoOz: You're the man. 👍

PR #808, opened 2019-01-18
fix: 🐛 Invalid input show style only on blur

GoOz:
It now shows the red borders for `:invalid` inputs only when your're not focus in it. 

That way it doesn't get red prematurely, like when you're typing the first letters of your email in a mail input and it gets red right away even if you're not finished. That sent the wrong UX message.


ptbrowne:
:shipit:

PR #809, opened 2019-01-18
feat: Add MuiSwitch style to global MUI theme

drazik:
This adds the Switch component to our MUI theme. You can see the result on https://drazik.github.io/cozy-ui/react/#switch.

### DIFF
--- a/react/__snapshots__/examples.spec.jsx.snap
+++ b/react/__snapshots__/examples.spec.jsx.snap
@@ -1010,6 +1010,8 @@ exports[`MuiCozyTheme/ExpansionPanel should render examples: MuiCozyTheme/Expans
 
 exports[`MuiCozyTheme/Menus should render examples: MuiCozyTheme/Menus 1`] = `"<div></div>"`;
 
+exports[`MuiCozyTheme/Switch should render examples: MuiCozyTheme/Switch 1`] = `"<div></div>"`;
+
 exports[`Radio should render examples: Radio 1`] = `
 "<div>
   <form>
--- a/react/MuiCozyTheme/theme.js
+++ b/react/MuiCozyTheme/theme.js
@@ -104,6 +104,48 @@ export const theme = createMuiTheme({
         paddingLeft: 24,
         paddingRight: 24
       }
+    },
+    MuiSwitch: {
+      root: {
+        width: 40,
+        height: 23
+      },
+      switchBase: {
+        width: 25,
+        height: 23
+      },
+      checked: {
+        '& + $bar': {
+          opacity: 1
+        }
+      },
+      icon: {
+        width: 16,
+        height: 16
+      },
+      bar: {
+        width: 25,
+        height: 12,
+        marginTop: -6,
+        marginLeft: -12,
+        backgroundColor: '#d6d8da',
+        opacity: 1
+      },
+      colorPrimary: {
+        '&$checked': {
+          color: 'white'
+        }
+      },
+      colorSecondary: {
+        '&$checked': {
+          color: 'white'
+        }
+      },
+      disabled: {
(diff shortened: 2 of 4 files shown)


react/MuiCozyTheme/Switch/Readme.md
```diff
@@ -0,0 +1,15 @@
...
+const Switch = require('@material-ui/core/Switch').default;
+
+<MuiCozyTheme>
+  <Switch color="primary" defaultChecked />
+  <Switch color="secondary" defaultChecked />
+  <Switch color="primary" disabled />
+  <Switch color="secondary" disabled />
+</MuiCozyTheme>
+
+```
```
ptbrowne: Could you add a small description in front of each `Switch` ? It looks a bit empty when looking at the styleguidist. A link to the original component would be great also :)

```
<pre>color="primary" defaultChecked</pre>
<Switch color="primary" defaultChecked />
```

image

CPatchane:
What is the purpose of this new component? I mean the rendering is very different from our current [toggle component](https://drazik.github.io/cozy-ui/react/#toggle) for the same usage. poke @Claiw @joel-costa

CPatchane:
Ok then but I still see a consistency issue here, we have to different styles of a switch component in our components set. So I have other questions:
Should we really have two different switch style only for a size need? 
Why this can't be an evolution of our current toggle component (like a subtle alternative, we use it for some other cozy-ui components) with maybe the same kind of style?
screenshot 2019-01-18 at 13 56 58
Or maybe we want a new style of switch component and so why this component doesn't replace our current toggle component?

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -1010,6 +1010,8 @@ exports[`MuiCozyTheme/ExpansionPanel should render examples: MuiCozyTheme/Expans
 
 exports[`MuiCozyTheme/Menus should render examples: MuiCozyTheme/Menus 1`] = `"<div></div>"`;
 
+exports[`MuiCozyTheme/Switch should render examples: MuiCozyTheme/Switch 1`] = `"<div></div>"`;
```
CPatchane: Should we consider this as a valid snapshot? I would prefer no tests than an empty snapshot like this which is a false positive (same remark for all MUI snapshots in fact)

drazik:
Just reworked it after a meeting with @joel-costa and @CPatchane:

* `Toggle` now renders a `Switch` from MUI
* The customized `Switch` now have the same dimension as the legacy `Toggle`

This way, there is no breaking change, just new style for the `Toggle` component and it just uses `Switch` under the hood instead of a custom implementation of the switch behaviour.

I just don't understand why tests are failing with this error. Will figure it out.

y-lohse:
@drazik did you ever figure out why the tests are failing? And how have you moved forward in cozy-banks without this component in cozy-ui?

drazik:
No I haven't. We put this component customization in banks (https://github.com/cozy/cozy-banks/blob/master/src/components/Switch.jsx). More than tests, the main question here was: do we make the current `Toggle` component render a MUI `Switch` under the hood? If yes, it means we need to have material-ui as a dependency instead of a dev dependency. And the look of all `Toggles` accross our apps will change, since this component and `Toggle` have the same functionalities, but not the same style at all.

Crash--:
Since we're now using MUI everywhere, maybe we should consider merging this PR ? cc @drazik

drazik:
I must check if the current implementation in banks have differences with this one first.

nono:
@ptbrowne should we close this PR?

ptbrowne:
Yes, thanks, this has been done since then. Hi @drazik !

PR #810, opened 2019-01-18
feat(Toggle): Add subtle variant

drazik:
This adds a `subtle` variant to the `Toggle` component. This variant is needed in Banks for the new panels UI on the "current balance" page.

image

See https://drazik.github.io/cozy-ui/react/#toggle


drazik:
Closing since we decided to go with #809 and keeping backward compatibility by rendering a `Switch` in `Toggle`.

PR #811, opened 2019-01-18
feat: ✨ Button font-size matches its size

GoOz:
Button variants with different sizes have now a matching font-size.

screenshot 2019-01-18 18 55 12

Normal button size -> font-size = 14px
Small button size -> font-size = 13px
Tiny button size -> font-size = 12px
Large button size -> font-size = 16px

Also added utility classes to handle font-size to a given text:
```
 .u-fz-tiny - Sets a tiny font-size (12px)
 .u-fz-xsmall - Sets a xsmall font-size (13px)
 .u-fz-small - Sets a small font-size (14px)
 .u-fz-medium - Sets a medium font-size (16px)
 .u-fz-large - Sets a large font-size (18px)
```

And finally, I fixed text-alignment utility classes by moving them to the right place and renaming them more clearly.

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -819,6 +819,43 @@ Display an chip that represents complex identity
 */
 
 /*
+ Font-size
+
+ Sets various font-size to a text
+
+ Markup:
+ <p class="{{modifier_class}}">Aliquam etiam natoque phasellus habitant interdum a hendrerit ac aliquet</p>
+
+ .u-fz-tiny - Sets a tiny font-size (12px)
+ .u-fz-xsmall - Sets a xsmall font-size (13px)
+ .u-fz-small - Sets a small font-size (14px)
+ .u-fz-medium - Sets a medium font-size (16px)
+ .u-fz-large - Sets a large font-size (18px)
+
+ Weight: 6
+
+ Styleguide utilities.text.fontsize
+*/
+
+/*
+ Text alignement
+
+ Sets a text alignement to a text
+
+ Markup:
+ <p class="{{modifier_class}}">Facilisis eget nam turpis id adipiscing facilisi odio ut fames, aliquet laoreet felis habitasse tincidunt litora leo senectus.</p>
+
+ .u-ta-left - Sets a left alignment
+ .u-ta-right - Sets a right alignment
+ .u-ta-center - Sets a center alignment
+ .u-ta-justify - Sets a justify alignment
+
+ Weight: 7
+
+ Styleguide utilities.text.alignment
+*/
+
+/*
  Ellipsis
 
  Adds ellipsis to a long text
@@ -829,7 +866,7 @@ Display an chip that represents complex identity
 
  .u-ellipsis - Ellipsis at the end
 
- Weight: 6
+ Weight: 8
 
  Styleguide utilities.text.ellipsis
 */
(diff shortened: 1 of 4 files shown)


stylus/cozy-ui/build.styl
```diff
@@ -864,10 +901,6 @@ Display an chip that represents complex identity
  <h4 class="u-title-h4">Heading 4</h4>
  <p class="u-text">Basic text</p>
  <p class="u-caption">Caption text</p>
- <p class="u-t-left">Left aligned text</p>
- <p class="u-t-right">Right aligned text</p>
- <p class="u-t-center">Center aligned text</p>
- <p class="u-t-justify">Justify aligned text</p>
```
Crash--: are you removing theses classes?

Crash--: Are we sure they are not used within an app?

GoOz: Moving only, not removing. it wasn't in the right place.

Crash--: You renamed `u-t-justify` to `u-ta-justify`, no?

GoOz: Ah yes, you meant that. Indeed. I don't think they were used since they weren't exposed globally before. But I'll look around. It's not used in UI though.

GoOz: Confirmed, it's only used on the onboarding infra, which is the project i'm working on right now.

GoOz:
[(𝐒) [ONB] Amélioration de choix de l'adresse mail](https://trello.com/c/6weCwMuy/367-%F0%9D%90%92-onb-am%C3%A9lioration-de-choix-de-ladresse-mail)

stylus/utilities/text.styl
```diff
@@ -61,10 +61,71 @@ $midellipsis
         > :first-child
             text-overflow '[...]'
 
+// @stylint off
+fz-tiny =
+    font-size rem(12)!important
+    line-height 1.3!important
+$fz-tiny
+    {fz-tiny}
```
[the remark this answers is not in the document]
Crash--: We just change `t-left` to `ta-left` for `text-align`, so having `fz` for `font-size` makes sense.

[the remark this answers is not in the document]
GoOz: Sorry about that, it is obvious to me but you're absolutely right, it is very important to explain the reasons why I chose "fz"  instead of anything else.

I made my choice of syntax based on the very popular et widely used text editor plugin [Emmet.io](https://emmet.io/) by numerous CSS developers because:
- It is actually a common vocabulary for a lot of people;
- It has abbreviations for every CSS properties figured out (I didn't check but I guess if not all then most of it), so it is a reliable reference for naming future utility classes and duplicate free (they really think it through);

I could have used 'fs' but then it would have been difficult to differentiate `font-size` from `font-style`.  Emmet chose `fz` for the size and & `fs` for the style and I figured why reinventing the wheel on that matter, that would be time & brain poorly spent.

Everybody hates naming things when coding. With this, nobody has to do it anymore, for utility classes anyway.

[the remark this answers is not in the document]
drazik: I agree with @GoOz. `fz` is like a standard shortcut for `font-size` thanks to Emmet. I'd go for it too.

PR #812, opened 2019-01-21
feat(Field): Extra field , actually just side label

sizvix:
Add the possibility to have an text ( or button in the future ) item on the right of the label.

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -2,9 +2,9 @@ import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 
+import { default as labelStyles } from '../Label/styles.styl'
 import styles from './styles.styl'
 import Label from '../Label'
-import { default as LabelStyles } from '../Label/styles.styl'
 import Input from '../Input'
 import SelectBox from '../SelectBox'
 import Textarea from '../Textarea'
@@ -26,7 +26,7 @@ class InputPassword extends React.Component {
         {showVisibilityButton && (
           <div
             className={cx(
-              LabelStyles['c-label'],
+              labelStyles['c-label'],
               styles['o-field-input-action']
             )}
             onClick={() => this.toggleVisibility()}
@@ -67,6 +67,7 @@ const Field = props => {
     onChange,
     readOnly,
     secondaryLabels,
+    side,
     size
   } = props
 
@@ -133,6 +134,11 @@ const Field = props => {
   return (
     <div className={cx(styles['o-field'], className)}>
       <Label htmlFor={id}>{label}</Label>
+      {side && (
+        <div className={cx(styles['o-side'], labelStyles['c-label'])}>
+          {side}
+        </div>
+      )}
       {inputType(type)}
     </div>
   )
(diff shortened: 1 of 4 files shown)


sizvix:
[🚨 (𝐗𝐗𝐋) [APR] Afficher le formulaire](https://trello.com/c/dr7OlUjS/548-%F0%9F%9A%A8-%F0%9D%90%97%F0%9D%90%97%F0%9D%90%8B-apr-afficher-le-formulaire)

ptbrowne:
Thanks sizvix !

1. Can you publish the styleguide on your repository to show the results ? 

```
yarn build:doc:react
yarn deploy:doc --repo git@github.com:sizvix/cozy-ui.git
```


2. If it's "actually just a side label", could this property be named `sideLabel` ? Or extra information ? What is the use case here ? Provide additional information to the user ? "additionalInfo" ?
3. No need to default the property to the empty string, undefined is OK.

ptbrowne:
It seems to me that this does not belong at the moment in Cozy-UI. Since it is so fresh, I think it should have more rounds of iteration inside the app before being put in cozy-ui. I feel like we are adding props for every single need instead of having composable APIs.

gregorylegarec:
@ptbrowne Actually, as stated with @Crash-- at the beginning of the harvest apps, we do not want to put any UI component into Cozy-Harvest-Lib. So anything we need is done first in Cozy-UI.

I don't think this change is breaking or could cause any trouble. It also improve our API for future needs.

So if your ok with @sizvix changes, I would like to merge this one. Do you agree ?

gregorylegarec:
@ptbrowne Thanks for those precisions 👍

react/Field/index.jsx
```diff
@@ -133,6 +134,17 @@ const Field = props => {
   return (
     <div className={cx(styles['o-field'], className)}>
       <Label htmlFor={id}>{label}</Label>
+      {sideLabel && (
+        <div
+          className={cx(
+            styles['o-side-label'],
+            LabelStyles['c-label'],
```
gregorylegarec: `LabelStyles` should start with a lowercase.

react/Field/index.jsx
```diff
@@ -133,6 +134,17 @@ const Field = props => {
...
+            styles['o-field-input-action']
```
gregorylegarec: It think that we should keep semantic here, as we are not displaying an _input action_ element, as we do for the password, we should not use a class named like that.

Instead, I propose to use `o-field-input-action` CSS properties into `o-side-label`, with a mixin or something.

What @GoOz says about this ?

react/Field/styles.styl
```diff
@@ -12,3 +12,8 @@
     top -3rem
     right 0
     cursor pointer
+
+.o-side
```
gregorylegarec: Last feedback: Maybe semantically it could be better to have only an `o-side` class. What do you think ?  @ptbrowne ? @GoOz ?

sizvix: ok to switch and place o-side on the main class and o-field-input-action extends this

PR #814, opened 2019-01-22
Alerter improvements

CPatchane:
* Provide a method to `buttonAction` to dismiss the alerter before the duration ends
* Fix background button issue
* Fix and update the doc

Example https://cpatchane.github.io/cozy-ui/react/#alerter

### DIFF
--- a/docs/webpack.config.js
+++ b/docs/webpack.config.js
@@ -26,7 +26,7 @@ module.exports = {
           {
             loader: 'stylus-loader',
             options: {
-              use: [ require('../stylus')() ]
+              use: [require('../stylus')()]
             }
           }
         ]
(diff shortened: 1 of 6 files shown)


docs/webpack.config.js
```diff
@@ -26,14 +26,19 @@ module.exports = {
...
         ]
       }
     ]
   },
   plugins: [
-    new MiniCssExtractPlugin('[name].css')
+    new MiniCssExtractPlugin('[name].css'),
+    new webpack.DefinePlugin({
+      'process.env': {
+        USE_REACT: 'true'
```
ptbrowne: Is it intended ? It means that we use react now in the styleguidist right ?

ptbrowne:
When you hover over the dismiss button, the background disappears, is it intended ? 

image

docs/webpack.config.js
CPatchane: The doc was not working correctly with portals since styleguidist uses React but without this variable it was using `preact-portal` for portals :/

CPatchane:
@ptbrowne I fixed the button issue and added some tests to the alerter 👍

PR #815, opened 2019-01-22
[select-box] Do not wrap

ptbrowne:
- Do not wrap select. A ref to the control is accessible via `ref.select.controlRef`
- Try not to do DOM operations while rendering.

View on : https://ptbrowne.github.io/cozy-ui/react/#!/SelectBox/5

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -222,6 +222,10 @@ const getNodeFromRef = ref => {
   return (ref && ref.current) || ref
 }
 
+const customComponents = {
+  DropdownIndicator,
+  Option
+}
 
 class SelectBox extends Component {
   constructor(props) {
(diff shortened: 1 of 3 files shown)


nono:
Ping @GoOz @CPatchane

react/SelectBox/SelectBox.jsx
```diff
@@ -216,22 +216,57 @@ ActionsOption.defaultProps = {
...
+  refreshMenuStyle() {
+    if (this.props.container && this.controlRef) {
+      this.menuStyle = computedMenuListHeightStyles(
+        getNodeFromRef(this.props.container),
+        this.controlRef
+      )
+    }
   }
 
-  handleClose = () => {
```
CPatchane: This way to write is not handled in cozy-ui?

ptbrowne: It is but it is harder to write tests and the lead dev of enzyme gives advice against it. More info on https://github.com/cozy/cozy-guidelines/pull/58

PR #817, opened 2019-01-22
Fix/wizard glitches

GoOz:
- Fixed some design issues
- Added progress bar on creation
- Added text button

### DIFF
--- a/stylus/components/wizard.styl
+++ b/stylus/components/wizard.styl
@@ -25,35 +25,28 @@ $wizard--waiting
 
 $wizard--scroll
     position absolute
+    padding-top rem(56)
 
     +medium-screen()
-        padding-top rem(80)
         box-sizing border-box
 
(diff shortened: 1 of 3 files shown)


stylus/components/wizard.styl
```diff
@@ -25,35 +25,28 @@ $wizard--waiting
...
 $wizard--scroll
     position absolute
+    padding-top rem(56)
 
     +medium-screen()
-        padding-top rem(80)
         box-sizing border-box
 
-    +tiny-screen()
+    @media (max-width: 33.9375rem), (max-height: 23.4375rem)
```
Crash--: What is the real target with this media query ? 

Since we're using it multiples times, do we need to create a new utility?

Crash--: Yep, sorry just a mixin. no need to add it to our generic collection.

PR #819, opened 2019-01-23
fix(field): Right position different with & whithout fullwidth

sizvix:
The position of the side element should be align on the right of the input.
With or without fullwidth option.

### DIFF
--- a/react/Field/Readme.md
+++ b/react/Field/Readme.md
@@ -61,7 +61,7 @@ const options = [
 </form>
 ```
 
-##### Password field without show/hide button
+#### Password field without show/hide button
 
 ```
 <form>
@@ -74,7 +74,7 @@ const options = [
 </form>
 ```
 
-##### Password field without show/hide button
+#### Side element
 
 ```
 <form>
@@ -84,3 +84,16 @@ const options = [
   />
 </form>
 ```
+
+#### Side element with fullwidth element
+
+```
+<form>
+  <Field
+    label="I'm a label"
+    side="(optional)"
+    fullwidth={true}
+  />
+</form>
+```
+
--- a/react/Field/styles.styl
+++ b/react/Field/styles.styl
@@ -10,9 +10,15 @@
 .o-side
     position absolute
     top 0
-    right 0
+    left 0
+    width 100%
+    max-width rem(512)
+    text-align right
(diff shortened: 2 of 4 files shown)


react/Field/Readme.md
```diff
@@ -84,3 +84,16 @@ const options = [
   />
 </form>
 ```
+
+#### Side element with fullwidth elemeent
```
gregorylegarec:
```suggestion
#### Side element with fullwidth element
```

react/Field/styles.styl
```diff
@@ -10,10 +10,16 @@
 .o-side
     position absolute
     top 0
```
gregorylegarec: `top 0` and `left 0` are default values, so I don't think we need to precise them.

gregorylegarec:
@GoOz This solution implies to introduce a new wrapper and introduce potential breaking changes, does it ?

GoOz:
My solution implies no change at all, just using what's already at our disposal. There's no need for another wrapper either.

react/Field/styles.styl
GoOz: That's actually not true. :)
Default `top` position value is the one the element has in the flow, while the `left` one depends on the direction set in the page (sure we use ltr, but still that's not always true)
https://codepen.io/GoOz/pen/Jxogax

my 2 cents.

GoOz:
Ok my mistake, I understood the problem the wrong way. I don't see a better solution so far.

ptbrowne:
Can you upload the styleguide ? It is requested in the PR template for a reason :/

gregorylegarec:
@sizvix Any news ? I am a little bit lost 😅

sizvix:
@ptbrowne @CPatchane , the styleguide is updated   ^^U
Is it OK for you ?

ptbrowne:
Next time, if you can link to the concerned component, it would be nice :) Less work for the reviewer, more chance for the PR to be accepted.

LGTM 👍 

It works for me now, but I am not really happy with the chosen design, if the side element is long or the container is small, the side element can go over the label :/

image

nono:
@gregorylegarec ping can we merge this PR?

nono:
Ping @gregorylegarec

PR #822, opened 2019-01-25
fix: Layout issue on iPhone X

Crash--:
This fix our layout issue with Drive on an iPhone X. 

See https://trello.com/c/IY5dJyDv/367-iphone-x-adaptations

### DIFF
--- a/stylus/objects/layouts.styl
+++ b/stylus/objects/layouts.styl
@@ -96,6 +96,7 @@ $app
         main
             padding-left env(safe-area-inset-left)
             padding-right env(safe-area-inset-right)
+            margin-bottom env(safe-area-inset-bottom)
 
         main,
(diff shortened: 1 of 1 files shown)


Crash--:
I followed this article : https://blog.phonegap.com/displaying-a-phonegap-app-correctly-on-the-iphone-x-c4a85664c493 

You already have done a lot of works. Seems that this one was missing. 

I think we'll have some other changes to make in a near future, but we need to talk first ;)

ptbrowne:
Naive question: is there a particular reason why this is not a `padding-bottom` like the other properties ?

LGTM.

nono:
@Crash-- can we merge this PR?

nono:
@Crash-- friendly reminder about this PR

Issue #666, opened 2018-10-30
Improve ActionModal


joel-costa:
Our links in action modales on mobile have currently a touch-zone really tight. We have areas with 34px height while Android recommends 48px and iOS 44px to optimize the touch comfort.

Could we increase touch-zones to set height to 48px and increase the margin to the left and the right of the icons to align them with the header : https://gyazo.com/51956a9d0a377b8d92a87f11b1e829cc

- [x] ActionMenuItem

Height: 48px
Can accept an Icon

- [ ] ActionMenuHeader

Header text should be colored in #charcoal-grey #32363f

https://github.com/cozy/cozy-ui/commit/d981fd23ce8458d5b92e3f67db15849ea1941581\

[Zeplin example](https://app.zeplin.io/project/584a6f4e838090371f59518f/screen/5abb56ec8db5855f9f817bfa)

PR #828, opened 2019-02-04
feat(ActionMenu): ✨ Add ActionMenuHeader component

GoOz:
fixes: #666

It accepts any children.

[Preview](https://gooz.github.io/cozy-ui/react/#actionmenu)

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -6,6 +6,8 @@ import styles from './styles.styl'
 import Overlay from '../Overlay'
 import once from 'lodash/once'
 import { Media, Bd, Img } from '../Media'
+import Icon from '../Icon'
+import { Bold } from '../Text'
 
 const TRANSITION_DURATION = 100 // need to be kept in sync with css
 
@@ -141,10 +143,7 @@ class ActionMenu extends Component {
     const { children, className } = this.props
     const { closing } = this.state
     return (
-      <div
-        className={cx(styles.ActionMenu, className)}
-        ref={this.handleWrapperRef}
-      >
+      <div className={className} ref={this.handleWrapperRef}>
         <Overlay
           style={{ opacity: closing ? 0 : 1 }}
           onClick={this.animateClose}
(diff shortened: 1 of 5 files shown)


react/ActionMenu/index.jsx
```diff
@@ -166,9 +165,47 @@ ActionMenu.propTypes = {
...
+          ellipsis
+          className={styles['c-actionmenu-header-filename']}
+        >
+          {filename}
+        </Bold>
+        {extension && (
+          <Bold tag="span" className="u-coolGrey">
+            {extension}
+          </Bold>
+        )}
```
ptbrowne: It seems it is more a FileActionMenuHeader than a generic ActionMenuHeader. Can you extract the file specific properties so that ActionMenuHeader is generic and can serve other applications than Drive ?

stylus/components/action-menu.styl
```diff
@@ -3,25 +3,43 @@
...
+        margin-top 0
+
+$actionmenu-header
+    box-sizing border-box
+    border-bottom rem(1) solid silver
+    padding rem(16)
+    min-height rem(64)
+
+$actionmenu-header-icon
+    flex-shrink 0
```
ptbrowne: Is it not handled by the `Media:Img` ?

GoOz: Oh yeah right, I misread. Thought `Img` component was an `img` HTML tag.
Will fix this.

react/ActionMenu/index.jsx
GoOz: I don't understand what you mean.

ptbrowne: I think `ActionMenuHeader` can be useful for other things than `File`.  For example in Banks, it could be useful to show the title of a transaction.

The way it is implemented now, it is very coupled to files (since it has `extension`, `filename` etc..). So for me, there are two solutions. Either rename the component to `FileActionMenuHeader` or make the `ActionMenuHeader` more generic and use it  to implement a FileActionMenuHeader than can be useful for Drive.

PR #830, opened 2019-02-05
feat: Add Filename component

GoOz:
I made a separate component to deal with filename with extension and icon as seen in the [ActionMenuHeader](https://gooz.github.io/cozy-ui/react/#actionmenu).

ActionMenuHeader will be more generic and accept any type of children.


GoOz:
[(𝐌) Mettre à jour le composant ActionMenu](https://trello.com/c/LIQ7A5Ek/606-%F0%9D%90%8C-mettre-%C3%A0-jour-le-composant-actionmenu)

PR #834, opened 2019-02-07
feat: Add disabled checkbox & radio style

GoOz:
Added props & style for disabled Checkbox & Radio.

Preview: [Checkbox](https://gooz.github.io/cozy-ui/react/#checkbox) & [Radio](https://gooz.github.io/cozy-ui/react/#radio)

### DIFF
--- a/react/Checkbox/Readme.md
+++ b/react/Checkbox/Readme.md
@@ -12,6 +12,12 @@
 <div><Checkbox label="This is a checkbox with an error" error /></div>
 ```
 
+### Checkbox when disabled
+
+```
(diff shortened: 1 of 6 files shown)


y-lohse:
In the demo, when I hover the standard checkbox and radio buttons, I have the `disabled` cursor in chrome, so I think something is wrong?

GoOz:
@y-lohse oh right o_o How could I not see this. `disabled` attribute don't take `false` as a valid value so every checkbox is disabled by default. Gonna fix this, right now.

GoOz:
Fixed!

PR #836, opened 2019-02-11
refactor: Refactored style for Empty view

GoOz:
So I removed the fixed positionning and added a `margin-top` to the Empty element according to the viewport height (minus the height of the cozy-bar and the navigation to be a little more precise), which basically should render exactly the same as before but without overlapping any other element so no more "pointer-events" issue.

The fix in Drive about that would be removable but won't break things if it stays (which it shouldn't obviously) so you have time to clean this up.

Also reduced the `margin-bottom` of the icon as requested by the designers.

### DIFF
--- a/stylus/components/empty.styl
+++ b/stylus/components/empty.styl
@@ -1,6 +1,9 @@
 @require '../settings/palette'
 @require '../settings/breakpoints'
 @require '../tools/mixins'
+@require '../objects/layouts'
+
+contentHeight = barHeight + navHeight
 
 $empty
     display          flex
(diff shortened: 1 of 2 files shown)


GoOz:
[(XS) composant empty : modifs mineures](https://trello.com/c/dCvKmaLU/701-xs-composant-empty-modifs-mineures)

GoOz:
[(M) Améliorer nos composants pour les vues vides](https://trello.com/c/4V0z4bh4/459-m-am%C3%A9liorer-nos-composants-pour-les-vues-vides)

stylus/components/empty.styl
```diff
@@ -1,6 +1,9 @@
 @require '../settings/palette'
 @require '../settings/breakpoints'
 @require '../tools/mixins'
+@require '../objects/layouts'
+
+contentHeight = barHeight * 2
```
ptbrowne: IMHO It is a bit unclear that it comes from the nav and the bar heights combined (and it is a bit of a coincidence that the two bars have the same height). I think it would be clearer to have `contentHeight = barHeight + navHeight` and have navHeight also defined in `objects/layouts`.

GoOz: It's not a coincidence per se, but I get your point, in case things change, even in other context, it'll be ready. Gonna change that.

Issue #838, opened 2019-02-12
global() does not work as expected in webpack builds


ptbrowne:
## Expectation

Utility classes and other classes that are not supposed to be scoped via CSS modules should not have a hash and thus should be de-duplicated via `postcss-discard-duplicates`.

## Problem

Utility classes that are supposed to be `:global` are not `:global`, thus a hash is added when outputted via CSS modules and they are not de-duplicated via postcss-discard-duplicates, resulting in a larger CSS.

## Explanation

We use the `global` mixin (custom mixin) to indicate that a class should be global.

https://github.com/cozy/cozy-ui/blob/8b60fd17f9664abcc16cc688cad6357818cc9dd1/stylus/utilities/text.styl#L29-L33

This allows us to work whether CSS modules are activated. It will wrap its content in `:global` if CSS modules are used and will not wrap otherwise.

https://github.com/cozy/cozy-ui/blob/e62458391874bc34fdffbef5356a7a770dcba7ea/stylus/tools/mixins.styl#L90-L96

CSS modules are known to be used if the `cssmodules` stylus variable is set. This variable is only set in `index.styl`.

https://github.com/cozy/cozy-ui/blob/8b60fd17f9664abcc16cc688cad6357818cc9dd1/stylus/cozy-ui/index.styl#L1

It means that any time the stylus loader loads a stylus file that do not happen to require `index.styl` (directly or indirectly via its dependencies), the `cssmodules` variable is not set.

## Possible solution

The stylus variable `cssmodules` should be set automatically. One possible entry point is the stylus plugin and the [define](http://stylus-lang.com/docs/js.html#definename-node) method.

https://github.com/cozy/cozy-ui/blob/master/stylus/index.js

ptbrowne:
FIxed by #839

PR #839, opened 2019-02-12
Optimize transpiled css sheet

CPatchane:
* Use postcss + cssnano to optimize `stylesheet.css`
* `transpiled/stylesheet.css` -> `transpiled/react/stylesheet.css`: it will avoid a breaking change when we will use directly the `react/` folder
* Use a `babel.config.js` instead of `.babelrc` + upgrade babel deps
* Stop using random hash for css modules but an id based on cozy-ui version

Optimisation result:
**Before**
before
**After**
after

Docs: http://cpatchane.github.io/cozy-ui/react


package.json
```diff
@@ -40,31 +39,27 @@
     "jest": "yarn test:jest",
     "test": "yarn test:jest",
     "test:jest": "env BABEL_ENV=transpilation jest --verbose --coverage",
-    "transpile": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --copy-files --ignore '**/test/**,**/__mocks__/**,**/*.spec.*'",
+    "transpile": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --copy-files --ignore '**/test/**,**/__mocks__/**,**/*.spec.*' --verbose",
```
ptbrowne: Ca permet de voir quoi en plus le --verbose ?

package.json
```diff
@@ -22,7 +22,6 @@
     "build:doc:react": "styleguidist build --config docs/styleguide.config.js",
     "clean:doc:kss": "rm -rf build/styleguide",
     "commitmsg": "commitlint -e $GIT_PARAMS",
-    "copy-files": "scripts/copy-files.sh",
```
ptbrowne: Inexisting script. Good catch.

package.json
CPatchane: The last version of babel doesn't show transpiled components last before, we have to add the `--verbose` option

babel.config.js
```diff
@@ -0,0 +1,23 @@
...
+        [
+          'css-modules-transform',
+          {
+            extensions: ['.styl'],
+            preprocessCss: './preprocess',
+            extractCss: './transpiled/react/stylesheet.css',
+            generateScopedName: `[name]__[local]___cozy_ui_${version.replace(
+              '.',
+              '_'
+            )}`
```
ptbrowne: Qu'est ce que ca apporte d'avoir la version de `cozy-ui` et d'enlever le hash ? J'ai peut être pas compris le but.

docs/README.md
```diff
@@ -32,7 +32,7 @@ If you use the transpiled components (from `cozy-ui/transpiled/react`), you need
 
 ```
 import Button from 'cozy-ui/transpiled/react/Button'
-import 'cozy-ui/transpiled/stylesheet.css'
```
ptbrowne: IMHO this commit should be merged with f9308afd36faca5d7a7d0597c23feb8a72354f20

babel.config.js
```diff
@@ -0,0 +1,23 @@
...
+  presets: ['cozy-app'],
+  env: {
+    transpilation: {
+      plugins: [
+        [
+          'css-modules-transform',
+          {
+            extensions: ['.styl'],
+            preprocessCss: './preprocess',
+            extractCss: './transpiled/react/stylesheet.css',
```
ptbrowne: Pourquoi le mettre dans `transpiled/react/stylesheet.css`, ca pourrait ne pas être un breaking change non ?

babel.config.js
CPatchane: With the hash we will have different classes at each imports, like:
```
      .u-coolGrey {
      .u-coolGrey--12da1 {
      .u-coolGrey--1U0_S {
      .u-coolGrey--2flc7 {
       .u-coolGrey--2G1pW {
      .u-coolGrey--2r5Fs {
      .u-coolGrey--2VhFP {
      .u-coolGrey--2vQlR {
      .u-coolGrey--2xPIP {
      .u-coolGrey--3HU7I {
      .u-coolGrey--758MC {
      .u-coolGrey--jkmeV {
      .u-coolGrey--MlX_r {
      .u-coolGrey--Sr1MO {
```
Using a know id here will allow postcss to optimize and remove duplicates correctly

babel.config.js
CPatchane: Putting the stylesheet file at the root of `transpiled/` forces all applications to import `cozy-ui/transpiled/stylesheet.css`, so they can't use the alias `'cozy-ui/react': 'cozy-ui/transpiled/react'` and when we will put back transpiled components into `react/` we will push a breaking change for all apps using it.
To avoid a breaking change right now we could copy the file into `transpiled/stylesheet.css` but since it's used only in Banks and Drive I think we can handle it using a new major version imho

babel.config.js
ptbrowne: The usual way not to add a hash is to use `:global` and we already use it via the `global()` mixin. BUT, we have a bug in the usage of `global()` that deactivates the wrapping in `:global`. See https://github.com/cozy/cozy-ui/issues/838. I would prefer if we'd solve this problem. I do not think removing the hash is a good way to solve the problem since it does not solve the root of the problem. WDYT ?

CPatchane: The last commits fix this to use correctly css-modules with stylus, the result is finally even better than using this fixed ID 👍

PR #840, opened 2019-02-12
chore: add 'others' props to some components

CozyKourai:
As discuss in mattermost, I'm adding 'others' props to some components in order to pass 'data-test-id' to use as selector in testcafé E2E testing in cozy-drive


Crash--:
> Also, naive question but, is it safe to use IDs for testing, since IDs are supposed to be unique, wouldn't it be better to use data-attributes instead?

We don't use `id`, we use `data-test-id`, but we decided it was better to write `id` in the exemple

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -235,6 +235,10 @@ exports[`Checkbox should render examples: Checkbox 5`] = `
 </div>"
 `;
 
+exports[`Empty should render examples: Empty 1`] = `"<div></div>"`;
+
+exports[`Empty should render examples: Empty 2`] = `"<div></div>"`;
```
Crash--: It doesn't work here. We should have <div id='empty' non ?

react/Empty/Readme.md
```diff
@@ -31,10 +31,9 @@ const styles = {
   }
 };
 
-<div style={styles.empty}>
+<div style={styles.empty} id='empty'>
```
Crash--: OK, this is why we don't have the id in the snapshot. 

We should not attach an id to this div but to the `Empty` component below ;)

react/Empty/index.jsx
```diff
@@ -5,9 +5,16 @@ import { Text, MainTitle } from '../Text'
...
+  icon,
+  title,
+  text,
+  children,
+  className,
+  ...others
+}) => {
   return (
-    <div className={cx(styles['c-empty'], className)}>
+    <div className={cx(styles['c-empty'], className)} {...others}>
```
CPatchane: We have to be careful, with React, we can have a lot of warning by passing non-dom attributes to a react component, what about a specific props named `htmlAttributes` containing all attributes we want to pass to the component? And we would be able to expose that way to all other components

Crash--: We had already had this discussion and we agreed to go this way as other UI libs are doing the same. 

The App has to be careful yes. Not the lib ;).

Crash--: See https://github.com/cozy/cozy-ui/issues/821

react/Empty/index.jsx
```diff
@@ -33,7 +40,11 @@ Empty.propTypes = {
   icon: iconPropType.isRequired,
   title: PropTypes.node.isRequired,
   text: PropTypes.node,
-  className: PropTypes.string
+  className: PropTypes.string,
+  /*
+   other = all html attributes
+   */
+  others: PropTypes.object
```
ptbrowne: `others` is not really a prop, you can remove it from here :)

react/Menu/index.jsx
```diff
@@ -167,7 +169,9 @@ Menu.propTypes = {
   /** Global Styles for MenuItems */
   itemsStyle: PropTypes.object,
   /** if you need fixed menu */
-  popover: PropTypes.bool
+  popover: PropTypes.bool,
+  /** other = all html attributes **/
+  restProps: PropTypes.object
```
ptbrowne: from here also

react/Menu/index.jsx
```diff
@@ -97,13 +97,15 @@ class Menu extends Component {
...
-      popover
+      popover,
+      ...restProps
     } = this.props
     const { opened } = this.state
     const Tag = popover === true ? Popover : 'div'
 
     return (
       <div
+        {...restProps}
```
ptbrowne: `{...restProps}` should be put at the end of the tag to be able to override any default behavior IMHO.

react/Nav/index.jsx
```diff
@@ -3,8 +3,10 @@ import Icon from '../Icon'
 import styles from './styles.styl'
 import cx from 'classnames'
 
-export const NavItem = ({ children }) => (
-  <li className={styles['c-nav-item']}>{children}</li>
+export const NavItem = ({ children, ...others }) => (
```
ptbrowne:
```suggestion
export const NavItem = ({ children, ...restProps }) => (
```

react/Nav/index.jsx
```diff
@@ -3,8 +3,10 @@ import Icon from '../Icon'
...
+  <li className={styles['c-nav-item']} {...others}>
```
ptbrowne:
```suggestion
  <li className={styles['c-nav-item']} {...restProps}>
```

react/Sidebar/index.jsx
```diff
@@ -3,13 +3,19 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
 
-const Sidebar = ({ children, className }) => (
-  <aside className={cx(styles['o-sidebar'], className)}>{children}</aside>
+const Sidebar = ({ children, className, ...others }) => (
```
ptbrowne:
```suggestion
const Sidebar = ({ children, className, ...restProps }) => (
```

react/Sidebar/index.jsx
```diff
@@ -3,13 +3,19 @@ import cx from 'classnames'
...
+  <aside className={cx(styles['o-sidebar'], className)} {...others}>
```
ptbrowne:
```suggestion
  <aside className={cx(styles['o-sidebar'], className)} {...restProps}>
```

react/Sidebar/index.jsx
```diff
@@ -3,13 +3,19 @@ import cx from 'classnames'
...
 )
 
 Sidebar.propTypes = {
   children: PropTypes.node,
-  className: PropTypes.string
+  className: PropTypes.string,
+  /*
+   other = all html attributes
+   */
+  others: PropTypes.object
```
ptbrowne: Can be removed, it is not really a prop

react/Empty/Readme.md
Crash--: Not resolved...

react/Menu/Readme.md
```diff
@@ -13,7 +13,7 @@ const Button = require('../Button').default;
 const showItem = itemData => alert(JSON.stringify(itemData));
 const showWarning = itemData => alert(itemData + ' is disabled');
 
-<Menu text='Click me !' onSelect={ showItem } onSelectDisabled={ showWarning }>
+<Menu id='Menu' text='Click me !' onSelect={ showItem } onSelectDisabled={ showWarning }>
```
Crash--:
```suggestion
<Menu id='menu' text='Click me !' onSelect={ showItem } onSelectDisabled={ showWarning }>
```

PR #843, opened 2019-02-20
fix(wizard): Previous button position

GoOz:
Wizard previous button was misplaced due to a previous fix


stylus/components/wizard.styl
```diff
@@ -250,7 +247,7 @@ $wizard-header-fixed
     padding-left rem(32)
 
 $wizard-previous
-    margin 0
+    margin 0 0 0 -2rem
```
Crash--: -2rem ? Where this value comes from?

PR #844, opened 2019-02-20
fix(wizard): Dualfield select height

GoOz:
Select in dual field was too big. It caused a visual unbalance

### DIFF
--- a/stylus/components/wizard.styl
+++ b/stylus/components/wizard.styl
@@ -463,7 +463,7 @@ $wizard-select
     margin rem(2)
     width rem(148)
     border 0
-    padding rem(11 38 11 8)
+    padding rem(10 38 10 8)
 
     &:hover
     &:focus
@@ -475,7 +475,7 @@ $wizard-select
 
 $wizard-select--medium
     @extend $select--medium
-    padding rem(6 38 8 8)
+    padding rem(5 38 7 8)
 
 $wizard-select--narrow
     width rem(40)


GoOz:
This is a change for a `select` inside a dual field so basically only for the wizard for now.

This may be a good candidate for a react component but since it drastically changed 3 times since the beginning of the onboarding's renewal, that I don't have clear specs and that this is clearly a tricky one. I'm not making it an extensible react component with clean styles just yet.
It's not meant to be used outside the wizard for the moment because most of the onboarding is not powered by React components and the onboarding is still under developement.

As for the "magic numbers", the select use the default select styles **but** unlike any other form elements it is smaller (for a visual effect on hover) making it an exception concerning its own padding. I don't know how to not use magic numbers when I wanna trade 2px of top and bottom padding to 2px of top and bottom margin. I have to override its default paddings somehow to make it work.

I don't get what you're expecting, really. The default paddings values of the select are not variables I just can't go with something like:

```styl
selectMargin = rem(2)

$wizard-select
   padding-top defaultSelectPaddingTop - selectMargin
   padding-bottom defaultSelectPaddingBottom - selectMargin
```

Not that it wouldn't be a cool thing to do but if we want to go that way there is some serious thinking of architecture to do first and this PR is just about a fix because it had a wrong calculation.

GoOz:
[(𝐒) [ONB] MEP du nouvel onboarding mobile](https://trello.com/c/DCThDhrq/1003-%F0%9D%90%92-onb-mep-du-nouvel-onboarding-mobile)

ptbrowne:
Suggestion, not blocking : I think I would be less unesasy if the numbers were replaced by variables with names expliciting their function.

Since this fixes a problem, LGTM for me :)

PR #845, opened 2019-02-20
fix(wizard): Fluid fixed header

GoOz:
Wizard's fixed header is a little bit smarter.
Margins are handled by the children elements instead of the parent.


GoOz:
[(S) [ONB] MEP du nouveau login](https://trello.com/c/4lLW4oow/1006-s-onb-mep-du-nouveau-login)

PR #848, opened 2019-02-22
Field: Expose autocomplete

gregorylegarec:
To manage autocompletion behavior in cozy-harvest-lib, we
need to be able to specify the `autocomplete` attribute of
`<input />` elements.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -55,6 +55,7 @@ InputPassword.defaultProps = {
 
 const Field = props => {
   const {
+    autoComplete,
     className,
     disabled,
     fieldProps,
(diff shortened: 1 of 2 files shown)


ptbrowne:
The HTML attribute is in lowercase (no capital C), I think we should keep it that way.

Crash--:
We need to use the Capital with React https://reactjs.org/docs/dom-elements.html

gregorylegarec:
@ptbrowne Also my bad, I faced the issue, eslint corrected me and then I forgot to paste the explanation in the PR description.

Issue #618, opened 2018-09-26
Sentry issues : cozy-ui/react/ActionMenu/index in options errorCannot read property 'addEventListener' of null


Crash--:
Currently taking a look at our Sentry issues and this one appears & affect a lot of users. I don't know exactly if it breaks something or not but if I can clean up our logs... ;)

```
TypeError
Cannot read property 'addEventListener' of null
./node_modules/cozy-ui/react/ActionMenu/index.jsx in options at line 115:18
    const close = once(() => {
      this.menuNode.removeEventListener('transitionend', close)
      this.close()
    })
    this.menuNode.addEventListener('transitionend', close, false)
    // in case transitionend is not called, for example if the element is removed
    setTimeout(close, TRANSITION_DURATION)
    this.applyTransformation(1.1)
  }
```

I don't know how it can happen, but I can add a simple check before adding the eventListener. What do you think ?

y-lohse:
The check sounds like an ok solution to me, although I'd be curious to find out when and how `menuNode` can be undefined.

PR #849, opened 2019-02-26
fix(ActionMenu): Add a check before adding / removing an event listener

Crash--:
Should fix https://github.com/cozy/cozy-ui/issues/618


nono:
Ping @Crash--

Crash--:
thanks for the reminder @nono !

PR #852, opened 2019-02-28
feat: Add flexbox utility classes

GoOz:
Added flexbox utility classes [inspired by Tachyons](https://github.com/tachyons-css/tachyons/blob/master/src/_flexbox.css).

Basically there's a class for every flexbox related properties.

Each one of them is also available for every basic Media Query.

For example, the class `u-flex` that displays an element with flexbox property goes like this
```css
.u-flex {
  display: flex;
}
@media (max-width: 23.438rem) {
  .u-flex-tt {
    display: flex;
  }
}
@media (max-width: 33.938rem) {
  .u-flex-t {
    display: flex;
  }
}
/* Etc, for every breakpoint */
```

[Documentation preview](https://gooz.github.io/cozy-ui/styleguide/section-utilities.html#kssref-utilities-flexbox-display)

~~📌 NB: Couldn't find a way (yet) to generate those classes with a function due to Stylus limitations.
See the [Stack Overflow question](https://stackoverflow.com/questions/54924167/calling-a-stylus-block-with-a-variable-in-an-iteration)~~
Found a workaround using `mixins()` instead of `{blocks}`. Thanks @nono 🙏

### DIFF
--- a/stylus/utilities/flexbox.styl
+++ b/stylus/utilities/flexbox.styl
@@ -1,0 +1,197 @@
+@require '../settings/breakpoints'
+
+flexbox()
+    display flex
+
+inline-flexbox()
+    display inline-flex
+
+/* 1 Fix for Chrome 44 bug.
+ * https//code.google.com/p/chromium/issues/detail?id=506893 */
+flexbox-auto()
+    flex 1 1 auto
+    min-width 0 /* 1 */
+    min-height 0 /* 1 */
+
+flexbox-none()
+    flex none
+
+flexbox-column()
+    flex-direction column
+flexbox-row()
+    flex-direction row
+flexbox-wrap()
+    flex-wrap wrap
+flexbox-nowrap()
+    flex-wrap nowrap
+flexbox-wrap-reverse()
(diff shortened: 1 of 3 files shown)


stylus/utilities/flexbox.styl
```diff
@@ -0,0 +1,197 @@
...
+
+breakpoints = {
+    'none': '',
+    'teeny': 'tt',
+    'tiny': 't',
+    'small': 's',
+    'medium': 'm',
+    'large': 'l',
+    'extra-large': 'xl'
+}
```
ptbrowne: What does none mean ?  When do we use it ?

I second bruno, why do we have so many breakpoints ?

GoOz: > What does none mean ? When do we use it ?

`none` means that there's no breakpoint to consider, in order to ouput the default class
```
.u-flex {
    display: flex;
}
```
> 
I second bruno, why do we have so many breakpoints ?

Because there is. I didn't add the breakpoints, it was already there. I just made sure there was a utility class working for every breakpoint we have in stock.

PR #856, opened 2019-03-05
fix(Menu): Top aligned Menu Icon

### LINKED ISSUE #654: MenuItem icon alignment
As part of an upcoming design, we need a `MenuItem` component, but with the icon being aligned at the top, and not in the middle, like this:

capture d ecran 2018-10-19 18 15 23

I don't think this is supported at the moment? Is that something you could integrate in cozy-ui @GoOz ?

GoOz:
Icons in menu item are now top aligned.

Fixed #654

[Preview](https://gooz.github.io/cozy-ui/react/#menu)


GoOz:
[(M) Améliorer la position des icones dans le composant menuItem](https://trello.com/c/89EVnDH1/440-m-am%C3%A9liorer-la-position-des-icones-dans-le-composant-menuitem)

PR #858, opened 2019-03-07
feat(Field): Expose Input's inputRef prop ✨

gregorylegarec:
It's pretty convenient to have this prop available in Field. It allow us to give them focus, for example in https://github.com/cozy/cozy-libs/pull/316.

### DIFF
--- a/react/Field/Readme.md
+++ b/react/Field/Readme.md
@@ -19,6 +19,38 @@ Like `Input` component, it can have the following properties:
 </form>
 ```
 
+- `inputRef`
+
+This property is mapped to `<Input />` component `inputRef` property.
+It gives access to the underlying `<input />` element, for example to give focus or move the caret.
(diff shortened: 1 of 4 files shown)


react/Field/Readme.md
```diff
@@ -19,6 +19,35 @@ Like `Input` component, it can have the following properties:
 </form>
 ```
 
+- `inputRef`
+
+##### Example
+
+```
```
ptbrowne: Can you add a description of prop ?

gregorylegarec: Does

> This property is mapped to `<Input />` component `inputRef` property.
It gives access to the underlying `<input />` element, for example to give focus or move the caret.

suits you ?

ptbrowne: Perfect.

PR #860, opened 2019-03-08
fix: Fix the prop-types for <Field>

nono:


nono:
Let's take #861 (better lint)

PR #861, opened 2019-03-08
Field: Add missing values for prop type 📝

gregorylegarec:
Add `date` and `select` to list of valid values for prop `type`.

This was causing warnings in Jest tests.


gregorylegarec:
Duplicate of #860, @nono is too quick.

gregorylegarec:
Actually I have a better lint.

PR #862, opened 2019-03-10
Enable tree shaking

ptbrowne:
The idea behind this PR is to enable apps to use tree-shaking with cozy-ui. As of now it is not possible for several reasons : 

- import/exports are transformed into require calls which prevents webpack to do static analysis and to flag unused exports
- sideEffects false is not present

The changes in this PR : 
- Do not touch import/export when transpiling otherwise webpack cannot use tree shaking in downstream apps (this is done via a config flag of babel-preset-cozy-app)
- sideEffects: false enables webpack to eliminate re-exports that are unnecessary in downstream apps. It has to be put on the lib side to tell webpack that the files of the package do not have side effects.

Additionally:

- transformRuntime helpers are imported from babel-runtime instead of being copied each time. This is done via a config flag in babel-preset-cozy-app.

before

```jsx
var _react = _interopRequireDefault(require("react"));

var _Icon = _interopRequireDefault(require("../Icon"));

var _classnames = _interopRequireDefault(require("classnames"));

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }

function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }

function _objectWithoutProperties(source, excluded) { if (source == null) return {}; var target = _objectWithoutPropertiesLoose(source, excluded); var key, i; if (Object.getOwnPropertySymbols) { var sourceSymbolKeys = Object.getOwnPropertySymbols(source); for (i = 0; i < sourceSymbolKeys.length; i++) { key = sourceSymbolKeys[i]; if (excluded.indexOf(key) >= 0) continue; if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue; target[key] = source[key]; } } return target; }

function _objectWithoutPropertiesLoose(source, excluded) { if (source == null) return {}; var target = {}; var sourceKeys = Object.keys(source); var key, i; for (i = 0; i < sourceKeys.length; i++) { key = sourceKeys[i]; if (excluded.indexOf(key) >= 0) continue; target[key] = source[key]; } return target; }
```

after `modules: false`

```jsx
function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }

function _objectWithoutProperties(source, excluded) { if (source == null) return {}; var target = _objectWithoutPropertiesLoose(source, excluded); var key, i; if (Object.getOwnPropertySymbols) { var sourceSymbolKeys = Object.getOwnPropertySymbols(source); for (i = 0; i < sourceSymbolKeys.length; i++) { key = sourceSymbolKeys[i]; if (excluded.indexOf(key) >= 0) continue; if (!Object.prototype.propertyIsEnumerable.call(source, key)) continue; target[key] = source[key]; } } return target; }

function _objectWithoutPropertiesLoose(source, excluded) { if (source == null) return {}; var target = {}; var sourceKeys = Object.keys(source); var key, i; for (i = 0; i < sourceKeys.length; i++) { key = sourceKeys[i]; if (excluded.indexOf(key) >= 0) continue; target[key] = source[key]; } return target; }

import React from 'react';
import Icon from '../Icon';
import cx from 'classnames';
```

after `helpers: true`

```jsx
import _extends from "@babel/runtime/helpers/extends";
import _objectWithoutProperties from "@babel/runtime/helpers/objectWithoutProperties";
import React from 'react';
import Icon from '../Icon';
import cx from 'classnames';
```

Before

image

After

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,6 +103,7 @@
     "webpack": "^4.17.2"
   },
   "dependencies": {
+    "@babel/runtime": "^7.3.4",
     "body-scroll-lock": "^2.5.8",
     "classnames": "^2.2.5",
     "date-fns": "^1.28.5",
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "watch:doc:react": "env BABEL_ENV=transpilation styleguidist server --config docs/styleguide.config.js",
     "watch:doc:kss": "nodemon --ext styl,md --watch stylus --exec 'yarn build:doc:kss && http-server build/styleguide -p 4242'"
   },
+  "sideEffects": false,
   "devDependencies": {
     "@babel/cli": "^7.2.3",
     "@babel/core": "^7.2.2",
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,5 +1,7 @@
 module.exports = {
-  presets: ['cozy-app'],
+  presets: [
+    ['cozy-app', { presetEnv: { modules: false } }]
+  ],
   env: {
     transpilation: {
       plugins: [
--- a/babel.config.js
+++ b/babel.config.js
@@ -1,6 +1,6 @@
 module.exports = {
   presets: [
-    ['cozy-app', { presetEnv: { modules: false } }]
+    ['cozy-app', { presetEnv: { modules: false }, transformRuntime: { helpers: true }}]
   ],
   env: {
     transpilation: {
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -2,7 +2,7 @@ import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 
-import { default as labelStyles } from '../Label/styles.styl'
+import labelStyles from '../Label/styles.styl'
 import styles from './styles.styl'
 import Label from '../Label'
 import Input from '../Input'
--- a/yarn.lock
+++ b/yarn.lock
@@ -839,6 +839,13 @@
   dependencies:
     regenerator-runtime "^0.12.0"
 
+"@babel/runtime@^7.3.4":
+  version "7.3.4"
+  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.3.4.tgz#73d12ba819e365fcf7fd152aed56d6df97d21c83"
+  integrity sha512-IvfvnMdSaLBateu0jfsYIpZTxAc2cKEXEMiezGGN75QcBcecDUKd3PgLAncT0oOgxKy8dd8hrJKj9MfzgfZd6g==
+  dependencies:
+    regenerator-runtime "^0.12.0"
+
 "@babel/template@7.0.0-beta.44":
   version "7.0.0-beta.44"
   resolved "https://registry.yarnpkg.com/@babel/template/-/template-7.0.0-beta.44.tgz#f8832f4fdcee5d59bf515e595fc5106c529b394f"


package.json
```diff
@@ -46,6 +46,7 @@
     "watch:doc:react": "env BABEL_ENV=transpilation styleguidist server --config docs/styleguide.config.js",
     "watch:doc:kss": "nodemon --ext styl,md --watch stylus --exec 'yarn build:doc:kss && http-server build/styleguide -p 4242'"
   },
+  "sideEffects": false,
```
nono: Did you find some guidelines to know when it is acceptable to use `sideEffects: false`? I've searched and didn't find some clear rules. But, IIRC, styles have no `export` and are viewed as as side-effect by webpack. So, we should probably exclude them here.

ptbrowne: > Did you find some guidelines to know when it is acceptable to use sideEffects: false? 

I have read https://github.com/webpack/webpack/tree/master/examples/side-effects. It is not acceptable when your code has sideEffects (inserting CSS files for example) but: 

* We are not in this case after transpilation
* We could be in this case without transpilation but I think it can be ignored. If we tell webpack that stylus files have side effects, it means that our component files have side effects, and we cannot optimize for component usages. We should strive for our stylus files for components not to have side effects (global CSS rules).

---

Longer answer :

### What does `sideEffects: false` tell webpack

Often, we have cases where we have `index.js` reexporting components : 

lib/A.jsx
```js
import styles from './A.styl'
export default () => <span>Component A</span>
```

lib/B.jsx
```js
import styles from './B.styl'
export default () => <span>Component B</span>
```

lib/index.js
```js
export { default as A } from './A'
export { default as B } from './B'
```

app/index.js
```js
import { A } from 'lib'
```

Here since we do not import B in app/index.js, the export is flagged as unnecessary in webpack: 

webpack::lib/index.js
```js
export { default as A } from './A'
/* harmony unused export */
export { default as B } from './B'
```

and it is dropped after the optimization phase.

optimized webpack::lib/index.js
```js
export { default as A } from './A'
```

And since the export is dropped, the code of B is no longer included in the final bundle. Yay !

### Side effects ?

The problem is when B has a side effect, for example setting a global in window, inserting a CSS etc... In our case though, the CSS is extracted when transpiling. Here are our non transpiled files : 

B.styl
```
.toto
  background: red
```

B.jsx
```jsx
import styles from './B.styl'
```

The content from B.styl is extracted into a big CSS stylesheet. 
```css
/* A.styl */
/* blablabla */

/* B.styl */
.toto {
  background: red
}
```

and the import is replaced by an object containing the name of the classes in the transpiled JS.

transpiled:B.js
```
const styles = { toto: 'toto' }
```

So after transpilation, we have no references to CSS files in our JS files; the big CSS stylesheet is meant to included by the host application. This means that our transpiled JS files have no side-effects.

---

This leaves the case of importing stylus files when using the non transpiled version, but I do not think this is a problem since we are not in the case of re-exporting. Webpack will not flag exports as unused (no exports to flag). So the CSS code cannot be ignored by webpack.

The only problem we could have would be the order of insertion of the CSS when working with non transpiled files. B.jsx can be ignored, thuse B.styl can be ignored. If B.styl includes global styles, like `body { background: red }`, it could potentially be a problem if this global CSS is inserted earlier or later. I personally think that we will have few cases like that, we should not rely on CSS order (inter-files) and we should not have global CSS rules in components styl.

nono: Thanks for the guide :+1:

PR #868, opened 2019-03-18
fix: Cross position with header

GoOz:
The modal cross button is supposed to be close to the top right corner of the modal if there's no title or any other content in the header.

That wasn't the case, it triggered only if the props `title` was undefined but it doesn't mean that there's no header at all, so I fixed this with the expected behaviour.

The method to do so feels a little "dirty". I'm open to suggestions if there's any better ones.

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -313,6 +313,9 @@ class Modal extends Component {
     const style = Object.assign({}, height && { height }, width && { width })
     const maybeWrapInPortal = children =>
       into ? <Portal into={into}>{children}</Portal> : children
+    const hasModalHeader = children.find(
+      child => child.type.name === 'ModalHeader'
+    )
     return maybeWrapInPortal(
       <div className={cx(styles['c-modal-container'], containerClassName)}>
         <Overlay
@@ -350,7 +353,8 @@ class Modal extends Component {
               {closable && (
                 <ModalCross
                   className={cx(closeBtnClassName, {
-                    [styles['c-modal-close--notitle']]: !title
+                    [styles['c-modal-close--notitle']]:
+                      !title && !hasModalHeader
                   })}
                   onClick={dismissAction}
                   color={closeBtnColor}


react/Modal/index.jsx
```diff
@@ -313,6 +313,9 @@ class Modal extends Component {
     const style = Object.assign({}, height && { height }, width && { width })
     const maybeWrapInPortal = children =>
       into ? <Portal into={into}>{children}</Portal> : children
+    const hasModalHeader = children.find(
+      child => child.type.name === 'ModalHeader'
```
Crash--: I think this will crash on Drive saying `can't read name of undefined` since Preact doesn't expose `type` object for its children.

Crash--: Here is a quick fiddle to illustrate. https://jsfiddle.net/j40snyd1/ enjoy :D

gregorylegarec: Do you know any alternative ?

Crash--: Also, when minified, name will change, type will not. 
```
 {this.props.children.find(({ type }) => type === ModalHeader)}
```

I think we should test if react / if preact and do similar thing.

y-lohse:
> The modal cross button is supposed to be close to the top right corner of the modal if there's no title or any other content in the header.

But if there's a title, it's also in the top right corner..? 😕

Asking this because I wonder if an explicit prop wouldn't be better a better alternative than a side-effect. Even if we drop preact support it's still not a great solution :/

GoOz:
To be precise…

If there's nothing in the header or no header at all it supposed to look like this
Screenshot 2019-04-02 19 02 17

If there's a header with some content, it's supposed to look like this
Screenshot 2019-04-02 19 02 30

We could use a props but i'm afraid it'll be forgotten. That's why I wanted something that checks automatically if there's a header or not and if not put a class on the close button. 

I'm very open to other ideas though.

y-lohse:
Would it possible to achieve this with pure css, with something along the lines of `.modal-header + .cross`

GoOz:
Not without changing the JSX too, I guess, since the cross is before the header in the DOM, we would need to move it.
But ultimately it could be done, yes. 👍

GoOz:
Since all our apps are now powered by React, can we move on this one?

ptbrowne:
I think https://github.com/cozy/cozy-ui/pull/868#discussion_r267411530 is not fixed.

GoOz:
@ptbrowne Yeah I didn't not mean to merge it. I meant we can work on this again. :)

ptbrowne:
Ah yes ;)

ptbrowne:
@y-lohse I think we can close right ?

y-lohse:
Yeah it's been stale for a while.

PR #870, opened 2019-03-18
feat: Add responsive spacing utility classes

GoOz:
Added responsive utility classes for spacing (padding & margin).

Basically, it's the same classes name as before, you just need to add the breakpoint suffix matching your breakpoint of choice. 

| Breakpoint (px size) | Suffix |
| --- | --- |
| `BP-teeny`  (375) | `tt` |
| `BP-tiny`  (543) | `t` |
| `BP-small`  (768) | `s` |
| `BP-medium` (1023) | `m` |
| `BP-large` (1200) | `l` |
| `BP-extra-large` (1439) | `xl` |

e.g. You want an element with `1rem` of margin all around on desktop and only `.5rem` on devices below tablets. 
You will have to set it like this: `<div class="u-m-1 u-m-half-s">My beautiful element</div>`

NB:
This seems a little verbose, I agree, but I had to repeat myself because:
- Stylus doesn't allow extending `$placeholders` inside Media Queries.
- Can't use `{block}` instead either, because, in a loop, it's the same syntax for both `{block}` and `{vars}`, so it just doesn't work.
- Stylus doesn't allow creating `mixins()` in a loop either.

Couldn't find any other solution but I will gladly update this code if someone does.

### DIFF
--- a/stylus/utilities/spaces.styl
+++ b/stylus/utilities/spaces.styl
@@ -1,6 +1,7 @@
 /*------------------------------------*\
   Space utilities
 \*------------------------------------*/
+@require '../settings/breakpoints'
 @require '../tools/mixins'
 
 // @stylint off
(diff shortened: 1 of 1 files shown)


GoOz:
[(𝐌) Contacts - Fiche contact révision UI](https://trello.com/c/0KDJUugd/1728-%F0%9D%90%8C-contacts-fiche-contact-r%C3%A9vision-ui)

GoOz:
[(𝐌) Contacts - Révision UI de la fiche contact](https://trello.com/c/yD0TpuOh/1115-%F0%9D%90%8C-contacts-r%C3%A9vision-ui-de-la-fiche-contact)

Issue #867, opened 2019-03-15
More utilities class


gregorylegarec:
Utilities classes are great but we're lacking classes for:
* border-radius
* background-color

GoOz:
@gregorylegarec What are you expecting them to do actually?

A set of classes for several border radiuses (previously decided with designers) and one to cancel it?

Same question for background-color but maybe with already available colors ?

gregorylegarec:
Exactly 👌

y-lohse:
There's a longer wishlist [here](https://github.com/cozy/cozy-ui/issues/687), inspired by tachyons

PR #880, opened 2019-03-22
feat: Add border-radius utility classes

GoOz:
Added those utility classes (obviously responsive)

- .u-bdrs-1 - Small border-radius (2px)
- .u-bdrs-2 - Medium border-radius (3px)
- .u-bdrs-3 - Large border-radius (4px)
- .u-bdrs-4 - Extra large border-radius (8px)
- .u-bdrs-circle - Round element with border-radius (100%)

[Preview](https://gooz.github.io/cozy-ui/styleguide/section-utilities.html#kssref-utilities-border)

Refs #867

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -1286,6 +1286,25 @@ Display an chip that represents complex identity
 */
 
 /*
+    Border
+
+    Border-radius
+
+    .u-bdrs-1 - 1st step of border-radius (2px)
+    .u-bdrs-2 - 2nd step of border-radius (3px)
+    .u-bdrs-3 - 3rd step of border-radius (4px)
+    .u-bdrs-4 - 4th step of border-radius (8px)
+    .u-bdrs-circle - Round element with border-radius (100%)
+
+    Markup:
+    <div class="{{modifier_class}}  u-bg-dodgerBlue" style="width:4rem; height:4rem; "></div>
+
+    Weight: 1
+
+    Styleguide utilities.border
+*/
+
+/*
     Display
 
     Classes to change the box model of an element
--- a/stylus/utilities/border.styl
+++ b/stylus/utilities/border.styl
@@ -1,0 +1,31 @@
+@require '../settings/breakpoints'
+@require '../tools/mixins'
+
+
+bdrs-1()
+    border-radius rem(2)
+
+bdrs-2()
+    border-radius rem(3)
+
+bdrs-3()
+    border-radius rem(4)
+
+bdrs-4()
+    border-radius rem(8)
+
+bdrs-circle()
+    border-radius 100%
+
+props = {
+    'bdrs-1': 'bdrs-1',
+    'bdrs-2': 'bdrs-2',
+    'bdrs-3': 'bdrs-3',
+    'bdrs-4': 'bdrs-4',
+    'bdrs-circle': 'bdrs-circle'
+}
+
+if cssmodules == true
+    cssModulesUtils(props, breakpoints)
+else
+    nativeUtils(props, breakpoints)


GoOz:
@y-lohse 

> I don't know if it's just me, but I don't find `bdrs` intuitive or easy to remember. What is it? `BorDer RadiuS`?
> I would find `br` a lot easier to remember.

Well, I used the naming [Emmet](https://emmet.io/) uses, and indeed it stands for `BorDer RadiuS`. `br` being the abbreviation for `border-right`. 
Meanwhile, Tachyon uses `br` for `border-right` & `br(0/1/2/…)` for `border-radius`, I think Tachyon is a little more confusing on this matter, but it's just what I feel.

> I also wonder about using `small`, `medium` etc versus numbers. Margins, paddings and dimensions are using numbers, while only font-size uses the "t-shirt" sizes. How do you decide to use one over the other?

Well, Margins, Paddings & dimensions use a scale.
Font-size & Border-radius don't, or a non-logical one anyway, it's purely arbitrary.

When using "shirt sizes", the size naming implies a logical order that the current values (font-sizes or radius values) can't do.

When using a logical scale, like `line-height` for spacings or dimensions you don't need "shirt sizes" to implies the order.

That's the way I see it anyway, but the complex thing is that I can't use number in class names for font-sizes or border-radius without being confusing. 
Let's say we use a class for border-radius like`.br-1`. Does the 1 stands for the scale step or the value ? Let's say that I know it's related to the scale step, and I know that `.br-1` = `border-radius: 2px`, `.br-2` = `border-radius: 3px` and `.br-3` = `border-radius: 4px`, what am I expecting `.br-4` to output? A logical scale would imply that I'm gonna get `border-radius: 5px` but you won't, you'll get `border-radius: 8px`.
Let's say the number in classes doesn't refer to the scale but to the value of the `border-radius`, wouldn't it be confusing to have `.br-2`, `.br-3`, `.br-4`, `.br-8` but no `.br-1` or `.br-5` and so on? We could create those "missing" classes but it's not part of our design guidelines, so it will never ever be used, it will remain useless code for ever.

As for using "shirt sizes" in spacings and dimensions which have a *lot* of possiblities can't be complicated at some point, there's a limited shirt sizes available, dimensions already have 10 possible values, and going with sizes like *XXXS* starts to make no sense to me.

But that's a very interesting question indeed, I'll take any other good ideas. :)

y-lohse:
Regarding `bdrs` vs `br` vs something else, it probably comes down to personal taste. I'd be curious to see what others think, but I can deal with remembering `bdrs` :)

_____

> Does the 1 stands for the scale step or the value ? 

I don't think it should ever stand for the value. As you explained, we never expose *all* the values, so that would be confusing. So it should always refer to a step on the scale.

But as a user of the utility classes, I don't really care whether the scale is "logical" or not. Nobody said steps on a scale need to strictly follow a mathematical function 😁 . 

On the other hand, I would rather not have to remember what property uses shirt sizes and what property uses numeric sizes. And I'd rather not remember t-shirt sizes at all (was it `xlarge` or `x-large`? etc). So for me, using `-1` `-2` `-3` etc should be the way to go for all scales.

Crash--:
> Regarding bdrs vs br vs something else, it probably comes down to personal taste. I'd be curious to see what others think, but I can deal with remembering bdrs :)

We already follow `Emmet` for other classes, let's stay consistent ? ;)

ptbrowne:
> On the other hand, I would rather not have to remember what property uses shirt sizes and what property uses numeric sizes. And I'd rather not remember t-shirt sizes at all (was it xlarge or x-large? etc). So for me, using -1 -2 -3 etc should be the way to go for all scales.

+1

GoOz:
Allrighty then. 🤠

ptbrowne:
:shipit:

PR #892, opened 2019-04-03
feat(I18n): Don't crash if the default lang can't be found

drazik:
An error is thrown if the date-fns locale corresponding to the default
lang can't be found, and it make the whole app crash. It should warn the
user that the locale can't be found in date-fns, but not crash the app.

### DIFF
--- a/react/I18n/format.jsx
+++ b/react/I18n/format.jsx
@@ -1,15 +1,23 @@
 import format from 'date-fns/format'
 import { DEFAULT_LANG } from '.'
 
+const getWarningMessage = lang =>
+  `The "${lang}" locale isn't supported by date-fns. or has not been included in the build. Check if you have configured a ContextReplacementPlugin that is too restrictive.`
+
 export const initFormat = (lang, defaultLang = DEFAULT_LANG) => {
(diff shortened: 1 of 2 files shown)


react/I18n/format.jsx
```diff
@@ -2,9 +2,14 @@ import format from 'date-fns/format'
...
 
 export const initFormat = (lang, defaultLang = DEFAULT_LANG) => {
-  const locales = {
-    [defaultLang]: require(`date-fns/locale/${defaultLang}/index.js`)
+  const locales = {}
+
+  try {
+    locales[defaultLang] = require(`date-fns/locale/${defaultLang}/index.js`)
+  } catch (err) {
+    console.warn(`The "${defaultLang}" locale isn't support by date-fns`)
```
ptbrowne:
```suggestion
    console.warn(`The "${defaultLang}" locale isn't supported by date-fns or has not been included in the build. Check if you have configured a ContextReplacementPlugin that is too restrictive.`)
```

I think our error messages too often do not provide a way for the user (developer in this case) to understand clearly what is going on or provide a suggestion.

PR #895, opened 2019-04-04
feat: Spinner now use Icon

Crash--:
should fix https://github.com/cozy/cozy-ui/issues/878 

https://crash--.github.io/cozy-ui/react/#spinner

### DIFF
--- a/react/Spinner/Readme.md
+++ b/react/Spinner/Readme.md
@@ -38,12 +38,21 @@ Show a spinner when loading something.
 </div>
 ```
 
-
 ### Message
 
 When you use `loadingType`, `<Spinner />` needs to be in an `<I18n />` wrapper as it uses `t` function.
 
-
 ```jsx static
-<Spinner loadingType='salut' />
+<Spinner loadingType="salut" />
+```
+
+```
+const I18n = require('../I18n').default;
+
+  <div>
+    <I18n lang='en' dictRequire={() => {}}>
+      <Spinner size='xxlarge' loadingType='hi' />
+    </I18n>
+
+</div>
 ```
--- a/react/Spinner/index.jsx
+++ b/react/Spinner/index.jsx
@@ -1,9 +1,9 @@
 import React from 'react'
+import Icon from '../Icon'
 import { translate } from '../I18n'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
-
 export const Spinner = ({
   t,
   loadingType,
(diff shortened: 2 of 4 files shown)


react/Spinner/index.jsx
```diff
@@ -13,19 +14,44 @@ export const Spinner = ({
...
-          [styles[`c-spinner--${size}`]]: size
+          [styles['c-spinner--nomargin']]: noMargin
         },
         className
       )}
     >
+      <Icon
+        icon="spinner"
+        color={color}
+        className={iconStyle['icon--spin']}
```
GoOz: `<Icon />` has a props `spin` that already add the class.
See https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L45-L47
Unless you had an issue with this but it should use the props, right?

Crash--: thanls!

react/Spinner/index.jsx
```diff
@@ -13,19 +13,39 @@ export const Spinner = ({
   size,
   className
 }) => {
+  let realSize = ''
+  switch (size) {
```
cedricmessiant: why not a mapping?

```
const realsizeMapping = {
  tiny: 8,
  small: 12,
  ...
}

const realsize = realsizeMapping(size)```

File: react/Spinner/Readme.md [hunk 3]
```diff
@@ -38,12 +38,21 @@ Show a spinner when loading something.
...
+
+```
+const I18n = require('../I18n').default;
+
+  <div>
+    <I18n lang='en' dictRequire={() => {}}>
+      <Spinner size='xxlarge' loadingType='hi' />
+    </I18n>
+
+</div>
```
cedricmessiant: maybe this doc change should be in a distinct commit?

y-lohse:
@GoOz do you know if these variables are still useful ? https://github.com/cozy/cozy-ui/blob/c3232993d654b818618990c7e9e04b53d641d8f6/stylus/settings/icons.styl#L52-L75 They seem unused in cozy-ui. How likely is it that they are used in apps?

Crash--:
> They seem unused in cozy-ui. How likely is it that they are used in apps? 

In fact some of them are actually used : `$icon-16` for instance. 

Also I wanted to remove `$icon-spinner-white` (and others with the associated svg) but they are used for the button

GoOz:
I guess apps that use the old spinner way have a chance to use those placeholders. Not sure which one though.

PR #897, opened 2019-04-09
fix: SelectionBar button label

Crash--:
Since `Button` has to have a `label` for accessibility, we should add a label to the selection bar. 

We have to create a new translation string in our app that use SelectionBar

### DIFF
--- a/react/SelectionBar/index.jsx
+++ b/react/SelectionBar/index.jsx
@@ -64,6 +64,8 @@ const SelectionBar = ({
         className={styles['coz-action-close']}
         onClick={hideSelectionBar}
         extension="narrow"
+        label={t('SelectionBar.close')}
+        iconOnly
       >
         <Icon icon="cross" />
       </Button>


ptbrowne:
It is inconvenient for apps to ship with UI translation strings. Maybe cozy-ui could export JSON translation files that could extend the app translation strings ?

Something that could be used like this ?

```
import { translate } from 'cozy-ui/react'
import { fr, en } from 'cozy-ui/locales'

translate.extend(fr)
translate.extend(en)
```

Maybe the translation string could be namespaced under cozy-ui : `cozy-ui.SelectionBar.close` ? I think it would be clearer.

y-lohse:
That would definitely be better 👍 .

I see 2 ways to handle this:

1. Combine translation files at runtime — this means that cozy-ui needs a separate transifex project, and exposes en, fr and es strings
2. Combine the translation files at uh… pre-compile time? So that any app that uses cozy-ui can import the english strings, and handle the translations as part of their usual process.

We could use 1 for other projects, but option 2 seems a lot easier.

ptbrowne:
I don't understand what you mean by combine ? I am not sure I have As an application developer, I do not want to manage the translations of cozy-ui so option 2 is a no-go for me.

y-lohse:
I meant combining the app's translation strings with the translations of cozy-ui. I also edited my comment, I meant to say option 1 (essentially your previous example) is easier to do.

(Option 2, in more details, would be to combine the app's `en.json` with cozy-ui's `en.json`, and push the combined file to transifex.)

nono:
Ping: what is the status of this PR?

y-lohse:
@Crash-- I think we can merge this? Improvements to the translations should be done in a separate PR.

Crash--:
@y-lohse agreed.

See https://github.com/cozy/cozy-ui/issues/914

PR #901, opened 2019-04-15
fix: Remove unused rules on u-flex-auto class

y-lohse:
These rules were interfering with other `min-height` and `min-width` rules without us being in the clear about what problem they solve.


GoOz:
It's a 2015 so called bug… I wouldn't bother about it anyway 🙃

PR #903, opened 2019-04-17
feat: Add AppLinker

ptbrowne:
We need to open apps throughout the cozy platform. The rules to open an app vary depending on whether we are on mobile, whether the native app is installed etc...

It is a refactor from AppItem in the bar. It is more low level than AppItem as it only deals with the logic of what to do onClick. The presentational logic is left to the implementer. This way we can use it for AppItem or for simple links.

### DIFF
--- a/react/AppLinker/index.jsx
+++ b/react/AppLinker/index.jsx
@@ -21,27 +21,6 @@ const memoizedCheckApp = expiringMemoize(
   appInfo => appInfo.appId
 )
 
-/**
- * Render-props component that provides onClick/href handler to
- * apply to an anchor that needs to open an app.
- *
- * Handles several cases:
- *
- * - On mobile app and other mobile app available
- * - On web (not mobile)
- * - On web mobile
- *
- * @example
- * ```
- * <AppLinker slug='banks' href='http://dalailama-banks.mycozy.cloud'>{
- *   ({ onClick, href, name }) => {
- *     <a href={href} onClick={onClick}>
- *       Open { name }
- *     </a>
- *   }
- * }</AppLinker>
- * ```
- */
 export class AppLinker extends React.Component {
   state = {
     nativeAppIsAvailable: null
--- a/react/AppLinker/index.jsx
+++ b/react/AppLinker/index.jsx
@@ -1,0 +1,146 @@
+/* global __TARGET__ */
+
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import {
+  checkApp,
+  startApp,
+  isMobileApp,
+  isMobile,
+  openDeeplinkOrRedirect
+} from 'cozy-device-helper'
+import { NATIVE_APP_INFOS } from 'cozy-ui/transpiled/react/AppLinker/native'
+
+import expiringMemoize from './expiringMemoize'
+
+const expirationDelay = 10 * 1000
+const memoizedCheckApp = expiringMemoize(
+  checkApp,
+  expirationDelay,
+  appInfo => appInfo.appId
+)
+
+/**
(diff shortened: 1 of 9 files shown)


react/AppLinker/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+  checkApp,
+  startApp,
+  isMobileApp,
+  isMobile,
+  openDeeplinkOrRedirect
+} from 'cozy-device-helper'
+
+import expiringMemoize from './expiringMemoize'
+import { appShape } from 'cozy-ui/transpiled/react/propTypes'
+import { NATIVE_APP_INFOS } from 'cozy-ui/transpiled/react/AppLinker/native'
```
Crash--: I don't know if cozy-ui has to know about that. But I don't know how we can have these informations dynamically. Any idea?

react/AppLinker/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @example
+ * ```
+ * <AppLinker app={{ slug: 'banks', href: 'http://dalailama-banks.mycozy.cloud' }}>{
+ *   ({ onClick, href }) => {
+ *     <a href={href} onClick={onClick}>
+ *       Open Banks
+ *     </a>
+ *   }
+ * }</AppLinker>
```
Crash--: 👍

react/propTypes.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types'
+
+export const appShape = PropTypes.shape({
+  slug: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+  namePrefix: PropTypes.string,
+  comingSoon: PropTypes.bool,
+  href: PropTypes.string,
+  links: PropTypes.shape({ icon: PropTypes.string.isRequired })
+})
```
Crash--: The idea is to write all our UI proptypes here?

react/AppLinker/index.jsx
ptbrowne: I think it is OK that cozy-ui has this knowledge, it does not change much.

react/propTypes.js
ptbrowne: Not necessarily, mostly those which are shared. AppLinker is a refactor of a component from the bar (AppItem), that had this propTypes, this is why I added it here.

react/AppLinker/index.jsx
```diff
@@ -0,0 +1,131 @@
...
+ * ```
+ */
+export class AppLinker extends React.Component {
+  state = {
+    isMobileAppAvailable: null
+  }
+
+  constructor() {
+    super()
```
drazik: Nit: It works as is, but it's cleaner to call `constructor` and `super` with `props` as parameter. More info [here](https://overreacted.io/why-do-we-write-super-props/)

react/AppLinker/index.jsx
```diff
@@ -0,0 +1,131 @@
...
+    if (appInfo) {
+      const isMobileAppAvailable = Boolean(await memoizedCheckApp(appInfo))
+      this.setState({ isMobileAppAvailable })
+    }
+  }
+
+  onAppSwitch = () => {
+    const { onAppSwitch } = this.props
+    if (typeof onAppSwitch === 'function') {
+      this.switchTimeout = setTimeout(() => {
```
y-lohse: The delay before the app switch comes from the cozy-bar, but I'm not sure we should have it here? This seems to be a UX thing specific to cozy-bar, no?

ptbrowne: You're right, I am removing the setTimeout logic and delay, it will be the responsibility of the app.

Issue #902, opened 2019-04-16
Replace remaining embedurl


y-lohse:
We are using the stylus `embedurl` in a few places: https://github.com/cozy/cozy-ui/search?q=embedurl&unscoped_q=embedurl

The problem is that these embeds don't go through webpack, which makes them impossible to override in B2B theming. They also rely on a svg with a fixed color, making it cumbersome to support new colors for these icons.

 The replacement technique depends on the situation, but most of the time an `Icon` component is preferable.

y-lohse:
This is nearly done. All "actual" icons are using the Icon component now, the checkboxes as well. AFAIL there's only the arrow for the select box that's still left. @GoOz can you confirm?

GoOz:
Yeah. 
Basically it remains only in the code the embedurl icon for an `input[type=checkbox]` and a `select` but the React components `CheckBox` and `SelectBox` don't use that style (well it does but it overrides it mostly). So the migration shoud be easy. 
I left the old styles because it's still used and I can't change those styles without changing the DOM, also it allows having something consistent with the Cozy design guidelines when you're not using the React components with the default DOM for an `input[type=checkbox]` and a `select`, which the React components don't support.

y-lohse:
Ok cool — so we have everything we need to have themed form components, it's just a matter of migrating. And we won't make a breaking change to force the migration at the moment, no rush 👍

PR #904, opened 2019-04-18
Fix/remove embed url

GoOz:
- Removed `u-error-warning`, `c-link--upload` & `c-link--delete` as it is old and obsolete
- Removed `embedurl()` from Button Client. Will require some changes on Drive (PR incoming)
- Updated busy button
  - use Spinner icon instead of CSS background
  - apply disabled styles to busy button
  - prevent click on busy button
- Add ButtonClient component (from Drive)
- Add BannerClient component (from Drive)

[New busy button's demo](https://gooz.github.io/cozy-ui/react/#button)

Refs #902


GoOz:
[(𝐒) [cozy-ui] Replace remaining embedurl (tous sauf le formulaire)](https://trello.com/c/zD7HtKuc/1284-%F0%9D%90%92-cozy-ui-replace-remaining-embedurl-tous-sauf-le-formulaire)

stylus/components/button.styl
```diff
@@ -99,11 +99,8 @@ $button
             cursor not-allowed
 
     &[aria-busy=true]
-        > span::after
-            @extend    $icon-16
-            @extend    $icon-spinner-white
-            position   relative
-            top        rem(-1)
+        opacity .5
+        pointer-events none
```
y-lohse: Why are we adding these 2 rules ?

stylus/components/button.styl
```diff
@@ -243,6 +215,7 @@ $button-client
     justify-content   center
     align-items       center
     height            auto
+    min-height        rem(56)
```
y-lohse: Why are we adding this?

stylus/components/button.styl
GoOz: For this:
> - apply disabled styles to busy button
> - prevent click on busy button

GoOz: It was some of the request from the designer about busy buttons

stylus/components/button.styl
GoOz: To match the design. It should have been done from the beginning. I may be overzealous on this one.

stylus/components/button.styl
y-lohse: Ah yes ok. Wouldn't it be possible to set a `disabled` attribute on the button when it's busy ?

GoOz:
@y-lohse 

> Thanks! Looks good overall, however the client button css rules are now very tied to the markup, and the markup is… in a different repo. Would it make sense to add the actual component in cozy-ui?

Agreed, although I'm a bit afraid about the functionnal part about cookie/storage. Could try though.

> (Also the CI is not happy)

Yeah couldn't figure out why yet, tests are okay locally…

stylus/components/button.styl
GoOz: I don't know. I asked myself the question but I figured that since there are two distinct aria-attributes for those two, it may not be a good idea, semantically speaking.
But I actually don't know enough on the aria topic to be sure, that's just a feeling I have.

y-lohse:
> Agreed, although I'm a bit afraid about the functionnal part about cookie/storage. Could try though.

Yeah ideally the storage part wouldn't be in cozy-ui, just the markup. The apps can have their own logic to show/hide this button.

react/PushClientButton/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+  return (
+    <a
+      href={href}
+      //eslint-disable-next-line react/jsx-no-target-blank
+      target="_blank"
+      className={cx(styles['c-btn-client'], className)}
+      onClick={onClick}
+    >
+      <figure>
+        <img width="32" height="32" src={DeviceIcon} />
```
Crash--: I'm wondering if `DeviceIcon` should not be a props with this default value. It'll be easier to override?

GoOz: hum could be yes. We don't have any other use of this banner but I guess it would be future proof.

react/PushClientBanner/Readme.md
```diff
@@ -0,0 +1,28 @@
+## Push Client Banner
+Banner to advertise for Cozy desktop client.
+
+### Basic usage
+
+```
+<BannerClient
+  text="Ajoutez vos documents et photos à votre Cozy depuis votre ordinateur : ils seront sauvegardés en lieu sûr et accessibles à tout moment !"
```
y-lohse: Not blocking, but the rest of the examples are in english, it would be cool to have this one in english too!

stylus/components/button.styl
y-lohse: Mmh alright. I'm worried about pointer-events as it effectively "kills" the element through css — makes it hard to figure out why events are not firing for a certain element. Would changing the cursor be a viable option, or would we miss something ?

react/PushClientBanner/styles.styl
```diff
@@ -0,0 +1,35 @@
...
+.c-banner-client
+    position relative
+
+    +small-screen('min')
+        margin-top     -1.875rem
+        margin-bottom  1.875rem
+
+    .c-btn-clientMobile
+        @extend $button-client-mobile
+        margin  0
```
y-lohse: Nit: could the margin be part of `$button-client-mobile`?

GoOz: I guess yeah.

stylus/components/button.styl
GoOz: The thing is that a button can be disabled but a link can't and the Button style can apply to either a button or a link. 
The `pointer-events` trick is the only one afaik that can't prevent clicks on both elements. 
Do you see any concrete usecase that could be a problem?

y-lohse: Yeah I understand why disabling it is not a good fit. My concern is something like this https://codesandbox.io/s/xl4w5j8y14 — but I don't know if we ever use stuff like that.

If we want to make the button look inert, I think `cursor: auto` would be enough? If we really want to prevent all clicks on busy buttons… I think js is the right place to do that, not a css rule. Unless `pointer-events` does achieves something that `cursor` doesn't?

GoOz: > My concern is something like this https://codesandbox.io/s/xl4w5j8y14 — but I don't know if we ever use stuff like that.

I don't get what I'm supposed to see here. :/

Also `cursor` doesn't prevent anything, it just change the cursor, you can still click the link/button. 
But I see your point about not really being CSS's job to deal with this behaviour.

I'm ok removing this property but what do we do for links then? Directly dealing with this in the component ?

y-lohse: Discussed it live; the choices are:

1. Only visually disable the button through css, actual clicks have to be disabled via js
2. Disable everything via css, but if we ever need interactions on a busy button we'll have to change this

There's no use case for 2 at the moment, so we'll keep it as is in the PR and update this part of the code if we ever need to.

docs/webpack.config.js
```diff
@@ -30,6 +30,11 @@ module.exports = {
             }
           }
         ]
+      },
+      {
+        test: /\.(png|gif|jpe?g|svg)$/i,
+        exclude: /(sprites|public)/,
```
y-lohse: Nit: I don't think the exclude rule is actually needed here.

PR #905, opened 2019-04-18
feat: Add InitialsAvatar component

cedricmessiant:
Component that displays an avatar that uses the initials of the contact's name as text.
(I think I need to add some docs about this component)


react/Avatar/InitialsAvatar.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import Avatar, { avatarPropTypes } from '.'
```
Crash--: Should we need to use aliased path on UI too? cc @y-lohse @GoOz

y-lohse: Not that I'm aware no.

GoOz: Is that a webpack thingy?

react/Avatar/InitialsAvatar.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    const name = {
+      givenName: 'Kevon',
+      familyName: 'Willms'
+    }
+    const component = renderer.create(
+      <InitialsAvatar name={name} size="small" />
+    )
+    const tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
+  })
```
ptbrowne: Can you add a test on a limit case where givenName or familyName does not exist ?

I suspect we'll have an undefined in the resulting string.

ptbrowne:
I advocated for the Avatar with text and without direct support for initials for flexibility reasons.
Here we are coupling the data model to the presentational layer, ie if someone wants to use InitialsAvatar, (s)he has to conform to the io.cozy.contacts doctype.

> Why not just put the getInitials() function into Avatar component?

At first I was appealed by this idea, but I think getInitials() does not belong in the presentational layer but on the model. IMHO there should be a Contact class or utils, nearer to the definition of the doctypes (for example in cozy-libs/cozy-doctypes) to be used for this purpose.

```
<Avatar text={Contact.getInitials(contact)} />
```

To me this is more flexible, and model and presentation responsibilities are better separated.

GoOz:
@ptbrowne ah yes this IS the reason I forgot. 🙃

cedricmessiant:
My initial problem was that we had two different `getInitials` functions (in drive and in contacts) and two different behaviors. In the end, I agree that we should not have this contact related code in `cozy-ui` (even if you can actually also use `InitialsAvatar` with any string). So we should move `getInitials` outside of contacts and drive but drive/contacts is not the right place. Thanks

Crash--:
Having an `<InititalsAvatar>` force us to be consistent between Applications. Having too much flexibility lead us to have multiple usages & implementations. 

Maybe we should rename it to reflect the fact that this component is coupled to `io.cozy.contacts` doctype ? 

I think having a method in `cozy-doctypes` is a good idea, but I don't think it'll be used and well-known. 

Maybe we should couple the two solutions : 
- create a getInitials in cozy-doctypes
- create a `<AvatarContact />` in UI calling `cozy-doctypes.getInitials()` ?

ptbrowne:
> I think having a method in cozy-doctypes is a good idea, but I don't think it'll be used and well-known.

I think we should push to have model methods and classes in cozy-doctypes. It should be the go-to when sharing model logic between apps. It should be known. Maybe we should advertise it better in the tutorials to build apps.

ptbrowne:
Ping @cedricmessiant, do you have time to work on this ?

cedricmessiant:
I think we can close it for now.

PR #907, opened 2019-04-19
fix: label prop is required in Button

cedricmessiant:
Now that `label` prop is required in `Button` component, we have warning in our apps because `SelectionBar` renders a `Button` without `label`. We could add an empty string as `defaultProp` for `label` but I think that the idea is to force users to make it explicit.

### DIFF
--- a/react/Field/Readme.md
+++ b/react/Field/Readme.md
@@ -43,7 +43,7 @@ class FieldWithFocus extends React.Component {
             inputRef={c => this.component = c}
             label="I can have focus"
             placeholder="Focus please" />
-        <Button onClick={this.setFocus}>Set Focus</Button>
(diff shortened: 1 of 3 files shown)


Crash--:
I don't think this it the right fix. We should add the label and iconOnly to our button call. Label is used for A11y purpose

cedricmessiant:
Change done, thanks

y-lohse:
Thanks! The changes look good, but the CI seems to have a problem?

ptbrowne:
Need to `jest -u` ;) 

https://travis-ci.org/cozy/cozy-ui/builds/522179447#L1243

cedricmessiant:
Sorry about this, thanks for the reviews!

PR #910, opened 2019-04-23
chore: Don't pin dependencies

y-lohse:
See https://github.com/cozy/cozy-ui/pull/693#issuecomment-479042864 — we don't pin dependencies in this repo, and we don't want renovate to do it either.


ptbrowne:
TIL :)

PR #912, opened 2019-04-23
Add xsmall modal size

CPatchane:
An new size with 24rem as width
Deployed on https://cpatchane.github.io/cozy-ui/react/#modal

### DIFF
--- a/react/Modal/Readme.md
+++ b/react/Modal/Readme.md
@@ -26,12 +26,13 @@ initialState = { modalOpened: false};
 
 ### Size
 
-Several sizes avalaible: `small`, `medium`, `large`, `xlarge`, `xxlarge`.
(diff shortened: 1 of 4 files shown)


ptbrowne:
What's up with the cross ?



I thought we were going responsive, the sizes of the modal do not seem responsive ? Maybe an issue to open ?

LGTM otherwise but it'd be nice if the cross was fixed since we are working on modals.

CPatchane:
Locally watching I don't have any issue with the cross :/
Screenshot 2019-04-23 at 16 48 59

GoOz:
Yeah that's an issue occuring on styleguidist sometimes. It's an issue related to CSS order, not to this PR.

GoOz:
> To be perfectly honest, I have the feeling that it should be the content that should dictate the size of these modals.

I don't agree. The container should set the available space for a given content and this content will naturally use all this available space. 
Also I don't think this is really relevant to this PR.

PR #917, opened 2019-04-24
fix: Mark css files as having side effects

y-lohse:
Fixes https://github.com/CPatchane/create-cozy-app/issues/1198 — the global css files we import from cozy-ui are definitely not side effect free.


CPatchane:
Thanks @y-lohse, I didn't see it was flagged as no side effects here 👍

PR #918, opened 2019-04-24
feat: Universal link

Crash--:
Added support for Universal link for iOS

### DIFF
--- a/react/AppLinker/index.jsx
+++ b/react/AppLinker/index.jsx
@@ -2,16 +2,17 @@
 
 import React from 'react'
 import PropTypes from 'prop-types'
-
 import {
   checkApp,
   startApp,
   isMobileApp,
   isMobile,
-  openDeeplinkOrRedirect
+  openDeeplinkOrRedirect,
+  isAndroid
 } from 'cozy-device-helper'
-import { NATIVE_APP_INFOS } from 'cozy-ui/transpiled/react/AppLinker/native'
+import { generateUniversalLink, getUniversalLinkDomain } from './native'
 
+import { NATIVE_APP_INFOS } from 'cozy-ui/transpiled/react/AppLinker/native.config'
 import expiringMemoize from './expiringMemoize'
 
 const expirationDelay = 10 * 1000
@@ -23,7 +24,8 @@ const memoizedCheckApp = expiringMemoize(
 
 export class AppLinker extends React.Component {
   state = {
-    nativeAppIsAvailable: null
+    nativeAppIsAvailable: null,
+    isFetchingAppInfo: false
   }
 
   constructor(props) {
@@ -67,17 +69,23 @@ export class AppLinker extends React.Component {
     })
   }
 
+  //Will be removed soon when Android universallink will be tested
   openNativeFromWeb(ev) {
     if (ev) {
       ev.preventDefault()
     }
 
-    const { href, slug } = this.props
+    const { href, slug, nativePath } = this.props
     const appInfo = NATIVE_APP_INFOS[slug]
+
     this.onAppSwitch()
-    openDeeplinkOrRedirect(appInfo.uri, function() {
-      window.location.href = href
-    })
+
+    openDeeplinkOrRedirect(
+      appInfo.uri + (nativePath === '/' ? '' : nativePath),
+      function() {
+        window.location.href = href
+      }
+    )
   }
 
   openWeb() {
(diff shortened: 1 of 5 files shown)


react/AppLinker/index.jsx
```diff
@@ -107,19 +115,36 @@ export class AppLinker extends React.Component {
...
+  href: PropTypes.string.isRequired,
+  /*
+    Path used for "native link"
+  */
+  nativePath: PropTypes.string
 }
 
 export default AppLinker
-export { NATIVE_APP_INFOS }
+export { NATIVE_APP_INFOS, getUniversalLinkDomain, generateUniversalLink }
```
y-lohse: Why are we exporting the universal link helpers from this component ?

react/AppLinker/native.js
```diff
@@ -1,14 +1,20 @@
...
-    name: 'Cozy Drive'
-  },
-  banks: {
-    appId: isAndroidApp() ? 'io.cozy.banks.mobile' : 'io.cozy.banks',
-    uri: 'cozybanks://',
-    name: 'Cozy Banks'
-  }
+import { UNIVERSALLINK_URL } from 'cozy-ui/transpiled/react/AppLinker/native.config'
+export const getUniversalLinkDomain = () => {
+  return UNIVERSALLINK_URL
```
y-lohse: Is there a point to having a function and a constant exposing the same value?

react/AppLinker/native.config.js
```diff
@@ -0,0 +1,15 @@
...
+    uri: 'cozydrive://',
+    name: 'Cozy Drive'
+  },
+  banks: {
+    appId: isAndroidApp() ? 'io.cozy.banks.mobile' : 'io.cozy.banks',
+    uri: 'cozybanks://',
+    name: 'Cozy Banks'
+  }
+}
+export const UNIVERSALLINK_URL = 'https://links.mycozy.cloud'
```
y-lohse:
```suggestion
export const UNIVERSAL_LINK_URL = 'https://links.mycozy.cloud'
```

and the same for the other uses of the constant?

react/AppLinker/native.js
```diff
@@ -1,14 +1,20 @@
...
+}
+/*
+  slug string drive 
+  path : /path/to/view
+  fallbackUrl: string : https://...mcyozy.cloud
```
y-lohse:
```suggestion
  fallbackUrl: string : https://...mycozy.cloud
```

react/AppLinker/native.spec.js
```diff
@@ -0,0 +1,36 @@
+import { generateUniversalLink, getUniversalLinkDomain } from './native'
+describe('native functions', () => {
+  it('should generate the universalink', () => {
+    const universalLink = generateUniversalLink({
+      slug: 'drive',
+      nativePath: '/files/1',
+      fallbackUrl: 'https://drive.cozy.tools/files/1'
+    })
+    const endLink =
+      getUniversalLinkDomain() +
```
y-lohse: I would suggest not using a tested function to generate the *expected* result. I think hardcoding the domain here would be ok, no?

react/AppLinker/index.jsx
Crash--: It seems easier to do: 
import AppLinker, {getUniversalLinkDomain} from 'cozy-ui/AppLinker'  

than : 
import AppLinker from 'cozy-ui/AppLinker'  
import {getUniversalLinkDomain} from 'cozy-ui/Applinker/native'

react/AppLinker/native.js
Crash--: I prefer to have a getter. It gives us more flexibility to change the behavior / add check if needed (like checking if the url ends with a '/' or not).

react/AppLinker/index.jsx
y-lohse: But is `getUniversalLinkDomain ` going to be used outside of cozy-ui? If yes we can leave it as it is :)

react/AppLinker/native.js
y-lohse: I don't mind the getter, but then wouldn't it be easier to not have the constant, or at least in the same file?

react/AppLinker/index.jsx
Crash--: Yes, it will. (in the mobile handleDeeplink process)

react/AppLinker/native.js
Crash--: We need to override the value :/

PR #920, opened 2019-04-25
fix(utilities): Add font-style utilities

CPatchane:
Screenshot 2019-04-25 at 15 23 29


CPatchane:
@y-lohse I don't know, since it's a valid value for this property I wanted to be exhaustive

y-lohse:
Alright, can you remove it please? We don't *need* to be exhaustive and it seems that `oblique` is rarely used but also inferior to italic. The rest GLTM, thank you!

PR #949, opened 2019-04-29
Field: Inject name prop

gregorylegarec:
Until now, we were not passing `name` prop to `Input`, `TextArea`, and `SelectBox` components.

Having it allows us a better management of inputs from ancestor component, like in Harvest were we need to know which fied has focus.

Inject name prop into Input, Textarea and SelectBox
components.

Related to https://github.com/cozy/cozy-libs/pull/396

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -80,6 +80,7 @@ const Field = props => {
     label,
     id,
     inputRef,
+    name,
     type,
     value,
     placeholder,
@@ -109,6 +110,7 @@ const Field = props => {
           <Textarea
             disabled={disabled}
             id={id}
+            name={name}
             placeholder={placeholder}
             value={value}
             error={error}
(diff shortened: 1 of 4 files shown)


react/Field/index.jsx
```diff
@@ -148,6 +151,7 @@ const Field = props => {
             fullwidth={fullwidth}
             id={id}
             inputRef={inputRef}
+            name={name}
```
Crash--: add name to proptype

react/SelectBox/SelectBox.jsx
```diff
@@ -273,6 +273,7 @@ class SelectBox extends Component {
       styles: reactSelectStyles,
       breakpoints: { isMobile },
       classNamePrefix,
+      name,
```
Crash--: add to proptype

gregorylegarec:
@Crash-- README updated 👍

PR #950, opened 2019-04-29
SelectBox: Retrieve input ref from ReactSelect

gregorylegarec:
Provide an `inputRef` prop which will retrive the input ref from `ReactSelect` component.

The `SelectBox now provides the same API as `Input` components.

In harvest, it will allow us to manage focus and focus event from input fields.

Related to https://github.com/cozy/cozy-libs/pull/396

### DIFF
--- a/react/SelectBox/Readme.md
+++ b/react/SelectBox/Readme.md
@@ -103,6 +103,26 @@ const options = [
 <SelectBox options={options} fullwidth />
 ```
 
+#### `inputRef`
+
+This property gives access to the underlying `<input />` element contained into a `ReactSelect` component, for example to give focus or move the caret.
+
+```
+const options = [
+  { value: 'chocolate', label: 'Chocolate' },
+  { value: 'strawberry', label: 'Strawberry', isDisabled: true },
+  { value: 'vanilla', label: 'Vanilla' },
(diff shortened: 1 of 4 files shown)


react/SelectBox/Readme.md
```diff
@@ -103,6 +103,26 @@ const options = [
 <SelectBox options={options} fullwidth />
 ```
 
+#### `inputRef`
+
+This property gives access to the underlying `<input />` element contained into a `ReactSelect` component, for example to give focus or move the caret.
```
y-lohse:
```suggestion
This property gives access to the underlying `<input />` element contained in a `ReactSelect` component, for example to give focus or move the caret.
```

gregorylegarec:
@Crash-- Done, my bad 🙏

gregorylegarec:
@y-lohse The thing is, we can get a ref to the select component, not the input itself. I retro-engineered (yeah ok I used a smart `console.debug()`) the select ref to access directly to the `<input />`

y-lohse:
Yeah I did some testing too and dug in the source code of react-select and couldn't find another way to do it either :)

gregorylegarec:
@y-lohse I am neither a big fan of this solution but it seems to be the only way to do this for now.

PR #951, opened 2019-04-30
fix: Update Spinner default color

GoOz:
Spinner now uses our blue (dodgerblue via primaryColor) for default color and not random one.


ptbrowne:
Cool !

Issue #797, opened 2019-01-15
The button should not have margins


CPatchane:
https://github.com/cozy/cozy-ui/blob/2ea6c2aa02c45ad0e3dd2ce5b900db666badb1e3/stylus/components/button.styl#L60

I think we should remove this line, the button should not have margins since the application should handle its position with the possibility to give it margins or not.

ptbrowne:
Can you open a PR with this ? Thanks :)

GoOz:
I agree but beware, it will break stuff (and fix some others obviously)

y-lohse:
Same goes for https://github.com/cozy/cozy-ui/blob/2ea6c2aa02c45ad0e3dd2ce5b900db666badb1e3/stylus/components/button.styl#L319 — buttons shouldn't actively cancel the margins that they may otherwise get.

CPatchane:
Closing it, feel free to open a new issue if needed.

PR #952, opened 2019-04-30
feat(Infos): :nail_care: Improve Infos component with an action button

CPatchane:
Screenshot 2019-04-30 at 12 16 23

See https://cpatchane.github.io/cozy-ui/react/#infos

### DIFF
--- a/react/Infos/index.jsx
+++ b/react/Infos/index.jsx
@@ -3,21 +3,56 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
 import Icon, { iconPropType } from '../Icon'
-import Text from '../Text'
+import Text, { SubTitle } from '../Text'
 
-const Infos = ({ icon, text, className }) => {
+const Infos = ({ actionButton, icon, isImportant, text, className, title }) => {
   return (
-    <div className={cx(styles['infos'], 'u-p-1', className)}>
-      <div className={styles['infos--container']}>
-        {icon && <Icon icon={icon} className={styles['infos--icon']} />}
+    <div
+      className={cx(
+        'u-flex',
+        'u-flex-column',
+        'u-maw-6',
+        'u-mih-2',
+        'u-bdrs-4',
+        'u-p-1',
+        'u-bg-paleGrey',
+        'u-ta-left',
+        {
+          'u-bg-chablis': isImportant
+        },
+        className
+      )}
+    >
+      {!!title && (
+        <SubTitle
+          className={cx({
+            'u-pomegranate': isImportant
+          })}
+        >
+          {title}
+        </SubTitle>
+      )}
+      <div className="u-flex u-mv-1 u-w-100">
+        {icon && (
+          <Icon
+            icon={icon}
+            className={cx(
+              styles['infos--icon'],
+              'u-w-1',
+              'u-h-1',
+              'u-flex-shrink-0'
+            )}
+          />
+        )}
         <Text
-          className={cx(styles['infos--text'], {
+          className={cx('u-ta-left', {
             ['u-pl-half']: icon !== null
           })}
         >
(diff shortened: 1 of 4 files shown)


react/Infos/index.jsx
```diff
@@ -27,7 +62,13 @@ Infos.defaultProps = {
 }
 
 Infos.propTypes = {
+  /** An button that will be render in the infos component */
```
ptbrowne:
```suggestion
  /** A button that will be rendered at the bottom */
```

react/Infos/index.jsx
```diff
@@ -27,7 +62,13 @@ Infos.defaultProps = {
...
+  actionButton: PropTypes.element,
+  /** An icon to display at the left of the text */
   icon: iconPropType,
-  text: PropTypes.string
+  /** An important information will be displayed with some red colors */
```
ptbrowne:
```suggestion
  /** An important information will be displayed with red colors */
```

I think we should removed "some" as it is a vague word without much information. IMHO, technical writing should be direct and to the point.

react/Infos/index.jsx
```diff
@@ -3,21 +3,56 @@ import cx from 'classnames'
...
         <Text
-          className={cx(styles['infos--text'], {
+          className={cx('u-ta-left', {
             ['u-pl-half']: icon !== null
           })}
         >
           {text}
         </Text>
       </div>
+      <div className="'u-flex-shrink-0'">{!!actionButton && actionButton}</div>
```
ptbrowne:
```suggestion
      <div className="u-flex-shrink-0">{!!actionButton && actionButton}</div>
```

ptbrowne:
This looks great ! Cool that we begin to have this kind of component.

Is it normal that the button be aligned on the left ? I think it's more common to have this kind of CTA on the right ? @Claiw ?

Nit: The button has a left margin which makes it a bit misaligned with the content on the left, is it normal ? It sure can be done in another PR. I think it is related to https://github.com/cozy/cozy-ui/issues/797.

ptbrowne:
Nit: I am not totally sure of using a boolean isImportant here, to be inline with the Button API, and more flexible for other usages (yellow warning for example), I think a `variant="important"` or `theme="important"` would be better (so that we can have a `variant="warning"` later). 

@Claiw Do you foresee this kind of components with another color or is it prematured ?

CPatchane:
> Is it normal that the button be aligned on the left ? I think it's more common to have this kind of CTA on the right ? @Claiw ?

We need to have this on the left for our usage in Home (ping @joel-costa) so I have no opinion about that.

> The button has a left margin which makes it a bit misaligned with the content on the left, is it normal ? It sure can be done in another PR. I think it is related to #797.

Yeah related to this issue, I don't know all the implications about this change so I prefer to let @GoOz working on that when possible. For now I had `u-m-0` on my buttons :/

> I am not totally sure of using a boolean isImportant here, to be inline with the Button API, and more flexible for other usages (yellow warning for example), I think a variant="important" or theme="important" would be better (so that we can have a variant="warning" later).

I thought about something like that but we just have two variants here and I don't know if we will need other variants in the future. I would say that if we need more we will have to use a theme property like it's done for the Button component.

Claiw:
Hello,

J'arrive après la bataille mais concernant les deux points de la discussion :

button aligned on the left :

* effectivement pour les modales par ex. on est plus sur le fait de mettre les boutons sur la droite
mais là on est plus dans du scan d'info sur desktop c'est plus logique de rester dans le sens de la lecture et de le mettre ferré a gauche
* (il faudrait effectivement avoir une marge egale entre la marge gauche et la marge du bas)

variants :
je n’ai pas d’autres couleurs que celles présentées dans la capture d’écran, je dirais donc que c’est un peu prématuré pour le moment.

PR #958, opened 2019-05-02
Feat/multiplefield

GoOz:
- Fixed `cross-small` svg that wasn't add to the sprite correctly
- Expand touch area of the round button on touch devices
- Add `block` props (true by default) to `<Label />` component to allow inline display
- Added classes for inline fields (only CSS for now)
- Added classes for double input fields (only CSS for now)

### DIFF
--- a/react/Label/Readme.md
+++ b/react/Label/Readme.md
@@ -9,6 +9,17 @@
 </form>
 ```
 
+#### Inline Label (Labels are displayed `block` by default)
+
+```
+<form>
+  <div>
+    <Label htmlFor="idInput2" block={false}>This is an inline label</Label>
+    <Input id="idInput2" placeholder="Recherche" />
+  </div>
+</form>
+```
+
 #### Label when there's an error
 
 ```
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -1,5 +1,4 @@
 @require '../settings/icons'
-@require '../settings/breakpoints'
 @require '../utilities/display'
 @require '../utilities/text'
 @require '../tools/mixins'
@@ -355,7 +354,7 @@ $button--round
         height rem(10)
 
     // Expand clickable area for touch devices
-    +small-screen()
+    @media (pointer: coarse)
         &:after
             content ''
             position absolute
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -1,4 +1,5 @@
 @require '../settings/icons'
+@require '../settings/breakpoints'
 @require '../utilities/display'
 @require '../utilities/text'
 @require '../tools/mixins'
@@ -352,6 +353,16 @@ $button--round
     svg
         width rem(10)
         height rem(10)
+
+    // Expand clickable area for touch devices
+    +small-screen()
+        &:after
+            content ''
+            position absolute
+            top rem(-14)
+            right rem(-14)
+            bottom rem(-14)
+            left rem(-14)
 
 /*------------------------------------*\
   Action button
--- a/assets/icons/ui/cross-small.svg
+++ b/assets/icons/ui/cross-small.svg
@@ -1,6 +1,3 @@
 <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 12 12">
-    <defs>
-        <path id="cross-small" d="M6 4.59l2.3-2.3a1 1 0 0 1 1.4 1.42L7.42 6l2.3 2.3A1 1 0 1 1 8.3 9.7L6 7.42l-2.3 2.3A1 1 0 1 1 2.3 8.3L4.58 6l-2.3-2.3A1 1 0 0 1 3.7 2.3L6 4.58z"/>
-    </defs>
-<use xlink:href="#cross-small"/>
+  <path d="M6 4.59l2.3-2.3a1 1 0 0 1 1.4 1.42L7.42 6l2.3 2.3A1 1 0 1 1 8.3 9.7L6 7.42l-2.3 2.3A1 1 0 1 1 2.3 8.3L4.58 6l-2.3-2.3A1 1 0 0 1 3.7 2.3L6 4.58z"/>
 </svg>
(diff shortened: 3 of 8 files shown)


GoOz:
[(𝐒) Contacts - Form édition de contacts - révision UI](https://trello.com/c/U2TJBrbs/1116-%F0%9D%90%92-contacts-form-%C3%A9dition-de-contacts-r%C3%A9vision-ui)

Crash--:
We should add an exemple / test with `block `

stylus/components/button.styl
```diff
@@ -353,6 +354,16 @@ $button--round
...
 
+    // Expand clickable area for touch devices
+    +small-screen()
+        &:after
+            content ''
+            position absolute
+            top rem(-14)
+            right rem(-14)
+            bottom rem(-14)
+            left rem(-14)
```
y-lohse: Smart.

GoOz:
> Regarding the labels becoming `display: block`:
> 
>     * I think it's a breaking change ?

Shouldn't be. I made the props true by default.

>     * Do we need the `block` prop, when we can use a utility class to toggle the `display` value ?

That's what I thought first, but I couldn't overwrite it with utilities due to CSS order issue then I thought that would make sense to make prop out of it since there should be more cases of inline labels.

react/Label/Readme.md
```diff
@@ -9,6 +9,17 @@
...
 ```
 
+#### Inline Label (Labels are displayed `block` by default)
+
+```
+<form>
+  <div>
+    <Label htmlFor="idInput2" block={false}>This is an inline label</Label>
+    <Input id="idInput2" placeholder="Recherche" />
+    </div>
```
ptbrowne:
```suggestion
  </div>
```

ptbrowne:
> That's what I thought first, but I couldn't overwrite it with utilities due to CSS order issue

That's surprising, CSS utilities should be `important!` right ?

> then I thought that would make sense to make prop out of it since there should be more cases of inline labels.

Also, it rings true with https://gist.github.com/maxhoffmann/df09f8d27edcfdab1eaa1085671eae16#advice-on-component-variations

y-lohse:
> Shouldn't be. I made the props true by default.

Yes, but before this PR it was unspecified, so labels would be displayed inline until now, no?

(regarding the prop vs class, I'll read up on the article and comment again later)

GoOz:
> Yes, but before this PR it was unspecified, so labels would be displayed inline until now, no?

I'm not sure I understand your question or the issue here.

The prop didn't exist before, indeed, but since I'm adding it `true` by default, every `Label` already used anywhere would inherit this new prop with `true` by default, so every occurences would still be displayed block, unless the prop `block` is set to `false`. But maybe I'm missing a point here.

y-lohse:
Ok cleared it up, I missed [this line](https://github.com/cozy/cozy-ui/pull/958/files#diff-833b687e04ab70b5ce966ee49e85e438L230) in my initial review.

GoOz:
Is this good then?

y-lohse:
There's still https://github.com/cozy/cozy-ui/pull/958/files#r281072593

> That's surprising, CSS utilities should be important! right ?

Yeah although at this point not all of them are. I guess the displays aren't? I'd be curious to know more about the css ordering issue in this case though. There's a `display inline` that overwrites the `display block` ?

GoOz:
> There's still https://github.com/cozy/cozy-ui/pull/958/files#r281072593
> 
> > That's surprising, CSS utilities should be important! right ?
> 
> Yeah although at this point not all of them are. I guess the displays aren't? I'd be curious to know more about the css ordering issue in this case though. There's a `display inline` that overwrites the `display block` ?

Oups didn't see those.

To be specific, at first I tried to add a `display: flex` with `u-flex` but it appeared that the styles for `c-label` were output after the utilities.
As for the `!important` I still like to think that it shouldn't be by default. It's very hard to overwrite those styles. It makes sense for spacing utilities but not so much with the others.

ptbrowne:
> It's very hard to overwrite those styles

I don't understand, a utility class is not meant to be overwritten, it has to be removed if it is to be overwritten. This is why it should be important.

GoOz:
Well yes but it should be possible to overwrite them with other utilities in some use cases (e.g. Media Queries), which means there should be a precise and respected CSS order to maintain.

Although if we add `!important` on every utility classes, then we shouldn't build our React components with them or it will be absolutely still and impossible to custom.

The way I see it there's two types of utility classes:
- the ones that are meant to build things from scratch (like Tachyon)
- the ones that are meant to custom/overwrite existing styles (like what we're doing)

If we choose to make those utilities for custom only with `!important` then there's some cleaning up to do because there are some PR that used utility classes for building components.

GoOz:
[(𝐒) Contacts - Form édition de contacts - révision UI](https://trello.com/c/pUe8ei8H/1730-%F0%9D%90%92-contacts-form-%C3%A9dition-de-contacts-r%C3%A9vision-ui)

y-lohse:
> To be specific, at first I tried to add a display: flex with u-flex but it appeared that the styles for c-label were output after the utilities.

Out of curiosity, in what output? The transpiled stylesheet, or somewhere else? The alternative is ok for me, but it would be great to keep listing the problems we have with css style order, in as much details as possible. It's a problem we have trouble understanding, so the more information we can gather about it, the better. And to be honest I'm more worried about that than whether we should use `!important` or not to sidestep the issue  :)

______

Regarding `!important` or not on our utility classes: I rewrote this message 3 times and changed my mind… it's a tricky subject, but I think we should use `!important` everywhere. As @GoOz points out, the problem is that

1. We won't be able to override a utility class with a regular css class — which I think is ok
2. We won't be able to override a utility class with another utility class — might lead to problems in some cases

GoOz:
> Out of curiosity, in what output?

Yes, it was on Contacts so transpiled.

ptbrowne:
> Well yes but it should be possible to overwrite them with other utilities in some use cases (e.g. Media Queries), which means there should be a precise and respected CSS order to maintain.

For the media queries, I think there not be a problem because we control the order precisely no ?

```
.u-toto {
  background: red !important;
}

@media (screen-width > 500px) {
  .u-toto {
    background: blue !important;
 }
}
```

The problem for me is the order inter file.

PR #959, opened 2019-05-02
feat: Add box-sizing utils

GoOz:
Added a utility class to set the `box-sizing` property to `border-box`

### DIFF
--- a/stylus/utilities/boxsizing.styl
+++ b/stylus/utilities/boxsizing.styl
@@ -1,0 +1,10 @@
+@require '../tools/mixins'
+/*------------------------------------*\
+  Box-sizing utility
+\*------------------------------------*/
+
+$box-sizing-border
+    box-sizing border-box
+
+// Global classes
+global('.u-bxz', $box-sizing-border)
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -1321,6 +1321,19 @@ Display an chip that represents complex identity
 */
 
 /*
+    Box-sizing
+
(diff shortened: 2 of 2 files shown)


stylus/utilities/boxsizing.styl
```diff
@@ -0,0 +1,10 @@
+@require '../tools/mixins'
+/*------------------------------------*\
+  Box-sizing utility
+\*------------------------------------*/
+
+$box-sizing-border
+    box-sizing border-box
+
+// Global classes
+global('.u-bxz', $box-sizing-border)
```
y-lohse: I can't find where the class name comes from. Is [this](https://acss.io/guides/helper-classes.html) the reference?

Again, `bxz` for `box-sizing border-box` is super unintuitive for me, but if it's a convention…

GoOz: No, [this is](https://docs.emmet.io/cheat-sheet/) the reference.

I do agree though. This is not really intuitive. So I was wondering if it would make sense to have like plain explicit classes and shortcut classes ? Like `u-boxsizing-border` & `u-bxz` that would do the same thing, just like an alias. 
In terms of CSS weight, I don't think this would be too much since there's no duplication.

But maybe I'm overthinking it…

y-lohse: Thanks for the link! We can keep it as it is then :)

We might add "explicit" names at some point, or maybe as an optional thing. But from experience it's not really required: at the beginning you have to look it up anyway, and after a while you remember the shortcuts.

GoOz: That is indeed why I know most of Emmet's ;)

PR #962, opened 2019-05-06
Document Viewer

y-lohse:
This PR takes the `Viewer` component from cozy-drive and makes it available as a cozy-ui component.

Still missing:

- ~~More unit tests~~
- A wrapper component that accepts a list of `io.cozy.files` ids as prop, for easier integration **will be another PR**
- ~~Making the fallback viewer extendable / replacable by the host app~~

The goal of this PR is to make the component easy to use in other apps, with a good API. We're aware that some of the internal code of the viewer is not ideal, but fixing all these problems is beyond the scope of this PR.

edit: [Online docs preview](https://yannick-lohse.fr/cozy-ui/react/#viewer)


y-lohse:
Here's a quick usage example, essentially copied from the doc (that I haven't published yet 😅)

```
import { Viewer, Overlay } from 'cozy-ui/react'

const files = [
  {
    _id: 'audio',
    class: 'audio',
    name: 'sample.mp3',
    mime: 'audio/mp3'
  },
  {
    _id: 'pdf',
    class: 'pdf',
    name: 'demo.pdf',
    mime: 'application/pdf'
  },
  {
    _id: 'text',
    class: 'text',
    name: 'demo.txt',
    mime: 'text/plain'
  },
  {
    _id: 'image',
    class: 'image',
    name: 'demo.jpg',
    mime: 'image/jpg'
  },
  {
    _id: 'none',
    class: 'unknown',
    name: 'Unsupported file type',
    mime: '???/???'
  }
];

initialState = {
  viewerOpened: false,
  currentFileIndex: 0
};

const openViewer = () => setState({ viewerOpened: true });
const closeViewer = () => setState({ viewerOpened: false });
const onFileChange = (file, nextIndex) => setState({ currentFileIndex: nextIndex });

<div>
    <button onClick={openViewer}>
      Open viewer
    </button>
    { state.viewerOpened &&
      <Overlay>
        <Viewer
          files={files}
          currentIndex={state.currentFileIndex}
          onClose={closeViewer}
          onChange={onFileChange}
          showNavigation={false}
          showToolbar={true}
        />
      </Overlay>
    }
</div>
```

y-lohse:
There's now a demo available here : https://yannick-lohse.fr/cozy-ui/react/#viewer

ptbrowne:
Nice music ! However in the demo, I could not see the cross, it seems to be invisible. Is it a CSS order issue ?

assets/icons/illus/magic-trick.svg
```diff
@@ -0,0 +1,88 @@
...
+         id="path13"
+         style="fill:#ffffff;fill-opacity:1" />
+      <path
+         d="M23.5683158,4.45357895 L22.4466316,4.07968421 L22.0727368,2.95831579 C21.9868421,2.70031579 21.7455789,2.52631579 21.4736842,2.52631579 C21.2017895,2.52631579 20.9605263,2.70031579 20.8746316,2.95831579 L20.5007368,4.07968421 L19.3790526,4.45357895 C19.1213684,4.53947368 18.9473684,4.78073684 18.9473684,5.05263158 C18.9473684,5.32452632 19.1213684,5.56578947 19.3790526,5.65168421 L20.5007368,6.02557895 L20.8746316,7.14694737 C20.9605263,7.40494737 21.2017895,7.57894737 21.4736842,7.57894737 C21.7455789,7.57894737 21.9868421,7.40494737 22.0727368,7.14694737 L22.4466316,6.02557895 L23.5683158,5.65168421 C23.826,5.56578947 24,5.32452632 24,5.05263158 C24,4.78073684 23.826,4.53947368 23.5683158,4.45357895 Z"
+         id="path15"
+         style="fill:#ffffff;fill-opacity:1" />
+    </g>
+  </g>
+</svg>
```
ptbrowne: I think it can be `svgo`ed :)

Crash--:
> On the documentation preview, everything works well for me, except I don't see the navigation arrows. Navigation using my keyboard's arrows works, but I can't navigate using the mouse.
```
 showNavigation={false}
```

Maybe you should showNavigation={true} for the exemple

y-lohse:
> However in the demo, I could not see the cross, it seems to be invisible. Is it a CSS order issue ?

Mmh that's strange, I can definitely see the cross, in chrome and firefox. Can you still reproduce the problem? Is it maybe a caching issue on your end?

Crash--:
If you cursor is outside of the window you don't see the "Download" and "cross".

ptbrowne:
> Can you still reproduce the problem?

I opened the example via the one example per page link on styleguidist, the IconSprite was not loaded. It can be ignored.

package.json
```diff
@@ -103,18 +105,21 @@
...
     "travis-deploy-once": "5.0.10",
-    "url-loader": "^1.1.2",
+    "url-loader": "1.1.2",
     "webpack": "4.27.1"
   },
   "dependencies": {
     "@babel/runtime": "^7.3.4",
     "body-scroll-lock": "^2.5.8",
     "classnames": "^2.2.5",
+    "cozy-client": "6.26.0",
```
Crash--: Should not be dep, right? It should be a peerDep and devDep ?

y-lohse: No I don't think so. [`cozy-client` is really used in `cozy-ui` now](https://github.com/cozy/cozy-ui/pull/962/files#diff-670d8a9e3dcb209b4f42fe9dd43ca070R3) so it's a real dependency.

Crash--: And? 

We use cozy-client in Harvest, but cozy-client is a peerdep. This is the responsibility of the App to require the lib. I though we already talked about our peerDep for our internal libs. Nop?

y-lohse: >  I though we already talked about our peerDep for our internal libs.

Was there a consensus on that?

I have to re-read the docs every time, but I'm fairly sure perDependency is wrong in this case. The viewer is not just "hooking into" cozy-client, it's not a plugin for cozy-client. It literally needs cozy-client to run.

It's confusing because the app has to initiate cozy-client and insert it into the context. So in that sense, both the app and the viewer need cozy-client. But from what I read peerDependencies mean that only the host actually *uses* the dependency, which isn't the case here.

PR #964, opened 2019-05-07
fix(PushClientButton): Update asset path

GoOz:
Fix icon path so the component can work in a transpiled way or a non-transpiled way.


GoOz:
[(𝐒) [cozy-ui] Replace remaining embedurl (tous sauf le formulaire)](https://trello.com/c/zD7HtKuc/1284-%F0%9D%90%92-cozy-ui-replace-remaining-embedurl-tous-sauf-le-formulaire)

PR #965, opened 2019-05-07
fix: New version

Crash--:
I clicked on rebase from the GH interface :'(


GoOz:
huh? I don't get it. Nothing changed.

Crash--:
In my previous PR, I squashed my commits and I didn't see that the generated commit where not "sementic release". So I didn't get my version. 

Had to force push & merge this new commit to have the new version :(

PR #968, opened 2019-05-09
fix: Remove margin-top on text when no title

GoOz:
Before:
Screenshot 2019-05-09 17 17 37

After:
Screenshot 2019-05-09 17 22 13

Preview: https://gooz.github.io/cozy-ui/react/#infos

### DIFF
--- a/react/Infos/index.jsx
+++ b/react/Infos/index.jsx
@@ -14,7 +14,8 @@ const Infos = ({ actionButton, icon, isImportant, text, className, title }) => {
         'u-maw-6',
         'u-mih-2',
         'u-bdrs-4',
-        'u-p-1',
+        'u-ph-1',
+        'u-pt-1',
         'u-bg-paleGrey',
         'u-ta-left',
         {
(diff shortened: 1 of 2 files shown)


ptbrowne:
It feels a bit like a hack to need JS to do this. I read somewhere that to avoid this kind of problem, some people used only vertical padding in one direction. This way it is always the responsibility of the top element to add the margin-bottom.

Found the article : https://csswizardry.com/2012/06/single-direction-margin-declarations/

Should we start doing that ?

y-lohse:
> I read somewhere that to avoid this kind of problem, some people used only spacing in one direction. This way it is always the responsibility of the top element to add the margin-bottom.

I haven't looked at this fix yet, but… [yes](https://github.com/cozy/cozy-ui/pull/953#pullrequestreview-232198039).

y-lohse:
Also, can we please have the preview somewhere so we can play with it?

GoOz:
> It feels a bit like a hack to need JS to do this. I read somewhere that to avoid this kind of problem, some people used only vertical padding in one direction. This way it is always the responsibility of the top element to add the margin-bottom.

I tend to do that myself, I guess it's doable without too much effort.

GoOz:
> Also, can we please have the preview somewhere so we can play with it?

https://gooz.github.io/cozy-ui/react/#infos

y-lohse:
> I tend to do that myself, I guess it's doable without too much effort.

Honestly I think it would be great for this component. Things look ok now, but I wasn't fond of the first conditional class, so building on top of that doesn't feel safe. It's also the second fix in short order for the same component, so a css do-over if it doesn't take too long would be helpful.

GoOz:
As expected, it wasn't too much effort. I also updated the preview.

PR #971, opened 2019-05-13
fix: PDF import

Crash--:
Cozy-UI should be tooling agnostic, since it can be consumed in several projects either with webpack but also with parcel or else. Also, having a strong dependency to "webpack loader" inside the source code, is not a good idea since, for exemple, Jest can't resolve it. 

The idea is : 
- If an application wants to use the Viewer, then it has to create an alias for react-pdf in their bundler config: 
```
react-pdf : react-pdf/dist/entry.webpack.js
```

I didn't try yet, but should work. Maybe @y-lohse you can test on your current project ? If it works, I'll update the Viewer's readme

Added after opening this PR : 
- Expose React-PDF only if we are on a React app ?


y-lohse:
Yep, I confirm this works for the jest-problems. LGTM, just need the readme update.

GoOz:
Seems to work on Home as well. 👍

Crash--:
@y-lohse @ptbrowne I've updated the PR

react/Viewer/Readme.md
```diff
@@ -1,8 +1,12 @@
+Note :
+
+- The `Viewer` can be used only in a `React Application`. You can't use it with `Preact`.
+- If you want to use the `Viewer` in your app and if you use Webpack, you have to add an alias in your configuration :`react-pdf : react-pdf/dist/entry.webpack.js`
```
y-lohse:
```suggestion
- If you want to use the `Viewer` in your app and if you use Webpack, you should add the following alias in your configuration : `react-pdf : react-pdf/dist/entry.webpack.js`. This will make sure a web-worker is used to render PDF files.
```

react/Viewer/Readme.md
```diff
@@ -1,7 +1,7 @@
 Note :
 
 - The `Viewer` can be used only in a `React Application`. You can't use it with `Preact`.
-- If you want to use the `Viewer` in your app and if you use Webpack, you have to add an alias in your configuration :`react-pdf : react-pdf/dist/entry.webpack.js`
+- If you want to use the `Viewer` in your app and if you use Webpack, you should add the following alias in your configuration : `react-pdf : react-pdf/dist/entry.webpack.js`. This will make sure a web-worker is used to render PDF files.
```
ptbrowne: A patch code block makes it clearer IMHO.

```patch
+ resolve: {
+   alias: {
+     'react-pdf' : 'react-pdf/dist/entry.webpack.js'
+   }
+ }
```

PR #973, opened 2019-05-13
Add BarButton

gregorylegarec:
The BarButton is targeted to be injected into CozyBar.

A first usage is to offer a back button on mobile.

```jsx
const { BarLeft } = cozy.bar
return (
  <BarLeft>
    <BaButton icon="back" />
  </BarLeft>
)
```

[Styleguidist](https://gregorylegarec.github.io/cozy-ui/react/#!/BarBackButton)


react/BarBackButton/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+  render() {
+    const { disabled, href, onClick } = this.props
+    return (
+      <MaybeLink
+        className={styles['barleft']}
+        href={!disabled && href}
+        onClick={!disabled && onClick}
+      >
+        <Icon icon="back" className={disabled ? 'u-silver' : 'u-coolGrey'} />
+      </MaybeLink>
```
CPatchane: It looks like a special icon button to me but it seems not specific to the bar, so I am not comfortable with the naming `BarBackButton`

ptbrowne: The u-silver, u-coolgrey behavior is specific to BarBackButton. Do you see another instance in Cozy that has this behavior ? I prefer not to abstract a behavior early here. I don't see what we would gain.

ptbrowne: Maybe BarButton instead of IconButton ?

<BarButton icon='back' />
<BarButton icon='sandwich' />

CPatchane: At least `BarButton` is still least specific than `BarBackButton` so 👍

gregorylegarec: So let's go for a `<BackButtion />` I assume that `icon` should now be a prop, or should we wait for the express need ?

CPatchane: It should be a prop like in @ptbrowne example

gregorylegarec:
Updated ✅

react/BarButton/Readme.md
```diff
@@ -0,0 +1,48 @@
...
+```jsx static
+  const { BarLeft } = cozy.bar
+  return (
+    <BarLeft>
+      <BarButton />
+    </BarLeft>
+    )
+```
+
+## Props
```
y-lohse: Note that you can document the prop directly with comments https://react-styleguidist.js.org/docs/documenting.html#code-comments-and-proptypes

ptbrowne: Yes please.

react/BarButton/Readme.md
```diff
@@ -0,0 +1,48 @@
...
+<BarButton icon="cube" />
+```
+
+### `onClick`
+
+Add an `onClick` event handler on the component.
+
+```jsx
+<BarButton icon="gear" onClick={() => alert('BarButton has been clicked')} />
+```
```
ptbrowne: I think we should only document via Proptypes.

react/BarButton/Readme.md
```diff
@@ -0,0 +1,48 @@
...
+
+Used to render the internal `<Icon />` component.
+
+```jsx
+<BarButton icon="cube" />
+```
+
+### `onClick`
+
+Add an `onClick` event handler on the component.
```
ptbrowne:
```suggestion
Adds an `onClick` event handler on the component.
```

react/BarButton/Readme.md
gregorylegarec: Don't you think it's pretty convenient to have those examples rendered in styleguidist to test them manually ?

It allows us to get the component snapshoted by Jest as well.

ptbrowne: > It allows us to get the component snapshoted by Jest as well.

You're right, I thought that with `jsx` we had only the code and not the component displayed. I forgot that it was `static` that was doing this.

PR #975, opened 2019-05-13
fix: AppLinker use full path

Crash--:
Since we override this file, we remove relative path


Crash--:
Looks like it's not really needed. Closing for now.

PR #978, opened 2019-05-14
test: Testing withLocales

ptbrowne:
Adding test to withLocales HOC. Tests are not passing yet.

@y-lohse @gregorylegarec Can you take a look ? Is it a mistake from my end ?

### DIFF
--- a/react/I18n/withLocales.jsx
+++ b/react/I18n/withLocales.jsx
@@ -1,13 +1,25 @@
 import React from 'react'
 import { I18n, translate } from './'
+import omit from 'lodash/omit'
 
-const withLocales = locales => Component =>
-  translate()(props => {
-    return (
-      <I18n dictRequire={localeCode => locales[localeCode]} lang={props.lang}>
-        <Component {...props} />
-      </I18n>
-    )
-  })
+const withLocales = locales => Component => {
+  // The inner components needs to receive t
+  const Translated = translate()(Component)
+
+  class Wrapped extends React.Component {
+    render() {
+      // Do not pass t downwards
+      const { lang, ...rest } = omit(this.props, 't')
+      return (
+        <I18n dictRequire={localeCode => locales[localeCode]} lang={lang}>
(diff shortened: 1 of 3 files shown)


y-lohse:
Just had a quick look but i see nothing wrong, it should use the component locale…🤔

ptbrowne:
Found it :) https://github.com/cozy/cozy-ui/pull/978/commits/32dd4fb46b75d8ce4acddc633c36c403d1a39139

We were not destructuring the t that was received from the outer context, so it was passed downwards. Since there was no `translate()` on the inner component, it did not get its `t` from the local context but only from the props, thus it was accessing the `t` from the outer context.

It is working in our existing code since components relying on this are wrapped in a `translate()`. The question is, should withLocales automatically wrap the inner component in a `translate()` ? It makes sense to me.

I wish there was a simpler way. Please tell me if you see something.

y-lohse:
Wow, nice find.

So if I understand correctly, in the example, if we did something like `const TComponent = withLocales(componentLocales)(translate()(Component))` [here](https://github.com/cozy/cozy-ui/pull/978/files#diff-73502645e23dea3ebd655ae828204850R29) — it would work even without automatically wrapping the inner component in `translate`, right?

ptbrowne:
Yes :)

react/I18n/withLocales.jsx
```diff
@@ -1,13 +1,28 @@
...
-    )
-  })
+const withLocales = locales => Component => {
+  // The inner components needs to receive t
+  const Translated = translate()(Component)
+
+  class Wrapped extends React.Component {
+    render() {
+      // Destructure t so that it is not passed downwards
+      // eslint-disable-next-line no-unused-vars
```
CPatchane: It would be better to use lodash.omit to avoid this kind of "hack"?

ptbrowne: Done, thx

PR #980, opened 2019-05-14
feat(SelectBox): Add default white background 📝

gregorylegarec:
Pretty urgent, needed to update the cozy-store in our customizations.

Following 57d6c62342c515f1b883cd57337597fa7e120eba, we now need to set a default background for SelectBox when it is not disabled. (We specified a transparent background to React-Select to be able to style the disabled state.)


react/SelectBox/styles.styl
```diff
@@ -8,6 +8,10 @@
     height 0
     overflow hidden
 
+.select
+    /* Ensure the select box has background */
+    background-color white
```
edas: Shouldn't you use the stylus color variables ?

PR #983, opened 2019-05-15
feat: Update icons collection

GoOz:
Icons collection update & cleaning
[Preview](https://gooz.github.io/cozy-ui/react/#icon)

Tasks:
- [x] Add new icons
- [x] Removed duplicated icons
- [x] Removed obsolete icons 
- [x] Renaming icons
- [x] Reoder icons in Styleguidist
- [x] Migration Script
- [x] Fix tests

master: icons-sprite.js -> **57ko**
PR: icons-sprite.js -> **113ko**

We basically double the size.

**BREAKING CHANGE:**
Icons moved from `/ui/` folder to `/illus/`:
- `bottom-select`
- `check-white`
- `cozy`
- `dash-white`
- `top-select`

Icons moved from `/illus/` folder to `/ui/`:
- `magic-trick`

Icons removed:
- `cross-white`
- `dots-white`
- `share-grey08`
- `arrow-left`, use `previous` instead
- `arrow-right`, use `next` instead
- `arrow`, use `dropdown` instead
- `back`, use `left` instead
- `check-circleless`, use `check` instead
- `cozy-negative`, use `cloud` instead
- `delete`, use `trash` instead
- `destroy`, use `trash` instead
- `display`, use `eye` instead
- `file-new`, use `file-add` instead
- `forward`, use `right` instead
- `hide`, use `eye-closed` instead
- `mail`, use `email` instead
- `small-arrow`, use `bottom` instead

Icons added:
- `device-phone`
- `device-tablet`
- `google`
- `answer`
- `apple`
- `archive`
- `attachment`
- `attention`
- `bank`
- `credit-card-add`
- `credit-card`
- `bell`
- `browser-brave`
- `browser-chrome`
- `browser-duckduckgo`
- `browser-edge`
- `browser-firefox`
- `browser-ie`
- `browser-opera`
- `browser-safari`
- `burger`
- `camera`
- `stack`
- `check-list`
- `circle-filled`
- `collect`
- `comment`
- `company`
- `compass`
- `contract`
- `contrast`
- `cloud`
- `cloud-happy`
- `movement-in`
- `crop`
- `dashboard`
- `data-control`
- `movement-out`
- `devices`
- `down`
- `dropdown-close`
- `dropdown-open`
- `dropdown`
- `dropup`
- `email`
- `eu`
- `euro`
- `file-new`
- `filter`
- `fingerprint`
- `flag-outlined`
- `flag`
- `flash-auto`
- `flashlight`
- `folder-add`
- `forbidden`
- `from-user`
- `globe`
- `pie-chart`
- `grid`
- `group-list`
- `groups`
- `heart`
- `history`
- `home`
- `info-outlined`
- `key`
- `laptop`
- `lightbulb`
- `link`
- `list`
- `location`
- `logout`
- `magic-trick`
- `magnet`
- `magnifier`
- `merge`
- `multi-files`
- `music`
- `new`
- `email-notification`
- `offline`
- `online`
- `palette`
- `percent-circle`
- `personal-data`
- `phone`
- `pin`
- `printer`
- `repare`
- `rise`
- `rotate-left`
- `rotate-right`
- `select-all`
- `setting`
- `share-circle`
- `sound`
- `star`
- `stats`
- `target`
- `telephone`
- `to-the-cloud`
- `trophy`
- `unlink`
- `unlock`
- `up`
- `videos`
- `wallet-add`
- `wallet`


GoOz:
[(𝐒) [cozy-ui] Avoir l'ensemble de nos icônes dans cozy-ui](https://trello.com/c/CYSqIDWb/1286-%F0%9D%90%92-cozy-ui-avoir-lensemble-de-nos-ic%C3%B4nes-dans-cozy-ui)

ptbrowne:
What is the reason behind the breaking change for icons that moved ? Can we just keep them for a moment ? Was there a deprecation notice ?

GoOz:
> What is the reason behind the breaking change for icons that moved ? Can we just keep them for a moment ? Was there a deprecation notice ?

Mostly because of thre removed ones.
The moved ones are mostly old school `embedurl` icons I didn't have time to track over the repos yet, I'm guessing some of them will disappear soon.

I don't know about the deprecation notice, this is a pretty straighforward fix (don't use this string, use this one, not really a refactoring material). Also I already double the size of the sprite, I don't really want to keep even more deprecated icons.

y-lohse:
Ok please chill everyone :). Thanks a lot for the initial work @GoOz , it will be great to finally have a complete icon sprite. There are some more thing to do on this PR, including having a transition path for everyone that is as easy as possible. More on this later.

y-lohse:
I think there are 3 main points we need to discuss before merging this:

1. Icon names
2. Migration path
3. Sprite size

I'll start with the icon names in this comment, i'll post about the other ones afterwards. We'll want to avoid updating the sprite too much after this PR, so let's take our time :)

- Would it be possible to group icons by theme in the preview? Eg. all arrows together, all circle icons together, etc. This will make the review easier.
- `cozy-laugh` and `cozy-negative` look very similar, do we want both? Same for `warning` and `attention`.
- Some icons have literal names (eye, eye closed, etc) while others have abstract names (credit and debit, data-control…). I think we should stay consistent, and I personally tend to prefer literal names.
- We have top, bottom and… back? that have the same style. Wouldn't top, bottom, left and right make more sense?
- `album` -> `album-add` vs `wallet` -> `wallet-plus` vs `file` -> `file-new` vs `folder` -> `folder-add` and maybe others — Not sure what the best choice is, but let's settle for one
- `percent` should probably be `percent-circle` (let's not shoot ourself in the foot again 😁). Same for `warn`
- On the other hand `graph-circle` is probably just `graph`, or `pie-chart` ?

There are probably some more names that can be more sensible. Can you do a round of editing on that and make sure everything is consistent?

y-lohse:
Now for the migration — even if the changes are relatively straight forward, we don't want to put people in the situation where they *have* to deal with it because they urgently need another recent cozy-ui feature.  
I also think we should avoid opening a migration PR in every repo, publishing the new UI version, and then merging all these requests. We now have too many apps for that.

For the icons that are used with the `Icon` component, we can implement an automatic migration. Check if the icon provided is one of those that needs an update, and if so, log a message and use the newer version. We can leave this in for a few weeks, and it has no impact on the sprite size.

The problem is for direct icon imports, eg `import MyIcon from 'cozy-ui/icons/ui/top-select.svg'` or whatever. Can we check if we have a lot of those across our apps? Even if we leave the deprecated ones, do we have a way to print a deprecation notice when they are required?

y-lohse:
I think we can talk about the bundle size more in depth later, but here are my initial thoughts.

60kb, white not insignificant, is not a blocker for me. Ideally, the sprite would only contain the icons used by an app, and would do detect those icons automatically. Alternatively the app can list the icons it needs… but that's a lot less user friendly.

Finally, we could potentially not include the illustration in the sprite. Those are the largest files, and probably the least used. They could be in a separate sprite, or we could force manual imports for them.

GoOz:
> I'll start with the icon names in this comment, i'll post about the other ones afterwards. We'll want to avoid updating the sprite too much after this PR, so let's take our time :)
> 
>     * Would it be possible to group icons by theme in the preview? Eg. all arrows together, all circle icons together, etc. This will make the review easier.
> 

Yep I thought about that but basically mean a much larger breaking change, anyhow, while we're at it, why not? I'm hugely in favour of that.

>     * `cozy-laugh` and `cozy-negative` look very similar, do we want both? Same for `warning` and `attention`.

Those two are really different though. The smile is not the same. It's not really obvious when the icon is 16px or less, but it really is. Although, I'm was not convinced by the name `cozy-negative` when it poped up and still not convinced.

>     * Some icons have literal names (eye, eye closed, etc) while others have abstract names (credit and debit, data-control…). I think we should stay consistent, and I personally tend to prefer literal names.

I tend to agree.

>     * We have top, bottom and… back? that have the same style. Wouldn't top, bottom, left and right make more sense?

Indeed, missed that one.

>     * `album` -> `album-add` vs `wallet` -> `wallet-plus` vs `file` -> `file-new` vs `folder` -> `folder-add` and maybe others — Not sure what the best choice is, but let's settle for one

👍

>     * `percent` should probably be `percent-circle` (let's not shoot ourself in the foot again 😁). Same for `warn`

haha yes.

>     * On the other hand `graph-circle` is probably just `graph`, or `pie-chart` ?

👍

> There are probably some more names that can be more sensible. Can you do a round of editing on that and make sure everything is consistent?

OK

GoOz:
> For the icons that are used with the `Icon` component, we can implement an automatic migration. Check if the icon provided is one of those that needs an update, and if so, log a message and use the newer version. We can leave this in for a few weeks, and it has no impact on the sprite size.

Ok I don't know how to do that, that seems a cool way to do it.

> The problem is for direct icon imports, eg `import MyIcon from 'cozy-ui/icons/ui/top-select.svg'` or whatever. Can we check if we have a lot of those across our apps? Even if we leave the deprecated ones, do we have a way to print a deprecation notice when they are required?

I have in mind the obvious ones which are the form elements (checkbox, radio, select, selectbox), some others I don't really know it there are still used such as the `cross-white` & `share-grey08` and I must investigate. I didn't clean those because I didn't take the time to look for it yet.

GoOz:
> 60kb, white not insignificant, is not a blocker for me. Ideally, the sprite would only contain the icons used by an app, and would do detect those icons automatically. Alternatively the app can list the icons it needs… but that's a lot less user friendly.

I'm confused about something, the sprite being transpiled before, isn't it an issue for customizing the sprite according to the need of the app?

> Finally, we could potentially not include the illustration in the sprite. Those are the largest files, and probably the least used. They could be in a separate sprite, or we could force manual imports for them.

I thought about it too. I think there should be a clean up first (as said before I'm not sure there all needed anyway) but in the end I'm not convinced it would be a big size improvement, it will but not so much. We'll see :P

y-lohse:
> Yep I thought about that but basically mean a much larger breaking change, anyhow, while we're at it, why not? I'm hugely in favour of that.

I don't understand why this would be a breaking change? I basically meant we could change the order of the icons in the demo [here](https://github.com/cozy/cozy-ui/pull/983/files#diff-6b1619e35d7df6756278666dee981e96R32), so it wouldn't have any effect outside of the demo.

> Those two are really different though. The smile is not the same. It's not really obvious when the icon is 16px or less, but it really is. Although, I'm was not convinced by the name cozy-negative when it poped up and still not convinced.

I mean I saw the difference, but as you say, at 16px it's really hard to tell. And at bigger sizes… one icon is smiling, the other is grinning. Are we really sure we need that much nuance?  
But yes if we do, cozy-negative doesn't really make sense as a name for a smiling icon.

ptbrowne:
> I'm confused about something, the sprite being transpiled before, isn't it an issue for customizing the sprite according to the need of the app?

You're right, the Sprite system is not fit if apps need only to import the Icons they want. It was a temporary system mainly to have a transpiled cozy-ui up and running quickly. This decision should be revisited.

GoOz:
> > Yep I thought about that but basically mean a much larger breaking change, anyhow, while we're at it, why not? I'm hugely in favour of that.
> 
> I don't understand why this would be a breaking change? I basically meant we could change the order of the icons in the demo [here](https://github.com/cozy/cozy-ui/pull/983/files#diff-6b1619e35d7df6756278666dee981e96R32), so it wouldn't have any effect outside of the demo.

Sorry I misread. I actually was thinking renaming them all with some pattern like `arrow-left`, `arrow-top` etc… but that would be a big breaking change.
Indeed, the preview could be rearranged.

> I mean I saw the difference, but as you say, at 16px it's really hard to tell. And at bigger sizes… one icon is smiling, the other is grinning. Are we really sure we need that much nuance?

Seems like it, according to designers, yes.

ptbrowne:
I see in #987 that

> With previous instead of arrow-left. There is a duplication of those two icons.
next& previous will stay when arrow-right & arrow-left will be removed.

I thought that one idea to consolidate icon naming was to use "literal names". Arrow-left and arrow-right seem to be the literal names.

ptbrowne:
> For the icons that are used with the Icon component, we can implement an automatic migration. Check if the icon provided is one of those that needs an update, and if so, log a message and use the newer version. We can leave this in for a few weeks, and it has no impact on the sprite size.

I think it would be better if we would provide a codemod to apply to our repos to automatically change the calls to `<Icon>` and the `import`s. My experience with deprecation warnings is that they are not taken into account.

https://astexplorer.net/#/gist/c78b1c9f816b0237c2a2a4875c82a33c/7a9aa4d1a3598626ac78d8cee8607acd544e2785

y-lohse:
>My experience with deprecation warnings is that they are not taken into account. I think it would be better if we would provide a codemod to apply to our repos to automatically change the calls to <Icon> and the imports.

Yeah if we can't avoid breaking changes we'll probably do that.

GoOz:
> I see in #987 that
> 
> > With previous instead of arrow-left. There is a duplication of those two icons.
> > next& previous will stay when arrow-right & arrow-left will be removed.
> 
> I thought that one idea to consolidate icon naming was to use "literal names". Arrow-left and arrow-right seem to be the literal names.

Well, I'm open to ideas, but… if you take a look a the reordered preview I updated this morning, you'll see that there's a lot of different arrows and since those two exist only in right and left directions because they're only used for a next and previous purpose, I figured it is still a better way to distinguish them from the others.

ptbrowne:
> since those two exist only in right and left directions because they're only used for a next and previous purpose, I figured it is still a better way to distinguish them from the others.

Ah yes OK, sorry I did not see that :)

y-lohse:
To complete this card, we need to provide a codemod to upgrade direct icon imports, and change the Icon component to map old names to new names to minimize breaking changes.

drazik:
Here is a codemod forked from the one @ptbrowne posted that update imports and names: https://astexplorer.net/#/gist/40ca6fca485471480db0d9a9c84aec63/d9880d137f2a9f5a098d9fca32013399e2355eb9

ptbrowne:
I think the codemod can be integrated directly in the PR in a `codemods` folder ? So that when upgrading, we can do `jscodemod --parser babel -t node_modules/cozy-ui/codemods/new-icons.js`.

drazik:
> I think the codemod can be integrated directly in the PR in a `codemods` folder ? So that when upgrading, we can do `jscodemod --parser babel -t node_modules/cozy-ui/codemods/new-icons.js`.

I just did it

PR #986, opened 2019-05-16
chore: Document viewer usage with webpack

y-lohse:
Added a doc section about using the pdf viewer on a public page, and moved the webpack doc accordingly.

### DIFF
--- a/react/Viewer/Readme.md
+++ b/react/Viewer/Readme.md
@@ -1,18 +1,3 @@
-Note :
-
-- The `Viewer` can be used only in a `React Application`. You can't use it with `Preact`.
-- If you want to use the `Viewer` in your app and if you use Webpack, you should add the following alias in your configuration :
-
-```patch
-+ resolve: {
-+   alias: {
-+     'react-pdf' : 'react-pdf/dist/entry.webpack.js'
-+   }
-+ }
-```
-
-This will make sure a web-worker is used to render PDF files.
-
 The `Viewer` component can be used to display the content of various file types. In order to download and display the files, it will need a `cozy-client` instance in the React context.
 
 Once rendered, the `Viewer` will take up all the available space in it's container (using `position: absolute`). It can be paired with the `Overlay` component to take up the whole screen.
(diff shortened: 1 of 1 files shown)


y-lohse:
[Intégration du viewer cozy-ui dans drive/photos](https://trello.com/c/hDMhH2WW/1967-int%C3%A9gration-du-viewer-cozy-ui-dans-drive-photos)

react/Viewer/Readme.md
```diff
@@ -86,3 +71,43 @@ const onFileChange = (file, nextIndex) => setState({ currentFileIndex: nextIndex
...
 ```
+
+### Using a worker for pdfjs
+
+For performance reasons, it is important to use a web worker when showing PDF files in the viewer. If you use webpack, you should add the following alias in your configuration :
+
+```patch
++ resolve: {
++   alias: {
++     'react-pdf$' : 'react-pdf/dist/entry.webpack.js'
```
ptbrowne: To be sure : the dollar is there to prevent an infinite loop in the alias resolution ?

react/Viewer/Readme.md
```diff
@@ -86,3 +71,43 @@ const onFileChange = (file, nextIndex) => setState({ currentFileIndex: nextIndex
...
++   }
++ }
+```
+
+By default, the worker file will be created in a folder on a private route of your cozy application. If you need to display PDF files on a public page, you must explicitly load the web worker in your application like this:
```
ptbrowne: I don't understand this part ? Surely it does not mean react-router route ? In this case, I think it would be clearer if the doc would not mention route nor cozy application, it is just created in the build directory right ?

react/Viewer/Readme.md
y-lohse: Nope! Luckily there's no infinite loop, even without dollar sign. But it makes sure that only imports of `react-pdf` are redirected to `react-pdf/dist/entry.webpack.js` — so if you `import from 'react-pdf/dist/something.js'`, you get `something.js` and not `entry.webpack.js`.

react/Viewer/Readme.md
y-lohse: > the worker file is just created in the build directory right ?

Yes that is correct. I'll use your suggestion ;)

y-lohse: There, slightly tweaked. Let me know what you think :)

react/Viewer/Readme.md
```diff
@@ -86,3 +71,45 @@ const onFileChange = (file, nextIndex) => setState({ currentFileIndex: nextIndex
...
+
+```patch
++ resolve: {
++   alias: {
++     'react-pdf$' : 'react-pdf/dist/entry.webpack.js'
++   }
++ }
+```
+
+By design, JS created in the build directory are only accessible (ie served by the stack) if you are logged in. It also goes for the worker file created for pdfjs. If you need it on a public page, you must tell webpack to create it in a public folder, that will be served by the stack even if the user is not logged in. You can configure which folders of your build are served publicly in your manifest.webapp.
```
ptbrowne: Vu avec Yannick : When aliasing react-pdf with this webpack specific line, a worker JS file will be created.

Issue #988, opened 2019-05-16
Merge/clean InputPassword Field and Field with `side`


CPatchane:
We have two way to add extra information/action on a input label:

* By default, the Field with `password` as type will display an action button (`hideLabel/showLabel`) to show or to hide the password in the password. This feature has a [dedicated component and DOM](https://github.com/cozy/cozy-ui/blob/v19.35.3/react/Field/index.jsx#L41-L59) with dedicated styles.
<img width="592" alt="Screenshot 2019-05-16 at 15 31 23" src="https://user-images.githubusercontent.com/10224453/57857583-bb03fa80-77ef-11e9-8761-c8faae8db2d8.png">

* We can use the `side` prop of the Field to add extra information. This uses a dedicated DOM outline](https://github.com/cozy/cozy-ui/blob/v19.35.3/react/Field/index.jsx#L178-L186) and styles:
<img width="593" alt="Screenshot 2019-05-16 at 15 31 26" src="https://user-images.githubusercontent.com/10224453/57857654-de2eaa00-77ef-11e9-9e4a-9cbd60732b7b.png">

Since they have similarities, we should use the same HTML outline and styles as much as possible to have something more maintainable. Maybe this is a good time to use some UI locales instead of some API like `secondaryLabels`.

PR #989, opened 2019-05-16
Fix show/hide and side on Field component

CPatchane:
* Fix show/hide position on password Field
* Fix side position on Field
* Update the styleguidist example:
Screenshot 2019-05-16 at 15 21 56

Those are some quick fixes, we have an issue since the two (show/hide and side) has the same visual result but use different DOM outline and CSS rules, I opened an issue to clean this #988

### DIFF
--- a/react/Field/Readme.md
+++ b/react/Field/Readme.md
@@ -108,6 +108,22 @@ const options = [
 </form>
 ```
 
+#### Password field with show/hide button
+
+```
+<form>
+  <Field
+    id="idFieldPassword"
+    label="I'm a password label"
+    type="password"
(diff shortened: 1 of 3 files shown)


Crash--:
This is not a blocker yet, but I think that if we can have a real styleguidist instead of screenshot, it will be better ;)

CPatchane:
Yep, I pushed on https://cpatchane.github.io/cozy-ui/react/

gregorylegarec:
What caused the issue ?

CPatchane:
@gregorylegarec I don't know, it seems broken since many versions but the example was not correctly updated :/

PR #990, opened 2019-05-17
feat: Responsive position utils

GoOz:
Made the position utility classes responsive, it kinda make sense since `position` is clearly a property that could change on different viewport sizes… and… I actually need it :x


GoOz:
[🔵 🍊 Fix Spinners on Home + Settings + Banks](https://trello.com/c/vC3pMfcd/1401-%F0%9F%94%B5-%F0%9F%8D%8A-fix-spinners-on-home-settings-banks)

PR #992, opened 2019-05-20
fix: Don't use setState if unmouting

Crash--:
I had a few warnings about : 
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in AppIcon (created by AppLinker)
```

### DIFF
--- a/react/AppIcon/index.jsx
+++ b/react/AppIcon/index.jsx
@@ -22,9 +22,14 @@ export class AppIcon extends Component {
       icon: preloaded,
       status: preloaded ? DONE : FETCHING
     }
+    this.isUnmounting = false
   }
 
+  componentWillUnmount() {
+    this.isUnmounting = true
+  }
   componentDidMount() {
+    this.isUnmounting = false
     this.load()
   }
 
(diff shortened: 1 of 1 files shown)


Crash--:
[Console polluée dans drive](https://trello.com/c/T7jaiLPj/1938-console-pollu%C3%A9e-dans-drive)

react/AppIcon/index.jsx
```diff
@@ -23,7 +23,12 @@ export class AppIcon extends Component {
       status: preloaded ? DONE : FETCHING
     }
   }
-
+  componentWillMount() {
```
y-lohse: Sorry I missed this in a similar PR on drive, but `componentWillMount ` is deprecated and we shouldn't add any. I think setting the flag in the constructor instead is more a good alternative?

Crash--: This! Will use didmount instead. Constructor is not the right place since it’s not related to the life cycle and can already be instanciated

PR #993, opened 2019-05-21
feat: Add Standard breakpoints for utility classes

GoOz:
- [x] Change `tiny` breakpoint from 543px to 480px
- [x] Set the standard breakpoints for utility classes
  - 480px (tiny)
  - 768px (small)
  - 1023px (medium)

There now should be 4 versions of a utility class: `foo`, `foo-t`, `foo-s`, `foo-m` instead of the previous 7.

### DIFF
--- a/stylus/settings/breakpoints.styl
+++ b/stylus/settings/breakpoints.styl
@@ -8,21 +8,18 @@
 // variables
 
 BP-teeny       =  375
-BP-tiny        =  543
+BP-tiny        =  480
 BP-small       =  768
 BP-medium      = 1023
 BP-large       = 1200
 BP-extra-large = 1439
 
-// Breakpoints collection for utilities
+// Standard breakpoints collection for utilities
 breakpoints = {
     'none': '',
-    'teeny': 'tt',
     'tiny': 't',
     'small': 's',
     'medium': 'm',
-    'large': 'l',
-    'extra-large': 'xl'
 }
 
 /*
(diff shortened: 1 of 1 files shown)


GoOz:
[(𝐗𝐒) [UI] Responsive sur les class utilitaire](https://trello.com/c/0ApLy8a9/1292-%F0%9D%90%97%F0%9D%90%92-ui-responsive-sur-les-class-utilitaire)

Crash--:
It has to be flagged has breaking change since we change the current behavior, right?

GoOz:
@Crash-- Oh. for the removed versions? right. I'll change that.

Crash--:
> @Crash-- Oh. for the removed versions? right. I'll change that.

We also change the tiny breakpoint => we can also break the behavior of some apps

GoOz:
> > @Crash-- Oh. for the removed versions? right. I'll change that.
> 
> We also change the tiny breakpoint => we can also break the behavior of some apps

Don't think so, there's not so much occurrences of this breakpoint inside apps. I only found 5 (4 in Drive/Photos, 1 in Home) and we already looked at them with @joel-costa and it was OK to change without meaningful repercussions.

ptbrowne:
> Don't think so, there's not so much occurrences of this breakpoint inside apps.

I do not think we should base our definition of BREAKING CHANGE based on utilisation by the apps. As a developer, I want to be warned about those things, and it's OK if there is no problem if it's written breaking change but actually nothing broke for me.

Kudos though for checking the usage and also with Joel 👍

Crash--:
Breaking change is about changing the current behavior.

GoOz:
Go !

PR #997, opened 2019-05-22
feat: Checkbox component uses Icon

GoOz:
`<Checkbox />` component now uses `<Icon />` to display its icon, allowing to customize the color/icon/size icon for a checkbox.

Preview: https://gooz.github.io/cozy-ui/react/#checkbox

Related PR #983

### DIFF
--- a/react/Checkbox/index.jsx
+++ b/react/Checkbox/index.jsx
@@ -1,6 +1,8 @@
 import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
+import Icon from '../Icon'
+import palette from '../palette'
 import styles from './styles.styl'
 
 const Checkbox = props => {
@@ -19,6 +21,7 @@ const Checkbox = props => {
     <label
       className={cx(
         styles['c-input-checkbox'],
+        styles['c-input-checkbox--svg'],
         {
           [styles['is-error']]: error
         },
@@ -34,7 +37,17 @@ const Checkbox = props => {
         disabled={disabled}
         {...restProps}
       />
-      <span>{label || children}</span>
+      <span>
+        {/* TODO: Change 'check-circleless' to 'check' when icons are cleaned up */}
+        <Icon
+          icon={mixed ? 'dash' : 'check-circleless'}
+          color={palette['primaryContrastTextColor']}
+          className={styles['c-input-checkbox-icon']}
+          aria-hidden
+          focusable="false"
+        />
+        {label || children}
+      </span>
     </label>
   )
 }
--- a/react/Checkbox/styles.styl
+++ b/react/Checkbox/styles.styl
@@ -1,4 +1,28 @@
 @require '../../stylus/components/forms'
+@require '../../stylus/settings/palette'
 
 .c-input-checkbox
     @extend $checkbox
+
+// Cancel the default behaviour for input checkbox
+.c-input-checkbox--svg span::after
+    display none
+
+.c-input-checkbox-icon
+    position absolute
+    left 0
+    top rem(3)
+    box-sizing border-box
+    width checkbox-size
+    height checkbox-size
+    padding rem(2)
+    transition-duration .2s
+    transition-property opacity, transform
+
+.c-input-checkbox input:not(:checked) + span .c-input-checkbox-icon
+    opacity 0
+    transform scale(0)
+
+.c-input-checkbox input:checked + span .c-input-checkbox-icon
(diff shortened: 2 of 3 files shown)


GoOz:
[(𝐒) [UI] ETD j'ai exploré la possibilité de remplacer les embedurl pour le formulaire](https://trello.com/c/UlDjSJdA/1293-%F0%9D%90%92-ui-etd-jai-explor%C3%A9-la-possibilit%C3%A9-de-remplacer-les-embedurl-pour-le-formulaire)

y-lohse:
Haven't looked at the code yet, but on the demo, the dash inside the mixed checkbox is black instead of white.

Crash--:
Also, is it wanted that the checkbox is not properly vertically centered?

react/Checkbox/styles.styl
```diff
@@ -1,4 +1,28 @@
 @require '../../stylus/components/forms'
+@require '../../stylus/settings/palette'
 
 .c-input-checkbox
     @extend $checkbox
+
+// Cancel the default behaviour for input checkbox
+.c-input-checkbox--svg span::after
```
y-lohse: Why did we add the class if we cancel the effect? Is it because we want the background but take control over the tick / dash ?

react/Checkbox/index.jsx
```diff
@@ -34,7 +37,17 @@ const Checkbox = props => {
         disabled={disabled}
         {...restProps}
       />
-      <span>{label || children}</span>
+      <span>
+        {/* TODO: Change 'check-circleless' to 'check' when icons are cleaned up */}
```
ptbrowne: It would be nice to have the PR linked here so that if someone passes after, (s)he can click on the link, see that the PR is merged and do the change.

react/Checkbox/index.jsx
```diff
@@ -19,6 +21,7 @@ const Checkbox = props => {
     <label
       className={cx(
         styles['c-input-checkbox'],
+        styles['c-input-checkbox--svg'],
```
ptbrowne: What are the cases where we do not have the --svg modifier ?

y-lohse:
> allowing to customize the color/icon/size icon for a checkbox.

Another question — this is theoretical at the moment, right? To change the color we would need to change the palette, to change the icon we need to override the svg, and there's no way to change the size at the moment?

GoOz:
> Haven't looked at the code yet, but on the demo, the dash inside the mixed checkbox is black instead of white.

Damn… didn't see that coming. The `dash` icon is part of the icon you can't change the color (yet), it's fixed in #983. Also the `dash-white` is not yet part of the sprite either. I don't really know what to do here. Should we wait #983 to be ok & merged before merging this one?

> Also, is it wanted that the checkbox is not properly vertically centered?

It's the same as before.

react/Checkbox/styles.styl
GoOz: It's for backwards compatibility. 
I don't touch the "old way" so it's still working and add a class for the "new way" to cancel the old way.
When we decide to definitively remove the old way, we can remove this class and this chunk of css along with the css part of the old way.

I couldn't find a smoother way.

react/Checkbox/index.jsx
GoOz: You'll get what we have today : an icon with `embedurl`

GoOz:
> > allowing to customize the color/icon/size icon for a checkbox.
> 
> Another question — this is theoretical at the moment, right? To change the color we would need to change the palette, to change the icon we need to override the svg, and there's no way to change the size at the moment?

Yes. Theorical only. I may have been overzealous when mentionning the size, but I guess it is also easier to change the icon's size in this PR than it was before.

Crash--:
> It's the same as before.

Is it the desired behavior? If no, should we fix it?

GoOz:
> > It's the same as before.
> 
> Is it the desired behavior? If no, should we fix it?

I guess not. I don't think this fix should be in this PR though as it has nothing to do with the use of `Icon` instead of `embedurl`

Crash--:
got it, thanks. Issue here to not forget https://github.com/cozy/cozy-ui/issues/998 ;)

y-lohse:
> Also the dash-white is not yet part of the sprite either. I don't really know what to do here. Should we wait #983 to be ok & merged before merging this one?

I think #983 still requires some work, meanwhile a black dash can be considered a bug. I would be in favor of fixing that icon independently from #983.

PR #1003, opened 2019-05-24
Fix/breakpoint utility

GoOz:
Some revert, cherry-picking and amend commit msg to get a new major version

PR #1007, opened 2019-05-24
Fix/re apply standard bp

GoOz:
Re-applying Standard Breakpoints commit.

See #1003


y-lohse:
Note that this PR probably doesn't need the revert commit, but anyway. LGTM, just need to merge #1006 first.

GoOz:
@y-lohse Yeah I know it's because this branch is based on #1006
I'll rebase this one when the other is merged.

PR #1008, opened 2019-05-27
fix: Dash icon

GoOz:
Fixes the `dash` icon so you can change its color (it wasn't possible to this day without using `strike` intead of `fill` CSS property).

It was fixed in #983 but since this PR is not quite ready to be merged and we need this icon to work for #997 so here it goes.


y-lohse:
Changes LGTM — but just to make sure, is there a place where we might rely on the dash icon being dark? Like the legacy checkbox?

GoOz:
> Changes LGTM — but just to make sure, is there a place where we might rely on the dash icon being dark? Like the legacy checkbox?

Well… I only can find it in the `Viewer` (the one you just add to Cozy-UI recently) and actually right now the color of that dash icon in the `Viewer` is not right but close enough for not being spotted.
The only other need for dash icon is for checkboxes but that's not even the SVG right now.
I can't even think of a place where this icon would be used properly since none of our UI icons is actually black (#000) and black is its color right now.

Do you want a fallback anyway? I can, I'm just not sure it's relevant since this icon is bugged since forever.

y-lohse:
Nope that's ok then, I think this is a legitimate bugfix 👍

PR #1009, opened 2019-05-27
fix: Do not transpile spec files

ptbrowne:
Spec files should not be distributed and transpiled.


ptbrowne:
> I can't find a good reason to put this in the babel.config file.

I had forgotten about the babel config file, it makes more sense to put this there since there is already the examples that are ignored. Thanks for the suggestion.

PR #1011, opened 2019-05-27
Add sizes, variants and themes to Chip

drazik:
This PR introduces 3 new props for the `Chip` component:

* `size`: add the `small` size. Existing size is `normal`
* `variant`: add `outlined` and `dashed` variants. Existing styles are `normal` variant
* `theme`: add `primary` and `error` variant. Existing colors are `normal` variant

You can see all this on https://drazik.github.io/cozy-ui/react/#chip

### DIFF
--- a/react/Chip/Readme.md
+++ b/react/Chip/Readme.md
@@ -75,3 +75,25 @@ const ContactChip = ({ contact }) => (
   <Chip variant="dashed">This is a dashed Chip</Chip>
 </div>
 ```
+
+### Themes
+
+```
+<div>
+  <div>
+    <Chip theme="normal">This is a normal Chip (default)</Chip>
+    <Chip theme="primary">This is a primary Chip</Chip>
+    <Chip theme="error">This is an error Chip</Chip>
+  </div>
+  <div>
+    <Chip theme="normal" variant="outlined">This is a normal Chip (default)</Chip>
+    <Chip theme="primary" variant="outlined">This is a primary Chip</Chip>
+    <Chip theme="error" variant="outlined">This is an error Chip</Chip>
+  </div>
+  <div>
+    <Chip theme="normal" variant="dashed">This is a normal Chip (default)</Chip>
+    <Chip theme="primary" variant="dashed">This is a primary Chip</Chip>
+    <Chip theme="error" variant="dashed">This is an error Chip</Chip>
+  </div>
+</div>
+```
--- a/react/Chip/Readme.md
+++ b/react/Chip/Readme.md
@@ -51,7 +51,8 @@ const ContactChip = ({ contact }) => (
 
 ```
 <div>
-  <Chip component="button">This is a button</Chip>
+  <Chip component="button" onClick={() => alert('You clicked')}>This is a button</Chip>
+  <Chip component="button" disabled>This is a disabled button</Chip>
   <Chip component="span">This is a span</Chip>
 </div>
 ```
--- a/react/Chip/Readme.md
+++ b/react/Chip/Readme.md
@@ -56,3 +56,12 @@ const ContactChip = ({ contact }) => (
   <Chip component="span">This is a span</Chip>
 </div>
 ```
+
+### Sizes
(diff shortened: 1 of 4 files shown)


react/Chip/Readme.md
```diff
@@ -46,3 +46,76 @@ const ContactChip = ({ contact }) => (
   <Chip.Button disabled><Icon icon='back' /></Chip.Button>
 </div>
 ```
+
+### Specify a root component
```
ptbrowne: Specify underlying tag/component ?

This was the only example that at first glance was not completely clear to me.

react/Chip/index.jsx
```diff
@@ -3,30 +3,56 @@ import React from 'react'
...
           styles['c-chip'],
+          styles[`c-chip--${size}Size`],
+          styles[`c-chip--${variant}Variant`],
+          styles[`c-chip--${theme}Theme`],
           'u-breakword',
-          className,
-          this.constructor.className
+          {
+            [styles['c-chip--round']]: rounded,
+            'u-c-pointer': onClick && !disabled,
```
ptbrowne: I dont think we should use u-c-pointer here because it is an utility class that won't be overrided easily. Prefer not to use utility classes in library components.

stylus/components/chip.styl
```diff
@@ -1,26 +1,58 @@
...
 $round-chip
-    width $chip-height
+    width $chip-height-normal
     text-align center
     justify-content center
 
+$small-size-chip
+    height $chip-height-small
+    padding (($chip-height-small - $chip-font-size-small) / 2)
+    border-radius ($chip-height-small / 2)
```
ptbrowne: You can put 50% for all the chips right ?

stylus/components/chip.styl
```diff
@@ -1,26 +1,58 @@
...
+$normal-size-chip
+    height $chip-height-normal
+    padding (($chip-height-normal - $chip-font-size-normal) / 2)
+    border-radius ($chip-height-normal / 2)
+
+$outlined-variant-chip
+    border 1px solid
+    // important is here because this variant needs to have a transparent
+    // background, no matter with which theme it is used
+    background-color transparent !important // @stylint ignore
```
ptbrowne: I don't like the important here but I do not see an easy win.

stylus/components/chip.styl
```diff
@@ -1,26 +1,58 @@
...
+
+$dashed-variant-chip
+    border 1px dashed
+
+$normal-theme-chip
+    background var(--paleGrey)
```
ptbrowne: background-color ?

stylus/components/chip.styl
GoOz: Yeah, or even `100%`. Or anything above half the size, really.

stylus/components/chip.styl
GoOz: It's either that or increasing the specifity weight in `react/Chip/styles.styl` like you did L30 ( which I tend to prefer) or create a new theme… :/

stylus/components/chip.styl
```diff
@@ -1,26 +1,58 @@
...
-    height $chip-height
-    background var(--paleGrey)
-    padding (($chip-height - $chip-font-size) / 2)
     box-sizing border-box
     line-height 1
     display inline-flex
     align-items center
     margin-right rem(4)
     margin-bottom rem(4)
+    border-width 0
```
GoOz: Nit: why not just `border: 0` ?

stylus/components/chip.styl
drazik: Actually with `50%` or `100%` we will end up with something like this: https://codesandbox.io/s/zealous-sun-9g03o

GoOz: OH yeah, sorry, you're absolutely right, the thing I said is for round object… 🤦‍♂

ptbrowne: Same mistake 😇

PR #1014, opened 2019-05-28
Add Popup component

gregorylegarec:
[Styleguidist](https://gregorylegarec.github.io/cozy-ui/react/#!/Popup)

In Harvest we are needing a better popup component than PopupOpener may offer.
Our needs:
* onClose callback
* onMessage callback
* Open the modal with a Button (PopupOpener renders an anchor tag)

This PR offers a component separated from PopupOpener logic.

To do:

* [x] Ensure everything works fine in Harvest
* [x] Add README and examples
* [x] Add tests
* [ ] Integrate in PopupOpener

**BONUS**

* [ ] Add a render props on PopupOpener to render anchor tag only when no children is passed


react/Popup/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+  handleMessage(messageEvent) {
+    const { popup } = this.state
+    const { onMessage } = this.props
+    const isFromPopup = popup.location.origin === messageEvent.origin
+    if (isFromPopup && typeof onMessage === 'function') onMessage(messageEvent)
+  }
+
+  handleClose(popup) {
+    this.removeListeners(popup)
```
y-lohse: `this.removeListeners` doesn't seem to accept any parameters.

react/Popup/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+    const { onClose } = this.props
+    if (typeof onClose === 'function') onClose(popup)
+  }
+
+  /**
+   * Check if window is closing every 500ms
+   * @param  {Window} window
+   */
+  monitorClosing(popup) {
```
y-lohse: Wouldn't it be possible to do something like `newWindow.addEventListener('close', this. handleClose)` instead of polling?

react/Popup/index.jsx
```diff
@@ -0,0 +1,133 @@
...
+    : screen.width
+  var height = window.innerHeight
+    ? window.innerHeight
+    : document.documentElement.clientHeight
+    ? document.documentElement.clientHeight
+    : screen.height
+
+  var left = width / 2 - w / 2 + dualScreenLeft
+  var top = height / 2 - h / 2 + dualScreenTop
+  // need to be set here to get from the OAuth opener
```
y-lohse: I think that comment has lost it's context? Otherwise what's the OAuth opener here?

gregorylegarec:
This PR will have a little update after will be merged. Please do not spent too much time on reviewing it until then.

gregorylegarec:
Finally updated. I think we should delay integration in `PopupOpener` in another PR.

react/Popup/index.jsx
gregorylegarec: As far as I know it the best way to monitor popup with cross-domain urls.

gregorylegarec:
Thanks for your opinion about storing `popup` into the state. I was also asking myself about it 👍

gregorylegarec:
@drazik You're absolutely right about `handleUrlChange`, but I realize that we should probably name it `handleLoadStart` instead. WDYT ?

drazik:
OK for me. But we need to ensure that this is OK if we do nothing particular when the `url` prop changes.

nono:
Ping @gregorylegarec

gregorylegarec:
@drazik @Crash-- After a little thinking I guess we should name things like that :
* Internally `handleUrlChange` should be named `handleLoadStart`
* For the API I suggest to rename `onUrlChange` to `onMobileAppUrlChange`, which could make things clearer and will allow us to not handle any URL change. For reminder we do need to handle URL changes only on mobile apps for now.

drazik:
OK. But my main concern was about the potential `url` prop value changes. Is it something that can happen? Should we handle it?

gregorylegarec:
@drazik Oh, ok, I finally get your point. I guess the `url` prop is not intended to change BUT it case it happens we should in a first time just ignore it, as I don't see any use case justifying changing popup's url.

ptbrowne:
Instead of ignoring the change of url, should we throw ? I am thinking of this article https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-rendering. Since the language does not prevent you from changing the URL, someone at some point will do it, if it is ignored, it will be a hard to detect bug. If handling a new URL it too complicated, we have several options:

- throw a big fat error
- make it clear that the prop is not meant to be changed, via the `initial` convention

gregorylegarec:
Good point @ptbrowne. I like the `initial` convention. Throwing an error here will just warn about how we want the Popup to be used. As the Popup component open a new window, throwing an error at re-render will not make anything crash. So it's pretty safe here.
By the way, I feel that the `initial` convention is a more React way to solve this case.

gregorylegarec:
Rebase + fixup, tell me if it's ok for you.

react/Popup/Readme.md
```diff
@@ -0,0 +1,21 @@
...
+
+```
+<div>
+  <button onClick={() => setState({ showPopup: true })}>
+    Show popup
+  </button>
+  <p>Popup is {state.showPopup ? "open" : "closed"}</p>
+  { state.showPopup &&
+    <Popup
+      url="http://example.org"
```
drazik:
```suggestion
      initialUrl="http://example.org"
```

react/Popup/index.jsx
```diff
@@ -0,0 +1,199 @@
...
+  componentWillUnmount() {
+    this.killPopup()
+  }
+
+  addListeners(popup) {
+    // Listen here for message FROM popup
+    window.addEventListener('message', this.handleMessage)
+
+    // rest of instructions only on mobile app
+    if (!isMobileApp()) return
```
drazik: Nit: I think if we invert this condition, the code is more self-explanatory and the comment can be removed

```js
if (isMobileApp) {
  popup.addEventListener('loadstart', this.handleLoadStart)
  popup.addEventListener('exit', this.handleClose)
}
```

react/Popup/index.jsx
```diff
@@ -0,0 +1,199 @@
...
+
+  removeListeners() {
+    const { popup } = this.state
+    window.removeEventListener('message', this.handleMessage)
+
+    // rest of instructions only if popup is still opened
+    if (popup.closed) return
+
+    // rest of instructions only on mobile app
+    if (!isMobileApp()) return
```
drazik: Same as above

react/Popup/index.jsx
```diff
@@ -0,0 +1,199 @@
...
+}
+
+/**
+ * Renders a popup and listen to popup events
+ */
+export class Popup extends PureComponent {
+  constructor(props, context) {
+    super(props, context)
+
+    // url is not expected to change
```
drazik: If url is not expected to change, can we avoid putting it in the state, and just use `this.props.initialUrl` everywhere? I don't like using state for things that are not part of `render` method. Since here `render` returns `null`, we should have nothing in the state.

react/Popup/index.jsx
```diff
@@ -0,0 +1,199 @@
...
+      `scrollbars=yes, width=${w}, height=${h}, top=${top}, left=${left}`
+    )
+    // Puts focus on the newWindow
+    if (popup.focus) {
+      popup.focus()
+    }
+
+    this.addListeners(popup)
+    this.startMonitoringClosing(popup)
+    this.setState({ popup })
```
drazik: Can we store the popup in `this.popup` instead of in the state? Extra re-renders in not a big concern since `render` returns `null`, but if we can avoid storing things unrelated to rendering in state, it would be nice.

react/Popup/Readme.md
gregorylegarec: Oh, thanks !

react/Popup/index.jsx
gregorylegarec: I love early returns but I must admit that you're right on this point :)

gregorylegarec:
I took your last feedback into account, @drazik. I think we're good.

PR #1015, opened 2019-05-29
feat: Add file-outline icon

drazik:
This icon will be used in banks. I think it has its place in the cozy-ui icon set.

See https://drazik.github.io/cozy-ui/react/#icon


GoOz:
Already added in the incoming PR #983 but since this PR is not ready to be merged just yet, I guess it's ok if you need it right away.

PR #1018, opened 2019-05-29
feat(Chip): Update outlined themes

drazik:
Some updates to normal and error themes for outlined variant on the Chip component. We want special colors for these cases, so I added specific classes for them.

See https://drazik.github.io/cozy-ui/react


ptbrowne:
Thanks @drazik

PR #1020, opened 2019-05-29
feat(Chip): Add hover and focus states

drazik:
Add hover and focus states for Chips. Only outlined and dashed variant have these states.

See https://drazik.github.io/cozy-ui/react


ptbrowne:
Does it mean that Chip not outlined should not be used for actions but only for information ?

PR #1021, opened 2019-05-31
Add FileInput component

cedricmessiant:
Adds a customizable `FileSelector` that uses a render prop to customize what to display. On change, a callback is called with the selected file(s). Not sure the naming is perfect though (`FileSelector`? `processSelected`? `renderContent`?)

Styleguidist here: https://cedricmessiant.github.io/cozy-ui/react/#fileselector


cedricmessiant:
[(𝐒) Composant select file](https://trello.com/c/02UskpST/2008-%F0%9D%90%92-composant-select-file)

react/FileSelector/Readme.md
```diff
@@ -0,0 +1,48 @@
+This component is used to display a customizable file selector.
+
+### Default
+
+```jsx
+<FileSelector
```
Crash--: I'm not a big fan of `FileSelector` name. Since UI is tied to Cozy, I'm expecting that `FileSelector` let me select a file in Drive :/.

react/FileSelector/Readme.md
```diff
@@ -0,0 +1,48 @@
...
+  renderContent={() => <span>Click me to choose file</span>}
+  processSelected={console.log}
```
Crash--: Again, we're just surcharging the native `onChange`. I think `onChange` or `onSelected` are more accurate.

react/FileSelector/Readme.md
```diff
@@ -0,0 +1,48 @@
+This component is used to display a customizable file selector.
```
Crash--: We should clarify that we wrap an input type=file so we can pass all the related attributes like `accept` `mutliple` ... ?

react/FileSelector/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+  className,
+  disabled,
+  hidden,
+  multiple,
+  processSelected,
+  renderContent,
+  ...inputProps
+}) => (
+  <label role="button" disabled={disabled} className={className}>
+    {renderContent()}
```
Crash--: It's more a `renderLabel` than a `renderContent`, but I'm not very good to find the right name :/

react/FileSelector/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+const FileSelector = ({
+  className,
+  disabled,
+  hidden,
+  multiple,
+  processSelected,
+  renderContent,
+  ...inputProps
+}) => (
+  <label role="button" disabled={disabled} className={className}>
```
Crash--: We should have a cursor:pointer ?

react/FileSelector/Readme.md
cedricmessiant: Maybe `InputFile` is just fine in fact?

react/FileSelector/Readme.md
cedricmessiant: :+1:

react/FileSelector/Readme.md
y-lohse: Or `FileInput`?

react/FileSelector/index.jsx
y-lohse: I would suggest using `children` here instead. Render props with custom names are cool for rendering a fraction of a component, ie. `renderModalHeader` or whatnot. Here there ins't anything else visible aside from the render prop, so I think `children` would work best.

react/FileSelector/index.jsx
GoOz: A label [can't have a `role`](https://www.w3.org/TR/html5/forms.html#the-label-element). It should be on the `span` or whatever child you choose.

react/FileSelector/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+    {renderContent()}
+    <input
+      hidden={hidden}
```
GoOz: Not sure about the `hidden` attribute.
I'm still looking for answers but I'm wondering if the input shouldn't be readable by a screen-reader in this case, and by using the `hidden` attribute it won't.

cedricmessiant: :+1:
It may explain why it was hidden by css on drive (see https://mattermost.cozycloud.cc/cozycloud/pl/97h8xkbeif8t3krcrj4acszify ). So I'm waiting for your feedback on this

GoOz:
So… It will be a little more complicated than expected.

You'll need to add `tabindex="0"` to the span to make it focusable with the keyboard. 
Also you'll need to add an `aria-controls="id_unique_du_input_file"` (and therefore add the unique id to the input) to let the screen reader know that this `span` controls this `input[type=file]`.
Finally you'll need to take care of the keyboard events, like hitting the return key will trigger the `input`.

More details on [this blog post](https://accessabilly.com/a-styled-accessible-file-upload/)

react/FileSelector/index.jsx
cedricmessiant: :+1:

cedricmessiant:
I did the changes. Unfortunately, I can't manage to add unit tests for accessibility (I can't find a way to know if the input[type=file] has been clicked) but otherwise, I think it's mergeable

Crash--:
can you push to your styleguidish ?

react/FileInput/index.jsx
```diff
@@ -0,0 +1,70 @@
...
+  children,
+  className,
+  disabled,
+  hidden,
+  multiple,
+  onChange,
+  ...inputProps
+}) => {
+  // generate a random id for the input
+  const fileinputId = `file-input-${Math.random()
```
GoOz: I don't know if it's best or not but FYI I've used `uniqueId` from lodash to get a unique ID in several components such as [Modal](https://github.com/cozy/cozy-ui/blob/master/react/Modal/index.jsx#L264)

react/FileInput/styles.styl
```diff
@@ -0,0 +1,2 @@
+.parent
```
GoOz: I think it would be best with a more explicit classname such as `c-input-file`, for consistency.

react/FileInput/index.jsx
cedricmessiant: Oh cool, I searched for it :-)

cedricmessiant:
@Crash-- Done https://cedricmessiant.github.io/cozy-ui/react/#fileinput

PR #1026, opened 2019-06-04
feat: Keep radio elements centered

ptbrowne:
If the height of the label is customised, for example to have a larger
clic area, the text was not aligned with the checkbox circle.

View it on http://ptbrowne.github.io/cozy-ui/react/#!/Radio

### DIFF
--- a/react/Radio/index.jsx
+++ b/react/Radio/index.jsx
@@ -4,7 +4,16 @@ import PropTypes from 'prop-types'
 import styles from './styles.styl'
 
 const Radio = props => {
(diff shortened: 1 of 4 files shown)


react/Radio/index.jsx
```diff
@@ -15,6 +24,7 @@ const Radio = props => {
         className
       )}
       aria-disabled={disabled}
+      style={style}
```
GoOz: Is the `style` prop really necessary? Can we just use `...restProps` or something?

ptbrowne: restProps is already applied on the input below.

PR #1028, opened 2019-06-04
fix: Use style to assign bg color to select box

ptbrowne:
Using an utility class makes it impossible to override bg color.

### DIFF
--- a/react/SelectBox/SelectBox.jsx
+++ b/react/SelectBox/SelectBox.jsx
@@ -19,7 +19,9 @@ const heights = {
 const customStyles = props => ({
   control: (base, state) => ({
     ...base,
-    backgroundColor: 'transparent',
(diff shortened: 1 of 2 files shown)


react/SelectBox/SelectBox.jsx
```diff
@@ -19,7 +19,7 @@ const heights = {
 const customStyles = props => ({
   control: (base, state) => ({
     ...base,
-    backgroundColor: 'transparent',
+    backgroundColor: props.disabled ? 'transparent' : 'white',
```
y-lohse: We want a transparent background when the selectbox is disabled ? 🤔

ptbrowne: Yes because the gray background color is managed via the `--disabled` class applied below. I am adding a comment.

PR #1029, opened 2019-06-04
Chore: transpiled stylesheet does not have hashes

ptbrowne:
This will enable the bar and the app to rely on the same
stylesheet.


ptbrowne:
Closing as we settled on another solution (bar using cozy-ui not transpiled so that it can take only what it needs) pretty easy to reopen if needed.

PR #1032, opened 2019-06-10
Add Card component

gregorylegarec:
🚫 Blocker for https://github.com/cozy/cozy-libs/pull/490

[Styleguidist](https://gregorylegarec.github.io/cozy-ui/react/#!/Card)

Card is an UI that can be used to isolate some content from the reste of the page.

Mainly used in coming Harvest UI (in konnector modal for example). Markup is mainly used to apply borders and a paddings styles.

Capture d’écran 2019-06-10 à 17 17 32
Capture d’écran 2019-06-10 à 17 17 38
Capture d’écran 2019-06-10 à 17 17 46


y-lohse:
Couldn't that be a variation of the [`Well`](https://docs.cozy.io/cozy-ui/react/#well)?

gregorylegarec:
@y-lohse Maybe, but actually a Card should not be an inset component.

ptbrowne:
> Couldn't that be a variation of the Well?

You mean that instead of writing `<Card />` you would write `<Well variation='card' />` ? If that is the case, I disagree. The goal of the Well component is to be inset. This is not the goal of the Card, one should not be a variation of the other.

y-lohse:
> Card is an UI that can be used to isolate some content from the reste of the page.

I think this fits both `Card` and `Well`. The `Well` also happens to be inset (and have a background).
I would suggest we:

- keep the `Card`
- add an `inset` variation-
- redirect `Well` to use `<Card inset />` and add a depreciation notice
- Add a `className` that can be applied to the wrapper for background colors etc

ptbrowne:
:+1: OK to replace a specific component with a more generic one ! Cool !

I think Well can be deprecated pretty quickly since I find occurences of Well only in Drive `drive/src/drive/web/modules/move/Header.jsx`.


https://getbootstrap.com/docs/4.3/migration/#panels-thumbnails-and-wells It seems to be the direction of bootstrap also.

Crash--:
Agreed with the last comment ;). 

Also, could be nice to have a snapshot for this component.

react/Card/index.jsx
```diff
@@ -0,0 +1,12 @@
+import React, { PureComponent } from 'react'
+
+import styles from './styles.styl'
+
+export class Card extends PureComponent {
+  render() {
+    const { children } = this.props
+    return <div className={styles['c-card']}>{children}</div>
```
drazik: I think you should spread props:

```jsx
import cx from 'classnames'

render() {
  const { className, ...rest } = this.props
  return <div className={cx(styles['c-card'], className)} {...rest} />
}
```

So if we need to give DOM props to the component, it's possible.

gregorylegarec: Well spotted, thanks.

gregorylegarec:
As I need this fist version of `Card`, I suggest to deprecate `Well` and make changes and tests in a second time.

PR #1041, opened 2019-06-13
fix(Checkbox): Check icon position when multiline label

GoOz:
When a checkbox have a label long enough to be on multiple lines, the `check` icon is misplaced (it's actually top aligned and white on white so you can't see it)

Before:
Screenshot 2019-06-13 11 30 05

After:
Screenshot 2019-06-13 11 30 15

### DIFF
--- a/react/Checkbox/styles.styl
+++ b/react/Checkbox/styles.styl
@@ -11,7 +11,7 @@
 .c-input-checkbox-icon
     position absolute
     left 0
-    top rem(3)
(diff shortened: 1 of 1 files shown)


GoOz:
[[UI] Multi-line checkbox is misplaced when checked.](https://trello.com/c/0KfMIcrJ/1541-ui-multi-line-checkbox-is-misplaced-when-checked)

PR #1045, opened 2019-06-14
fix(I18n): Improve hocs proptypes handling + expose proptypes

CPatchane:
Get exposed prop-types is useful when you want to check or get an exhaustive list of provided props and the related types for developing or for testing

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -74,14 +74,20 @@ I18n.childContextTypes = i18nContextTypes
 
 // higher order decorator for components that need `t` and/or `f`
 export const translate = () => WrappedComponent => {
-  const Wrapper = (props, context) => (
-    <WrappedComponent
-      {...props}
-      t={context.t}
-      f={context.f}
-      lang={context.lang}
-    />
-  )
+  const Wrapper = (props, context) => {
+    return (
+      <WrappedComponent
+        {...props}
+        t={context.t}
+        f={context.f}
+        lang={context.lang}
+      />
+    )
+  }
+  Wrapper.propTypes = {
+    ...(WrappedComponent.propTypes || {}),
+    ...i18nContextTypes
+  }
   Wrapper.contextTypes = i18nContextTypes
   return Wrapper
 }
--- a/react/I18n/withLocales.jsx
+++ b/react/I18n/withLocales.jsx
@@ -19,14 +19,20 @@ const withLocales = localesOrRequire => Component => {
 
   class Wrapped extends React.Component {
     render() {
-      // Do not pass t downwards
-      const { lang, ...rest } = omit(this.props, 't')
+      const { lang } = this.props
+      // Do not pass translate props downwards
+      // since the component is already augmented with translate()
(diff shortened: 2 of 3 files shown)


CPatchane:
[(𝐋) 🔵 🍊 [HOME] Migration de  la Home sur React](https://trello.com/c/CWnlkCYC/1442-%F0%9D%90%8B-%F0%9F%94%B5-%F0%9F%8D%8A-home-migration-de-la-home-sur-react)

drazik:
What is this `childPropTypes` thing? I can't find it in the react docs. Is it "standard" ?

ptbrowne:
> What is this childPropTypes thing? I can't find it in the react docs. Is it "standard" ?

Reading this, I have the impression that childPropTypes is the same as contextTypes and I am not sure we should introduce a new name for them. Is there a reason or could we stick we contextTypes ?

y-lohse:
And if it's the same as `contextTypes ` then I'm not sure it makes to have it on the `withLocales` HOC? Since it receives the `i18n` context props as… well, as regular props, and not through the context.

I'm all for exporting the proptypes but I'd need more explanation about why we are using `childPropTypes ` here before giving a 👍

CPatchane:
No this is not standard but this is different from `contextTypes`:
* `contextTypes` means context properties provided to the HOC itself by the context
* `propTypes` means props provided to the HOC itself by a parent
* by `childPropTypes` I was meaning props passed from the HOC to its child, I did see this naming somewhere but I can't remember where but this is not a standard React naming convention I think

y-lohse:
If it's not standard, maybe we could do something like:

```
import { I18nPropTypes } from 'cozy-ui/react/I18n'
...
MyTranslatedComponent.propTypes = {
  ...I18nPropTypes
}
```

Whenever we need the proptypes?

CPatchane:
The case is here https://github.com/cozy/cozy-libs/pull/496/files#diff-54041063e1ea319e56e155199be0736bR45
We have a component with a Button, we want to omit some props so we use this childContextTypes
Maybe your solution will works @y-lohse, I'll give it a try

CPatchane:
It works correctly by providing those propTypes to the wrapped component 👍, I updated the PR @y-lohse

react/I18n/index.jsx
```diff
@@ -74,14 +74,20 @@ I18n.childContextTypes = i18nContextTypes
...
+      <WrappedComponent
+        {...props}
+        t={context.t}
+        f={context.f}
+        lang={context.lang}
+      />
+    )
+  }
+  Wrapper.propTypes = {
+    ...(WrappedComponent.propTypes || {}),
```
drazik: That sounds weird but you can actually destructure `undefined`:

```js
> { ...undefined }
{}
```

So you can just do

```js
...WrappedComponent.propTypes
```

even if `WrappedComponent.propTypes` is undefined.

CPatchane: Fixed by b39adba1c709803f39830c506dd0552d71b334f3

PR #1046, opened 2019-06-14
fix(Field): Correctly handling select value proptype

CPatchane:
The Field value props should be a string for all types except for the `select` where the value should be an object.
Add also a test for that.

### DIFF
--- a/react/Field/index.spec.js
+++ b/react/Field/index.spec.js
@@ -1,0 +1,35 @@
+import React from 'react'
+import Field from './'
+import { shallow } from 'enzyme'
+
+describe('Field component', () => {
+  beforeEach(() => {
+    // by default, proptypes checking just log errors
+    jest.spyOn(console, 'error').mockImplementation(message => {
+      throw new Error(message)
+    })
+  })
+
+  afterEach(() => {
+    jest.resetAllMocks()
+    jest.restoreAllMocks()
+  })
+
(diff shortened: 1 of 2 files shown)


react/Field/index.spec.js
```diff
@@ -0,0 +1,37 @@
...
+describe('Field component', () => {
+  afterEach(() => {
+    jest.resetAllMocks()
+  })
+
+  it('should expect object type value for type=select', () => {
+    // mock for proptypes
+    console.error = jest.fn(message => {
+      throw new Error(message)
+    })
```
ptbrowne: Prefer to replace console.error in a beforeEach (since it is repeated twice) with a spyOn (to be restored in a afterEach), since we are modifying a global object, it is better to be hygienic even if I think Jest resets the jsdom environment between each test.

CPatchane:
[(𝐋) 🔵 🍊 [HOME] Migration de  la Home sur React](https://trello.com/c/CWnlkCYC/1442-%F0%9D%90%8B-%F0%9F%94%B5-%F0%9F%8D%8A-home-migration-de-la-home-sur-react)

react/Field/index.spec.js
CPatchane: This test file could have more tests about the component itself since we just need a simple mock for those two test case and that could break other test case I prefer to keep them here

PR #1052, opened 2019-06-18
Feat/ghost button

GoOz:
For a new type of button we need:
- [x] A new theme for button named "ghost"
- [x] A new prop `align` to change the label alignment
- [x] A new prop `extraRight` to add some text, number, component to the right of the button

[Preview](https://gooz.github.io/cozy-ui/react/#button)

### DIFF
--- a/react/Button/Readme.md
+++ b/react/Button/Readme.md
@@ -57,6 +57,23 @@ const { Button } = require('./index');
 </div>
 ```
 
+#### Text alignment
+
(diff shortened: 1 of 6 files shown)


GoOz:
[Modifier le composant Bouton pour autoriser un composant à sa droite](https://trello.com/c/wq90MfLL/1550-modifier-le-composant-bouton-pour-autoriser-un-composant-%C3%A0-sa-droite)

PR #1053, opened 2019-06-18
feat(Chip): Add new theme & new size

GoOz:
Adds a new theme & a new size for the Cip component.

[Preview](https://gooz.github.io/cozy-ui/react/#chip)


GoOz:
[Modifier le composant Bouton pour autoriser un composant à sa droite](https://trello.com/c/wq90MfLL/1550-modifier-le-composant-bouton-pour-autoriser-un-composant-%C3%A0-sa-droite)

stylus/components/chip.styl
```diff
@@ -21,6 +24,12 @@ $round-chip
     text-align center
     justify-content center
 
+$tiny-size-chip
+    height $chip-height-tiny
+    padding 0 (($chip-height-tiny - $chip-font-size-tiny) / 2)
+    border-radius ($chip-height-tiny / 2)
```
y-lohse: Wouldn't 50% do the trick as well?

react/Button/Readme.md
```diff
@@ -81,7 +81,7 @@ const { Button } = require('./index');
 const Chip = require('../Chip').default;
 
 <div>
-  <p><Button extension="full" size="large" label='Label' extraRight={<Chip size="small" theme="primary" className="u-m-0">1/2</Chip>} /></p>
+  <p><Button theme="ghost" extension="full" size="large" label='Label' extraRight={<Chip size="tiny" theme="full" className="u-m-0">1/2</Chip>} /></p>
```
y-lohse: I don't think this should be part of the diff?

drazik:
When I implemented the primary theme for the normal variant, I did it almost blindly. This is used nowhere. I think this can be the primary theme of the normal variant, so we don't introduce another combination.

stylus/components/chip.styl
ptbrowne: Could the padding be done via em ? So that we have different size for free just by changing the font size ?

ptbrowne: https://jsbin.com/pipomijiki/1/edit?html,css,output

GoOz: @y-lohse I think we already got this conversation and no, this 50-100% trick works only with circle objects.

GoOz: @ptbrowne There is. `em` has the strongest cascade functionnality of all CSS. Your way works if there is **not any** parent that change its font-size, otherwise this won't be consistent.
Not that it can't be done but you have to be extra careful, specially when this is a component that can be called anywhere, and I don't think the economy of a few lines is worth the trouble.

GoOz:
@drazik ah cool, I wasn't really fond of `full` anyway ;)

react/Button/Readme.md
GoOz: Mmmmh Ok I'll make another PR out of it.

stylus/components/chip.styl
ptbrowne: > There is. em has the strongest cascade functionnality of all CSS. Your way works if there is not any parent that change its font-size, otherwise this won't be consistent.

If we put `.Chip--normal { font-size: 1rem }`, `.Chip--xs { font-size: 0.75rem }`, it would be OK right ?

GoOz: Should be yeah, but that's not just it.
The actual different sizes are not proportionnal. The ratio `height/font-size` is not the same for each size.
- tiny = 2
- small = 2.285714286
- normal = 2.5

drazik: Maybe we should leave a TODO comment here (and/or an issue) and propose to use proportional sizes to the design team.

GoOz: I'm not comfortable telling the designers what to do just to get a few less bytes of CSS, honestly.

ptbrowne: OK, thanks to have checked the ratios.

react/Button/Readme.md
y-lohse: Well if it's not a mistake and actually needs to land on master, I don't mind leaving it as part of this PR :)

stylus/components/chip.styl
drazik: That's why I said "propose". That's just a suggestion: if they are OK then alright, otherwise :man_shrugging:

react/Button/Readme.md
GoOz: It wasn't a mistake but you're kinda right anyway, it was intended to update the `extraRight` demo with the new Chip theme & size but it's not really related to the core of this PR. 

I already removed it so I'll add it later in a clean & independant PR once this once is merged.

GoOz:
@drazik Ok I misunderstood 😅 
I juste updated the code and the preview. Should be ok now.

PR #1055, opened 2019-06-19
feat: Add InlineCard component

GoOz:
Taken from Banks, here is `InlineCard` (previously known as `TextCard`)

Preview to come…


GoOz:
[Faire le composant Inline Card dans UI (à rapatrier de TextCard de Banks)](https://trello.com/c/Bzo1CBNh/1549-faire-le-composant-inline-card-dans-ui-%C3%A0-rapatrier-de-textcard-de-banks)

ptbrowne:
Awesome, thanks :tada:

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -1271,6 +1271,8 @@ exports[`Infos should render examples: Infos 1`] = `
 </div>"
 `;
 
+exports[`InlineCard should render examples: InlineCard 1`] = `"<div></div>"`;
```
Crash--: I don't think this is want we want?

GoOz: Indeed… but oddly I did just like I did with others… I'll investigate further.

GoOz: Should be better now. :)

PR #1058, opened 2019-06-21
fix: Move dimension props from wrapper to modal

### LINKED ISSUE #899: [modal] Modal height set modal wrapper height
When passing `height` prop to a `<Modal />` component, the height is applied to wrapper and the final height is not consistent.

Capture d’écran 2019-04-12 à 12 43 32

GoOz:
Move the `style` prop (which contains `width` & `height` props) from the Modal wrapper to the Modal itself.

Originally, those props were added to force a modal height for Intent and it worked for them but it couldn't be used for regular modals.
Now it can & still works for Intent modals.

Fixes #899

### DIFF
--- a/react/Modal/index.jsx
+++ b/react/Modal/index.jsx
@@ -327,7 +327,6 @@ class Modal extends Component {
               },
               wrapperClassName
             )}
-            style={style}
             onClick={closable ? this.handleOutsideClick : undefined}
           >
             <div
@@ -342,6 +341,7 @@ class Modal extends Component {
                 },
                 className
               )}
+              style={style}
               role="dialog"
               aria-modal="true"
               aria-labelledby={title ? titleID : null}


GoOz:
[[UI] ETD je peux spécifier la hauteur d'une modale](https://trello.com/c/plOlxIpK/1498-ui-etd-je-peux-sp%C3%A9cifier-la-hauteur-dune-modale)

y-lohse:
That's a breaking change though, right? Have you checked if our apps use this style prop?

GoOz:
@y-lohse Shouldn't break no. I tested it on Sandbox, it still works with Intents either way.

y-lohse:
Oh right, `style` is not directly a prop, we're only changing what `width` and `height` affect. Should be good then.

PR #1062, opened 2019-06-24
Text: Add Uppercase component ✨

gregorylegarec:
Also add u-uppercase utility class.

As seen with @GoOz, we concluded that it was a good thing to do to make a new component as we made for `<Bold />`


gregorylegarec:
I forgot, my bad 🙏 https://gregorylegarec.github.io/cozy-ui/react/#!/BaseText

PR #1068, opened 2019-06-25
feat: Invert Preact/React flag so that React becomes the default

ptbrowne:
BREAKING CHANGE

I  changed the environment variable from USE_REACT=true to USE_PREACT so that apps using React have the least amount of job to do.


ptbrowne:
Ping yannick, I changed of approach to be less breaking (apps upgrading cozy-ui, and still on Preact would only have to add an environment variable).

PR #1079, opened 2019-07-01
Handle fallback icon in AppIcon

CPatchane:
Needed to provided an icon to fallback different from the cube

### DIFF
--- a/react/AppIcon/index.jsx
+++ b/react/AppIcon/index.jsx
@@ -5,6 +5,7 @@ import styles from './styles.styl'
 
 import Icon from '../Icon'
 import palette from '../palette'
+import { iconPropType } from '../Icon'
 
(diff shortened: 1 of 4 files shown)


react/AppIcon/index.jsx
```diff
@@ -107,6 +114,8 @@ AppIcon.propTypes = {
   alt: PropTypes.string,
   /** Required if fetchIcon is not provided */
   app: PropTypes.oneOfType([AppDoctype, PropTypes.string]),
+  /** Icon to fallback on error (optional), default cube icon */
+  fallbackIcon: PropTypes.string,
```
Crash--: is it a string or an IconProptypes ? You have access to iconPropType from `Icon`

CPatchane: I didn't see that, indeed, I add it 👍

PR #1083, opened 2019-07-02
fix(nav): Fix Icon alignment

GoOz:
Icons in nav were not correctly aligned.

**Avant:**
Screenshot 2019-07-02 15 26 16

**Après:**
Screenshot 2019-07-02 15 26 24

PR #1084, opened 2019-07-03
fix(Nav): Icon alignement on mobile

GoOz:
#1083 introduced a regression on mobile navigation. Icons weren't correctly aligned anymore.
So here's the fix. 

Also the styleguide example wasn't using the right classnames, so It's fixed too.


GoOz:
[(𝐒) Amelioration mineure de la sidebar](https://trello.com/c/387TgpV1/1379-s-amelioration-mineure-de-la-sidebar)

Issue #1074, opened 2019-06-27
Button: Need to manage line-height


gregorylegarec:
In cozy-harvest-lib, we need to set explicitely the line-height of our "Synchronize" button. See https://github.com/cozy/cozy-libs/pull/537#discussion_r298070682. This specific line-height allow us to make a correct horizontal alignement between button and informations on its left.

We should offer a way to set this line-height without using the `style` attribute, either with an utility class or a Button dedicated prop

GoOz:
Not sure the dedicated prop is relevant. I think the `Button` component has quite enough props as it is. Also this could be something that other component needs so I think a utility class is the way to go.

y-lohse:
I also think a utility class in the app is the way to go, this seems like a very specific use case.

ptbrowne:
Can you post a screenshot of where it's needed ?

ptbrowne:
ping ?

GoOz:
![](https://github.trello.services/images/mini-trello-icon.png) [Ajouter des classes utilitaires pour le line-height](https://trello.com/c/CQHirY57/1604-ajouter-des-classes-utilitaires-pour-le-line-height)

gregorylegarec:
@ptbrowne Sorry I missed your ping.

This is what it looks like with a 1.4 line-height on the button on the right. Without it, the button is not aligned with the text on the left.
![Capture d’écran 2019-07-04 à 10 56 00](https://user-images.githubusercontent.com/776764/60653694-d46a1000-9e4a-11e9-8e85-1b989d6016b8.png)

ptbrowne:
J'ai essayé de reproduire le soucis dans un codesandbox pour mieux comprendre, je suis pas sur qu'on ait vraiment besoin de line-height ici, est ce qu'utiliser Media ne suffirait pas ?

https://codesandbox.io/s/fast-tdd-e147o

gregorylegarec:
It's not exactly the markup we have in the [`LaunchTriggerCard` component](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-harvest-lib/src/components/cards/LaunchTriggerCard.jsx#L28).

Moreover this part is responsive; on small screen the button is below the two texts.

PR #1088, opened 2019-07-03
feat: Add line-height utility classes

GoOz:
Adds utilty classes for line-height :

- `u-lh-tiny` - Sets a tiny line-height (1)
- `u-lh-xsmall` - Sets a xsmall line-height (1.1)
- `u-lh-small` - Sets a small line-height (1.2)
- `u-lh-medium` - Sets a medium line-height (1.3)
- `u-lh-large` - Sets a large line-height (1.4)
- `u-lh-xlarge` - Sets a large line-height (1.5)

Fixes: #1074

### DIFF
--- a/stylus/tools/mixins.styl
+++ b/stylus/tools/mixins.styl
@@ -87,13 +87,21 @@ reset()
     padding 0
     border  0
 
-global(selector, placeholder)
-    if cssmodules == true
-        :global({selector})
-            @extend {placeholder} // @stylint ignore
-    else
-        {selector}
-            @extend {placeholder} // @stylint ignore
+global(selector, placeholder, type = extend)
+    if type == extend
+        if cssmodules == true
+            :global({selector})
+                @extend {placeholder} // @stylint ignore
+        else
+            {selector}
+                @extend {placeholder} // @stylint ignore
+    else if type == mixin
+        if cssmodules == true
+            :global({selector})
+                {placeholder}() // @stylint ignore
+        else
+            {selector}
+                {placeholder}() // @stylint ignore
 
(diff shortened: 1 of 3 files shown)


GoOz:
[Ajouter des classes utilitaires pour le line-height](https://trello.com/c/CQHirY57/1604-ajouter-des-classes-utilitaires-pour-le-line-height)

y-lohse:
Mmh meh.

Are there many uses cases on our designs were we have custom line heights? And do we need to make them responsive?

Sorry but since I realised our utility class file is already large-ish, I think we need to be a bit more careful about what utility classes we add.

GoOz:
Alright.
No we don't necessarily need those classes to be responsive.
And no we don't have a lot of custom line-height but we did have a usecase where we needed it.

GoOz:
@y-lohse I removed the responsive versions

stylus/tools/mixins.styl
```diff
@@ -87,13 +87,21 @@ reset()
...
     border  0
 
-global(selector, placeholder)
-    if cssmodules == true
-        :global({selector})
-            @extend {placeholder} // @stylint ignore
-    else
-        {selector}
-            @extend {placeholder} // @stylint ignore
+global(selector, placeholder, type = extend)
```
y-lohse: why do we need support for mixins all o a sudden?

GoOz: `cssModulesUtils` & `NativeUtils` are the functions that's create utility classes with their responsive versions, and they allow mixins (actually we have to use mixins because `extend` doesn't work with media queries).

But the legacy function here that create global classes (so utility classes) only allows `extends`.

Issue #710, opened 2018-11-22
Add "dual field" component


GoOz:
As it is used for [authentification](https://github.com/cozy/cozy-ui/pull/708).
<img width="504" alt="screenshot 2018-11-22 11 08 35" src="https://user-images.githubusercontent.com/1280069/48896113-1d82cc00-ee47-11e8-81a8-dfce01daeee5.png">
May need a little bit of Design specifications.

y-lohse:
Before implementing this, we need to list the various use cases (editable domain, protocol or not, etc).

ptbrowne:
What are the responsibilities of this component ?

y-lohse:
Yes that's more or less the question. We have a chat scheduled with @joel-costa to clear things up :)

PR #1093, opened 2019-07-05
feat: Add InputGroup component

GoOz:
Add new InputGroup component.

Basically it's a wrapper component with a prepend or append component that can be pretty much anything but it has been specially made for a `Text` component, an other `Input` or a `SelectBox`.

[Preview](https://gooz.github.io/cozy-ui/react/#!/InputGroup)

Fixes: #710


GoOz:
[(𝐌) [UI] Ajouter le composant d'url](https://trello.com/c/eo92heRU/612-%F0%9D%90%8C-ui-ajouter-le-composant-durl)

ptbrowne:
Field is a combination of Label and Input.
DualField seems more close to a Input.

I was thinking of renaming DualField to DualInput and then thought that it would be even simpler to have a side property in Input. If the implementation between the current DualField and Input are too different, Input could completely delegate to the current DualField implementation if it has a side property. What do you think ?

react/InputGroup/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+      focused: false
+    }
+  }
+
+  handleFocus() {
+    this.setState(() => ({ focused: true }))
+  }
+
+  handleBlur() {
+    this.setState(() => ({ focused: false }))
```
ptbrowne:
```suggestion
    this.setState({ focused: false })
```

It is not necessary to use the functional variant of setState when state is not used to compute the new value of state.

react/InputGroup/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+    super(props)
+    this.handleFocus = this.handleFocus.bind(this)
+    this.handleBlur = this.handleBlur.bind(this)
+    this.state = {
+      focused: false
+    }
+  }
+
+  handleFocus() {
+    this.setState(() => ({ focused: true }))
```
ptbrowne:
```suggestion
    this.setState({ focused: true })
```

stylus/components/forms.styl
```diff
@@ -467,6 +467,49 @@ $field-inline
         flex-direction column
         margin-left 0
 
+$inputgroup
+    display inline-flex
+    flex-direction row
+    align-items stretch
+    width 100%
+    max-width rem(512)
```
ptbrowne: I am not sure we want to set a max-width here. It seems like an arbitrary limit.

react/InputGroup/Readme.md
```diff
@@ -0,0 +1,89 @@
+### InputGroup with a appended text
```
drazik:
```suggestion
### InputGroup with an appended text
```

react/InputGroup/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+<form>
+  <div>
+    <InputGroup prepend={<Bold className="u-pl-1">text</Bold>}>
+      <Input placeholder="Placeholder" />
+    </InputGroup>
+  </div>
+</form>
+```
+
+### InputGroup with a appended input
```
drazik:
```suggestion
### InputGroup with an appended input
```

react/InputGroup/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+<form>
+  <div>
+    <InputGroup append={<Input placeholder="@domain.tld" className="u-w-4"/>}>
+      <Input placeholder="Email" />
+    </InputGroup>
+  </div>
+</form>
+```
+
+### InputGroup with a appended select
```
drazik:
```suggestion
### InputGroup with an appended select
```

react/InputGroup/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+<form>
+  <div>
+    <InputGroup append={<SelectBox inset options={options} className="u-w-4" />}>
+      <Input placeholder="URL" />
+    </InputGroup>
+  </div>
+</form>
+```
+
+### Full width InputGroup with a appended text
```
drazik:
```suggestion
### Full width InputGroup with an appended text
```

react/InputGroup/Readme.md
```diff
@@ -0,0 +1,89 @@
...
+<form>
+  <div>
+    <InputGroup fullwidth append={<Bold className="u-pr-1">text</Bold>}>
+      <Input placeholder="Placeholder" />
+    </InputGroup>
+  </div>
+</form>
+```
+
+### Errored InputGroup with a appended text
```
drazik:
```suggestion
### Errored InputGroup with an appended text
```

GoOz:
@drazik Very different, like… a lot. It looks the same but it's definitely not done the same way.

drazik:
OK I would be in favor of branching directly in the `Input` component do avoid adding a new component. But I just saw the discussion on mattermost about this topic. Since the decision has already been discussed, that's ok

stylus/components/forms.styl
GoOz: This is the same `max-width` the `input` has.

ptbrowne: A shared variable between input and inputgroup would be more explicit IMHO :)

PR #1094, opened 2019-07-08
Field: Remove Field label top-margin

gregorylegarec:
BREAKING CHANGE

[Styleguidist](https://gregorylegarec.github.io/cozy-ui/react/#!/Field)

Fixes https://github.com/cozy/cozy-libs/issues/629

In https://github.com/cozy/cozy-libs/pull/625 we hacked the whole form by setting a negative top-margin, as cards should define a bottom margin.

The objective of this PR is to avoid using a default top-margin on Field's labels element.

It implies to decrease the top position of side elements as well.

List of components potentially impacted by this change:
* https://github.com/cozy/cozy-banks/blob/4306a26d4b297ae56ecbfcef914256adfa156db4/src/ducks/transfers/steps/Amount.jsx 
* https://github.com/cozy/cozy-libs/blob/b5f3840001627846395dcffc410ae8fb0074d27a/packages/cozy-harvest-lib/src/components/AccountForm/AccountField.jsx
* https://github.com/cozy/cozy-libs/blob/a389feb98a9e75c46a9d634095260d6028b2cdaf/packages/cozy-harvest-lib/src/components/TwoFAModal.jsx

### DIFF
--- a/react/Field/styles.styl
+++ b/react/Field/styles.styl
@@ -9,7 +9,7 @@
 
 .o-side
     position absolute
-    top 1.5rem
+    top 0.5rem
     left 0
     width 100%
     max-width rem(512)
--- a/stylus/components/forms.styl
+++ b/stylus/components/forms.styl
@@ -238,7 +238,6 @@ $label
 
 $label--block
     display block
-    margin-top rem(16)
     padding rem(8) 0
 
 $input--disabled


gregorylegarec:
@y-lohse This fits our need for Harvest and will allow us to remove an ugly hack.

PR #1095, opened 2019-07-08
fix(label): Remove required for htmlFor

Crash--:
It seems that `for` is not mandatory in the HTML Spec. In some case, we use the label without any input.

### DIFF
--- a/react/Label/index.jsx
+++ b/react/Label/index.jsx
@@ -24,7 +24,7 @@ const Label = props => {
 
 Label.propTypes = {
   children: PropTypes.node.isRequired,
-  htmlFor: PropTypes.string.isRequired,
+  htmlFor: PropTypes.string,
   className: PropTypes.string,
   error: PropTypes.bool,
   block: PropTypes.bool


GoOz:
No it's not mandatory in the HTML Spec as is, but it kinda is if used with an `input`, unless the `input` is inside the `label`.

I'm curious, what are the cases of a `label` without `input` or any other form element you have in mind?

ptbrowne:
> It seems that for is not mandatory in the HTML Spec

> No it's not mandatory in the HTML Spec as is,

Please include links to the specs :D https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label

> Alternatively, you can nest the <input> directly inside the <label>, in which case the for and id attributes are not needed because the association is implicit:

It is explicitly optinal when using input inside of label.

GoOz:
> It is explicitly optinal when using input inside of label.
And we don't do that most of the time, so when used with an `input` there should be a couple id/for.

Crash--:
We used `<Label>` for that element (Piece d'identité) : 
Capture d’écran 2019-07-09 à 08 10 05

Because at the beginning the following composant was an input but now this is just a button. I don't know if `Label` makes sense or if we had to change to something like `Caption` but it does what we want currently.

ptbrowne:
:shipit:

PR #1102, opened 2019-07-11
Fixes in wizard styles

drazik:
We had two problems with the wizard styles in cozy-authentication:

* The badge on the logo was misaligned: 

|Before|After|
|------|-----|
|image|image|

We want to logo on this page to look like the mobile app's logos where the badge is aligned on the right

* The fixed back arrow was misaligned:

|Before|After|
|------|-----|
|image|image|

This arrow was just weirdly placed.

### DIFF
--- a/stylus/components/wizard.styl
+++ b/stylus/components/wizard.styl
@@ -176,7 +176,7 @@ $wizard-logo-badge
     justify-content center
     position absolute
     bottom rem(6)
-    right rem(16)
+    right rem(-6)
     width rem(32)
     height rem(32)
     border .125rem solid white
--- a/stylus/components/wizard.styl
+++ b/stylus/components/wizard.styl
@@ -166,17 +166,17 @@ $wizard-logo
     position relative
     margin 0 auto
     width rem(120)
-    height rem(120)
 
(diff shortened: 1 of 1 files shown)


GoOz:
Did you only check the rendering on mobile ? or a desktop view as well ?

stylus/components/wizard.styl
```diff
@@ -241,8 +241,8 @@ $wizard-desc--footer
 
 $wizard-header-fixed
     position fixed
-    top rem(14)
-    left 0
+    top rem(8)
+    left rem(16)
```
ptbrowne: I do not understand why there are "magic" values here. It seems we are moving the header where in fact we want to increase the padding of the arrow.

drazik:
> Did you only check the rendering on mobile ? or a desktop view as well ?

I didn't check on desktop, but it's actually totally broken in this context:

image

The arrow is obviously alone in the upper left corner of the window. I think this arrow should not have `position: fixed` and be aligned vertically with the using `align-content: center` on their parent element.

drazik:
I reworked it so the arrow can be placed alongside the title.

PR #1106, opened 2019-07-17
feat: Page layout

ptbrowne:
The page layout is used when you want to have a button at the bottom
of the screen, that is shown above keyboard if the content of the page is small enough. Otherwise, it stays at the bottom.


react/Page/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+        height: keyboardState === 'showing' ? keyboardHeight : 0
+      }}
+    >
+      {' '}
+    </div>
+  )
+}
+
+export const PageLayout = ({ children, extraHeight = 0 }) => {
+  const { keyboardState, keyboardHeight } = useKeyboardInfo()
```
Crash--: Can we add : 
if(!window.Keyboard){ console.warn('This component works with ...')} ? 

Also does it work on mobile web version also? Or do we need to do a check?

Crash--:
Also, can you set the peerDep & devDep to 16.8 since you are using hooks ? https://github.com/cozy/cozy-ui/blob/master/package.json#L137

react/Page/index.jsx
ptbrowne: - Added the warning in the "mount/update" phase of the effect.
 - On mobile web, the button will be behind the keyboard

PR #1113, opened 2019-07-23
feat: Add InputGroup.Unit

ptbrowne:
Make a component for something that we use in several apps.


ptbrowne:
> LGTM but snapshots should fail

Good catch, the InputGroup examples were not snapshotted. I added them. Snapshot failure will be for next time :)

Crash--:
Thanks

Issue #1096, opened 2019-07-10
Stack layout


ptbrowne:
Following the https://every-layout.dev/layouts/stack/ article, I was wondering if it would make sense to have some components/utility classes for stacks.

```
.stack > * + * {
  margin-top: 1rem
}

.stack--dense > * + * {
  margin-top: 0.5rem
}

.stack--sparse > * + * {
  margin-top: 1.5rem
}
```

and their responsive companions.

ptbrowne:
Ping @joel-costa

GoOz:
It would, yes. This is meant to be some kind of utility classes, right?
Althought there should be a complete documentation to highlight the strict DOM that is needed for this to work as expected.

drazik:
A component like the one we have in banks (https://github.com/cozy/cozy-banks/blob/master/src/components/Stack/index.jsx) would be great yes!

y-lohse:
Do you see a use case that the {Material-UI List](https://docs.cozy.io/cozy-ui/react/#list-1) component doesn't solve? It also has this concept of dense for example.

GoOz:
@y-lohse yeah but the List component comes with some styles (visually I mean). I think what @ptbrowne needs is a "transparent" component that just acts as a layout, in order to apply a consistent spacing on all its children.

PR #1115, opened 2019-07-24
feat: Stack component

ptbrowne:
Add the Stack component that is used to vertically space the elements it
contains. This should help to have a coherent vertical spacing.

Fixes https://github.com/cozy/cozy-ui/issues/1096

Preview on https://ptbrowne.github.io/cozy-ui/react/#!/Stack


react/Stack/index.jsx
```diff
@@ -0,0 +1,20 @@
...
+const sparseStyle = styles['Stack--sparse']
+const denseStyle = styles['Stack--dense']
+
+const Stack = ({ sparse, dense, ...props }) => {
+  return (
+    <div
+      {...props}
+      className={cx(props.className, styles.Stack, {
+        [sparseStyle]: sparse,
+        [denseStyle]: dense
```
drazik: We can pass `sparse={true}` and `dense={true}`, which is weird. What do you think about `variant` or `spacing` with a single string value (`"sparse"` / `"dense"`)?

PR #1116, opened 2019-07-24
Figure component

ptbrowne:
Bring the Figure/FigureBlock components from Banks.

image

### DIFF
--- a/react/Figure/FigureBlock.jsx
+++ b/react/Figure/FigureBlock.jsx
@@ -1,0 +1,48 @@
+import React from 'react'
+import Types from 'prop-types'
+import classNames from 'classnames'
+import Figure from 'components/Figure/Figure'
+import styles from 'components/Figure/FigureBlock.styl'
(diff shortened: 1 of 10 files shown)


react/Figure/FigureBlock.jsx
```diff
@@ -0,0 +1,51 @@
...
+  symbol,
+  coloredPositive,
+  coloredNegative,
+  signed,
+  decimalNumbers = 0,
+  figureClassName,
+  withCurrencySpacing
+}) => (
+  <div className={classNames(styles['FigureBlock'], className)}>
+    <div className={labelStyle}>{label}</div>
```
drazik: Can't we use the `Label` component directly?

ptbrowne: No it is a `label` element.

drazik: Of course :)

PR #1120, opened 2019-07-26
fix: Use browserslist-config-cozy instead of custom autoprefixer config

drazik:
In postcss config, we gave the browsers list to autoprefixer plugin
directly. But the `browsers` property is deprecated and will be removed
in next major version (see
https://github.com/postcss/autoprefixer/releases/tag/9.6.0).

We also use autoprefixer through autoprefixer-stylus, but it uses a very
old version of autoprefixer (7.1.0) that doesn't understand the `dead`
query that our browserslist config uses. A PR fixing this have been
merged and will hopefuly be released soon, see
https://github.com/jescalan/autoprefixer-stylus/pull/180. That's why I
left the custom browserslist config for autoprefixer-stylus in
`build:css`, `build:css:utils` and `build:css:kss` scripts for now. But
we should upgrade this module and stop using a custom browserslist
config for it as soon as possible.

This fixes deprecation warnings from autoprefixer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -187,5 +187,8 @@
     "snapshotSerializers": [
       "enzyme-to-json/serializer"
     ]
-  }
+  },
+  "browserslist": [
+    "extends browserslist-config-cozy"
+  ]
 }
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
   "homepage": "https://github.com/cozy/cozy-ui",
   "scripts": {
     "build:css:all": "yarn build:css && yarn build:css:utils",
-    "build:css": "env CSSMODULES=false stylus -c --include stylus -o dist/cozy-ui.min.css stylus/cozy-ui/build.styl --use ./node_modules/autoprefixer-stylus --with \"{ browsers: ['last 2 versions'] }\"",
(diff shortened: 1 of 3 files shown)


drazik:
[autoprefixer-stylus v1.0.0](https://www.npmjs.com/package/autoprefixer-stylus) using autoprefixer@9.5.1 has been released. We can upgrade it and make it use our browserslist config.

ptbrowne:
> We can upgrade it and make it use our browserslist config.

Should we do this in this PR ?

drazik:
> > We can upgrade it and make it use our browserslist config.
> 
> Do you want to do it in this PR ?

Yes.

> Do you know why (if there is an advantage) of using autoprefixer-stylus and not directly using autoprefixer on the generated CSS ? I wonder if we should base ourselves onto this plugin.

I don't know it there is an advantage. I think we use it because it's easily usable as a stylus plugin. Maybe we can drop this dep and use autoprefixer directly.

y-lohse:
This PR still needs the update of autoprefixer-stylus, right?

drazik:
Yes, I just pushed this.

PR #1121, opened 2019-07-29
fix: Transpiled stylesheet contains utility classes

### LINKED ISSUE #1119: Transpiled components require to import utility css
[In the docs](https://github.com/cozy/cozy-ui/#as-a-components-library), we say that:

> If you use the transpiled components (from cozy-ui/transpiled/react), you need to import the stylesheet (once):
> ```
> import Button from 'cozy-ui/transpiled/react/Button'
> import 'cozy-ui/transpiled/react/stylesheet.css'
> 
> <Button />
> ```
> You're now ready to use Cozy UI's (p)React components

But this promise is false: when using some components (for example `Caption`), some styles are not applied because the needed utility classes are not included in the transpiled stylesheet.

Here is a repro: https://codesandbox.io/s/charming-jones-ut5ro. On this sandbox, you can see that `Text` and `Caption` styles are totally missing. And some utility classes used by `Infos` are also missing. This means that the user is forced to import the utility css (https://github.com/cozy/cozy-ui/#utility-classes) to get working components.

I think the transpiled CSS should contain all the utility classes used by the react components. But I don't find why some utility classes are included  and others are not.

This problem is not visible in the styleguide, we found it in banks on the `Caption` component.

drazik:
This should fix #1119. But some utility classes are likely to be
duplicated (since some of them were already present). More work is
needed to deduplicate them, but this fixes our base promise (transpiled
stylesheet contains everything the components use).

### DIFF
--- a/react/utilities.js
+++ b/react/utilities.js
@@ -2,3 +2,4 @@
 
 // Special file for utilities to be included in the transpiled stylesheet
 import palette from '../stylus/utilities/palette.styl' // 130k
+import utilies from '../stylus/cozy-ui/utils.styl'


ptbrowne:
I am not sure that further deduplication work is needed since postcss already deduplicates if I am not mistaken.

```
> echo ".a { background: red; } .a { background: red }" | npx postcss
.a{background:red}
```

and

```
$ postcss transpiled/react/stylesheet.css --replace
```

is already part of the transpilation.

LGTM :)

PR #1124, opened 2019-08-19
feat: Make radio and checkbox accessible

### LINKED ISSUE #1122: Hide native checkbox and radio in a more accessible way
Today we use `display: none` and `visibility: hidden` to hide checkboxes and radio buttons. This is not a good way to do this as these controls will be totally inaccessible for screen readers users.

It also make testing of checkbox and radio buttons hard in testcafé. It throws the following error: `The element that matches the specified selector is not visible`.

See https://webaim.org/techniques/css/invisiblecontent/ for techniques to hide content will staying accessible.

drazik:
In Radio and Checkbox components, inputs were hiddden using
`display: none` and `visibility: hidden`, which totally hide the
element. For accessibility reason, we actually want the inputs to be
only visually hidden, but still accessible to screen readers.

Fix #1122

For information, the technique used is from https://cloudfour.com/thinks/see-no-evil-hidden-content-and-accessibility/

I also added styles for the focused state.

See https://drazik.github.io/cozy-ui/react/#!/Checkbox and https://drazik.github.io/cozy-ui/react/#!/Radio

### DIFF
--- a/stylus/tools/mixins.styl
+++ b/stylus/tools/mixins.styl
@@ -74,7 +74,8 @@ hide()
 /*
     visually-hide()
 
-    visually-hide() mixin visually hides an element. The element is still reachable. This mixin doesn't take parameters.
+    visually-hide() mixin visually hides an element. The element is still visible by screen readers.
+    This mixin doesn't take parameters.
 
     Weight: 3
 
--- a/stylus/tools/mixins.styl
+++ b/stylus/tools/mixins.styl
@@ -72,12 +72,33 @@ hide()
     visibility  hidden !important  // @stylint ignore
 
 /*
+    visually-hide()
+
+    visually-hide() mixin visually hides an element. The element is still reachable. This mixin doesn't take parameters.
+
+    Weight: 3
+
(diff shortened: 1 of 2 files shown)


stylus/tools/mixins.styl
```diff
@@ -71,13 +71,34 @@ hide()
     display     none !important  // @stylint ignore
     visibility  hidden !important  // @stylint ignore
 
+/*
+    visually-hide()
+
+    visually-hide() mixin visually hides an element. The element is still reachable. This mixin doesn't take parameters.
```
ptbrowne:
```suggestion
    visually-hide() mixin visually hides an element. The element is still visible by screen readers.
    This mixin doesn't take parameters.
```

y-lohse:
Code LGTM, but on your example, the checkbox in their checked state have… well they don't have a check mark?

ptbrowne:
Un autre bug de UI, l'icône utilisée ne semble pas être la bonne s/check-circleless/check, je suis en train de checker.

PR #1127, opened 2019-08-20
feat: Customize Field label props

y-lohse:
The Field component renders a Label and an Input, but we don't have a lot of control over them. We already had a `fieldProps` prop that allowed to pass props down to the Input, and this PR does the same for the label.

Not super fond of piling on props like this, but the Field also has a built-in logic to toggle a password field between password and plain text that we would have to reimplement otherwise :(

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -79,6 +79,7 @@ const Field = props => {
     className,
     disabled,
     fieldProps,
+    labelProps,
     fullwidth,
     label,
     id,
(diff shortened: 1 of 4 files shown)


react/Field/index.jsx
```diff
@@ -245,6 +250,7 @@ Field.propTypes = {
 
 Field.defaultProps = {
   fieldProps: {},
+  labelProps: {},
```
ptbrowne: Nit: No need for default object here since labelProps is not required (spreading `...undefined` will not result in an error)

PR #1128, opened 2019-08-21
feat(Field): Custom secondaryComponent

Crash--:
We can now pass a secondaryComponent prop to InputPassword
in order to manage ourself the component.

Also edited a few css lines, to fix an issue where the
"hide"/"show" was capturing the click on the all width

https://crash--.github.io/cozy-ui/react/#!/Field/21

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -34,12 +34,14 @@ class InputPassword extends React.Component {
       showLabel,
       showVisibilityButton,
       fullwidth,
+      secondaryComponent,
       ...restProps
     } = this.props
     const { visible } = this.state
+
     return (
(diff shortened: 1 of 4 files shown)


ptbrowne:
On ne peut pas taper dans ce Field ?

react/Field/index.jsx
```diff
@@ -51,6 +53,18 @@ class InputPassword extends React.Component {
             {visible ? hideLabel : showLabel}
           </div>
         )}
+        {showVisibilityButton && secondaryComponent && (
```
ptbrowne: Nit: I think the naming is not optimal here : showSecondaryComponent ou visibilityComponent mais pas les deux ?

PR #1132, opened 2019-08-23
fix: Min/max none

y-lohse:
[`none` is not a valid value for `min-width` and co.](https://developer.mozilla.org/en-US/docs/Web/CSS/min-width). I've replaced `none` with `auto`. The previous `u-miw-none` classes will work now, but are deprecated — please use `u-miw-auto` instead.


ptbrowne:
This is also something that should be detected by csslint, see #1136 .

PR #1141, opened 2019-08-28
feat: Spacing values are a bit more spaced

ptbrowne:
Stack spacing values were so close that they were useless. I figured that since it is still very new, we could do a breaking change here.

@joel-costa  I am interested in your input here for vertical spacing.

The styleguide is visible here: https://ptbrowne.github.io/cozy-ui/react/#/Stack

BREAKING CHANGE: sparse/dense spacing values are no longer recognized

### DIFF
--- a/react/Stack/Readme.md
+++ b/react/Stack/Readme.md
@@ -17,28 +17,44 @@ const Card = require('../Card').default;
 </Stack>
 ```
 
-You can use `sparse` and `dense` value for "spacing" to have more/less padding.
+You can use `xs`, `s`, `l`, `xl`, and `xxl` values for "spacing" to have less/more padding.
 
 ```
-const Card = require('../Card').default;
+const Card = require('../Card').default
+const spacings = ['xs', 's', 'l', 'xl', 'xxl'];
 
 <div>
-<p>
-Dense:
-</p>
-<Stack spacing="dense">
-  <Card>Homer Simpson</Card>
-  <Card>Marge Simpson</Card>
-</Stack>
 
-<p>
-Sparse:
-</p>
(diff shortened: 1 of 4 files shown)


react/Stack/Readme.md
```diff
@@ -17,25 +17,35 @@ const Card = require('../Card').default;
 </Stack>
 ```
 
-You can use `sparse` and `dense` value for "spacing" to have more/less padding.
+You can use `s` and `l` and `xl` values for "spacing" to have less/more padding.
```
drazik:
```suggestion
You can use `s` and `l` and `xl` values for "spacing" to have less/more margin.
```

ptbrowne:
> I'm not sure what kind of input you're looking for about the spacing?

I am wondering if @joel-costa already has some typical margins with names, it would be good that we use the same.

ptbrowne:
OK, for info I updated the PR after a call with Joel, the idea is that we use those spacing values in other files where we use spacing, for example utility classes.

react/Stack/styles.styl
```diff
@@ -1,20 +1,18 @@
...
-.Stack > * + *
-    margin-top .5rem
+spacing_values = {
+    xs: .5rem,
+    s: .75rem,
+    m: 1rem, // the standard, no modifier needed
+    l: 1.5rem,
+    xl: 2rem
+    xxl: 3rem
+}
```
drazik: If I understand well, this should be removed to use the map from `utilities/spaces.styl`

ptbrowne: Thanks :) This was indeed the idea, and I mixed up :).

PR #1146, opened 2019-08-30
fix(Infos): Fix the text propTypes

Crash--:
In a few use cases, we pass to infos a ReactMardown component
So it's not a string but an object

### DIFF
--- a/react/Infos/index.jsx
+++ b/react/Infos/index.jsx
@@ -71,7 +71,8 @@ Infos.propTypes = {
   icon: iconPropType,
   /** An important information will be displayed with red colors */
   isImportant: PropTypes.bool,
-  text: PropTypes.string,
+  /** Can be either a Text, or an element (for instance a ReactMardown component) */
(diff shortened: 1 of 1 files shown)


react/Infos/index.jsx
```diff
@@ -71,7 +71,8 @@ Infos.propTypes = {
   icon: iconPropType,
   /** An important information will be displayed with red colors */
   isImportant: PropTypes.bool,
-  text: PropTypes.string,
+  /** Can be either a Text, or an object (for instance a ReactMardown component) */
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
ptbrowne: I think `element` instead of `object` would be better  ? 

https://fr.reactjs.org/docs/typechecking-with-proptypes.html

```
  // Un élément React.
  optionalElement: PropTypes.element,
```

Crash--: thanks

PR #1151, opened 2019-09-10
Add CipherIcon component

drazik:
This icon will be used in various places, so I extract it from cozy-harvest-lib to cozy-ui now.

This is a konnector icon with a keychain icon: 
image


react/CipherIcon/styles.styl
```diff
@@ -0,0 +1,17 @@
...
+@require 'settings/palette'
+@require 'tools/mixins'
+
+.c-card {
+    border rem(1) solid var(--silver)
+    border-radius rem(8)
+    padding rem(16)
+    text-decoration none
+    display block
+}
```
ptbrowne: Nit: I think you can remove the curly braces

drazik: Actually I forgot to remove this file after `cp` from another component

react/CipherIcon/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  fetchIcon() {
+    const { client, konnector } = this.props
+
+    return client.stackClient.getIconURL({
+      type: 'konnector',
+      slug: konnector.slug || konnector
+    })
+  }
```
ptbrowne: In https://github.com/cozy/cozy-ui/pull/1152 I will work so that user provided fetchIcon gets props automatically. This will remove the need to have to bind the fetchIcon in the constructor and will enable fetchIcon to be completely extracted.

```
// In another file
const fetchKonnectorIcon = ({ client, konnector }) => 
   client.stackClient.getIconURL({
      type: 'konnector',
      slug: konnector.slug || konnector
    })

const CipherIcon = React.memo(({ className }) => (
      <div className={cx('u-dib u-pos-relative', className)} {...rest}>
        <div className="u-pos-absolute u-right-0 u-bottom-0">
          <AppIcon fetchIcon={fetchKonnectorIcon} className="u-w-1 u-h-1" />
        </div>
        <Icon icon="keychain" size={32} />
      </div>
    ))
```

react/CipherIcon/styles.styl
ptbrowne: This styles are not used in this PR right ?

react/CipherIcon/index.jsx
drazik: :+1:

PR #1152, opened 2019-09-10
WIP feat: Memoize fetchIcon

ptbrowne:
Do not review now.

Goals:

- Automatically memoize fetchIcon even if it is provided by the user.
- fetchIcon should be called with the props of the current component

See https://github.com/cozy/cozy-ui/pull/1151/files#r322771963

### DIFF
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -1,64 +1,20 @@
-const _loaded = {}
-
-export const preload = async (app, domain, secure) => {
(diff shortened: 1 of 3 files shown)


nono:
Ping @ptbrowne

ptbrowne:
Mmh I still think this could be a good thing.

Crash--:
closing it since we rewrote this component few month ago using cozy-client and its memoization

PR #1157, opened 2019-09-17
feat: Automated visual regression tests

ptbrowne:
Automated visual regression on components inside styleguidist.

1. Styleguidist is built after test and build have passed.
2. Puppetter opens the styleguidist, and screenshots all components individually
   to the `screenshots` directory
3. Screenshots are uploaded to ArgosCI (by the team behind materialUI, does
   the diff of the screenshots between master and the current branch)
4. A bot should tell us if there are visual differences

Notes:

- At material-ui, they blacklist some components if their visual regressions tests
  are too flaky. I think we will be able to do that when we see that we have
  flaky visual regressions.

See https://www.argos-ci.com/cozy/cozy-ui/builds/54259 for an example of visual failure.

I haven't yet found how to set up the bot, see my question here https://github.com/argos-ci/argos/issues/405


scripts/screenshots.js
```diff
@@ -0,0 +1,140 @@
+const puppeteer = require('puppeteer')
+const path = require('path')
+const fs = require('fs')
+const sortBy = require('lodash/sortBy')
+
+const DEFAULT_PUPPETEER_VIEWPORT = {
+  width: 800,
+  height: 600
+}
```
Crash--: Are we sure to want this resolution as default? We should use the most used resolution on our product

.travis.yml
```diff
@@ -11,9 +11,12 @@ script:
 - yarn lint
 - yarn test
 - yarn build:css:all
+- yarn build:doc:react
+- yarn screenshots
+- yarn argos --token $ARGOS_TOKEN --branch $TRAVIS_BRANCH --commit $TRAVIS_COMMIT
```
Crash--: So you have configured on the `argos` side that `master` is the "default" and like that `argos` will check the diff between `master` and `$TRAVIS_BRANCH` ? 

Can't wait to have a PDC on that subject :)

scripts/screenshots.js
ptbrowne: This is the default resolution of puppeeter so that we can reset the viewport to the original size. Then, for each component the viewport is scaled to accomodate the full height of the examples.

.travis.yml
ptbrowne: Argos detects that master is the default by looking at the base branch of the repository.

scripts/screenshots.js
ptbrowne: We could set the viewport to the most used resolution or another resolution but this was not what this code was meant to do.

scripts/screenshots.js
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Screenshot a component to the screenshot directory, taking care of
+ * resizing the viewport before-hand so that the viewport fits the
+ * component.
+ */
+const screenshotComponent = async (page, { link, name }, screenshotDir) => {
+  await page.goto(link, { waitUntil: 'load', timeout: 0 })
+
+  // Getting the dimensions of the root wrapper
+  const rootBcr = await page.evaluate(() => {
```
y-lohse: What does `Bcr` stand for?

scripts/screenshots.js
```diff
@@ -0,0 +1,140 @@
...
+    const { width, height } = document
+      .querySelector('#rsg-root')
+      .getBoundingClientRect()
+    return { width, height }
+  })
+
+  // Fitting the viewport to the component
```
y-lohse: If the root wrapper is responsive, won't we end up with screenshots that are always the same resolution, regardless of the component?

scripts/screenshots.js
ptbrowne: bounding client rect

scripts/screenshots.js
ptbrowne: The width will not changed, only the height changes to accomodate the component example if it is too long to be displayed in the default 800x600 viewport.

ptbrowne:
> Do they live on ArgosCI somehow?

Yes, they live there.

PR #1158, opened 2019-09-17
fix: Add example for PageLayout in styleguide

ptbrowne:
The example had not been added to the styleguide.
Since the behavior is only shown on mobile, I only provided
a static example.

EDIT: Add viewable example URL https://ptbrowne.github.io/cozy-ui/react/#/PageLayout


ptbrowne:
It is a static example so you will only see the code :)

y-lohse:
Aah, I didn't understand what you meant with "static example" but yeah, cool 👍  Thanks for the upload!

Issue #1165, opened 2019-09-23
Info container can shrink


ptbrowne:
Info container can shrink and this leads to unfortunate display bugs.

<img width="820" alt="image" src="https://user-images.githubusercontent.com/465582/65420636-f8157480-de01-11e9-8b67-c8339c8d1939.png">

PR #1166, opened 2019-09-23
No shrink for Infos

ptbrowne:
Fix for https://github.com/cozy/cozy-ui/issues/1165, setting `flex-shrink: 0` prevents the Info container to be shrinked.


drazik:
`Info` is not always used in flex context. I think it's the app's role to know that in a particular context `Info` should not shrink and pass it the `u-flex-shrink-0` class name.

ptbrowne:
Yes you are right, thanks !

PR #1167, opened 2019-09-23
feat: Set percentage in example to fixed value

ptbrowne:
Otherwise, Argos sees difference every time

### DIFF
--- a/react/PercentageLine/Readme.md
+++ b/react/PercentageLine/Readme.md
@@ -1,8 +1,8 @@
 ```
-initialState = { percentage: Math.random() * 100 };
+initialState = { percentage: 0.5 * 100 };
 
 <>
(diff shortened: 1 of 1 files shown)


react/PercentageLine/Readme.md
```diff
@@ -1,8 +1,8 @@
 ```
-initialState = { percentage: Math.random() * 100 };
+initialState = { percentage: 0.5 * 100 };
 
 <>
     <button onClick={() => setState({ percentage: Math.random() * 100})}>Random percentage</button>
-    <PercentageLine  color='var(--primaryColor)' value={state.percentage}/>
+    {(state.percentage).toFixed(2)}% <PercentageLine  color='var(--primaryColor)' value={state.percentage}/>
```
Crash--: TIL `.toFixed` :o

PR #1168, opened 2019-09-23
fix: Can not read propTypes of undefined

Crash--:
I didn't take the time to dig on this one, but in Drive, we had a failing test because of that error. I don't know how we can have an `undefined` as `WrappedComponent` but this is the case 

This is a TMP fix, but since this regression was introduced by this commit https://github.com/cozy/cozy-ui/commit/91aa079 I'll dig into it later

### DIFF
--- a/react/I18n/index.jsx
+++ b/react/I18n/index.jsx
@@ -85,7 +85,8 @@ export const translate = () => WrappedComponent => {
     )
   }
   Wrapper.propTypes = {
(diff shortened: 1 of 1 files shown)


ptbrowne:
I think it would be better to find the root cause here.

Crash--:
I'll in a few days. This issue and breaking change was introduced in a minor version of UI and never documented. So I really think we should fix it ASAP here.

PR #1171, opened 2019-09-24
fix(navitem): SVG is now well displayed on mobile

Crash--:
Since https://github.com/cozy/cozy-ui/pull/1083/files
there is a regression with the bottom nav bar on mobile.
 SVG and text were not correctly displayed.

So this fix is basically a kind of revert for medium_screen() 

Capture d’écran 2019-09-24 à 17 31 14
Capture d’écran 2019-09-24 à 17 31 29

### DIFF
--- a/stylus/components/nav.styl
+++ b/stylus/components/nav.styl
@@ -62,14 +62,12 @@ $nav-item-icon
 
     :hover > &
         color var(--charcoalGrey)
-
     +medium-screen()
         display block
-        font-size rem(24)
         margin-right 0
 
         svg
-            margin 0 auto
+            margin 0 auto 3px auto
             width rem(24)
             height rem(24)
 
@@ -113,7 +111,7 @@ $nav-link
         text-align           center
         color                var(--slateGrey)
         font-size            rem(10)
-        line-height          1
+        line-height          rem(12)
         background-position  center top
         background-size      rem(24)
 
--- a/stylus/components/nav.styl
+++ b/stylus/components/nav.styl
@@ -67,7 +67,7 @@ $nav-item-icon
         margin-right 0
 
         svg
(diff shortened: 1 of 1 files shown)


stylus/components/nav.styl
```diff
@@ -62,7 +62,9 @@ $nav-item-icon
 
     :hover > &
         color var(--charcoalGrey)
-
+    +small-screen()
+        svg
+            display inline
```
ptbrowne: I don't understand what is going on here ? As a guess, the padding between the icon and the text comes from the line-height here?

Crash--: Woo, this is not exactly what I wanted to commit... I wanted to add this display inline here : https://github.com/cozy/cozy-ui/blob/7881ef467b8ba651ee654ade3346fe1a3bcf5880/stylus/components/nav.styl#L73 since we are already doing few stuff with the svg.

ptbrowne: Still I do not exactly understand how putting inline fixes the padding between the icon and the text, is it because of the line-height ?

Crash--:
Au final j'ai ajouté le margin bottom et corrigé une line height pour que ça corresponde plus aux spec de design

stylus/components/nav.styl
```diff
@@ -62,14 +62,12 @@ $nav-item-icon
...
         color var(--charcoalGrey)
-
     +medium-screen()
         display block
-        font-size rem(24)
         margin-right 0
 
         svg
-            margin 0 auto
+            margin 0 auto 3px auto
```
ptbrowne:
```suggestion
            margin 0 auto 3px
```

will fix the red travis

ptbrowne:
LGTM, FYI you can go to Argos and accept the change.

PR #1174, opened 2019-09-26
fix(ActionMenuItem): Handle onClick on Item

Crash--:
Before this commit, if we attached an onClick to an
ActionMenuItem, the onClick was not called since
the propagation of the event was not stopped
resulting in the close of the menu instead of the
call of the onClick props.

With this fix, we stop the propagation if an onClick
is here, and we call it after.

https://crash--.github.io/cozy-ui/react/#actionmenu

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -167,9 +167,17 @@ const ActionMenuHeader = ({ children }) => {
   return <div className={styles['c-actionmenu-header']}>{children}</div>
 }
 
-const ActionMenuItem = ({ left, children, right }) => {
+const ActionMenuItem = ({ left, children, right, onClick }) => {
+  const onClickEnhanced = e => {
+    if (onClick) {
+      // we need to stop propagation here so that the menu doesn't close itself
+      e.stopPropagation()
+      onClick()
+    }
+  }
+
   return (
-    <Media className={styles['c-actionmenu-item']}>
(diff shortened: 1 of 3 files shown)


Crash--:
I think it's the responsibility of the caller to deal with : 
```jsx
 <ActionMenuItem
              onClick={() => {
                alert('test')
                this.toggleMenu()
              }}
            >
```

WDYT?

drazik:
You're right :)

react/ActionMenu/index.jsx
```diff
@@ -167,15 +167,29 @@ const ActionMenuHeader = ({ children }) => {
...
       {left && <Img className="u-mh-1">{left}</Img>}
       <Bd className={left ? 'u-mr-1' : 'u-mh-1'}>{children}</Bd>
       {right && <Img className="u-mr-1">{right}</Img>}
     </Media>
   )
 }
 
+ActionMenuItem.propTypes = {
+  left: PropTypes.bool,
+  right: PropTypes.bool,
```
drazik: `left` and `right` seem to be `PropTypes.node` since they are rendered (see https://github.com/cozy/cozy-ui/pull/1174/files#diff-38b8cc61b38494873608fee164497ad7R181 and https://github.com/cozy/cozy-ui/pull/1174/files#diff-38b8cc61b38494873608fee164497ad7R183)

Crash--: Oh right! Thanks. Fixed.

PR #1181, opened 2019-10-02
feat(utilities): Add 1-half to dimensions scale

drazik:
For the needs of the password manager's read-only identifier field, we need  an icon with a `1.5rem` width (see https://zpl.io/VD1N3rr). This step was not existing on the dimensions scale, which surprised me since we have it in margin/padding scales. This PR adds it.

I didn't add others « half » steps that are present for margin and padding but not for dimensions (`half` and `2-half`). Do you think we should add it now?

See https://drazik.github.io/cozy-ui/styleguide/section-utilities.html#kssref-utilities-dimensions

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -1745,8 +1745,11 @@ Display an chip that represents complex identity
     [Can be used with breakpoints modifier suffixes (`-t`,`-s`,`-m`)](section-settings.html#kssref-settings-breakpoints)
 
     .u-w-auto - Cancel width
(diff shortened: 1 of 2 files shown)


Crash--:
> Do you think we should add it now?

I think we should (my POV)

drazik:
> I am curious to know if it is the same size in mobile/desktop ?

In my current case, it's the same size in mobile and desktop.

drazik:
Added `half`, `1-half` and `2-half` to `min/max-widht/height` and `width/height` scales.

PR #1182, opened 2019-10-02
feat(Icon): Add bicolor file icon

Crash--:
See `file-bicolor` here: https://crash--.github.io/cozy-ui/react/#!/Icon/3

The idea is to have a `file` icon that we can fill with one color and use an opacity on that color to have a second color. 

To be able to do that, we need to have in the SVG two times the same SVG : 
- One filled in #FFF
- One that will be filled later 

Why ? 

Because if we have a background on the node we want to use this Icon, the opacity will not be correct. Adding a blank factor acts like a reset of the background. 

https://crash-drive.mycozy.cloud/public?sharecode=oQk6rR7TGDzB (I can't upload it on github, I don't know why) 
credit goes to @ptbrowne for this gorgeous img


y-lohse:
Ok the trick is cool, but i'm not super fond of the name `file-bicolor`. What is it used for, and why do we need a different variation than the normal `file` icon?

Crash--:
Current `File` is only about 1 color. It's similar to `file-type-files` but without the hardcoded colors. 

`file-type-files` can't be overridden and I didn't want to create a breaking change just for a colored icon.

y-lohse:
Ok I think I understand the purpose of that new icon. 

So it's just the name now. I don't think `bicolor` is really a word :D. A few suggestions:

- file-shadow or file-shaded (as opposed to the "flat" version that we already have)
- file-duotone
- and uh… that's all I've got

ptbrowne:
For what it's worth I think the "tone" is more appropriate here: two-tone or duotone seems good.

Crash--:
Let's go for file-duotone thanks

PR #1184, opened 2019-10-02
Add ModalBackButton

drazik:
Before adding the new `ModalBackButton` component, I splitted the `Modal/index.jsx` file into multiple subcomponents files. I found the file was pretty large and thought it was better to split it.

The `ModalBackButton` is a button that position itself in the modal header. It is useful to handle multi-step processes inside a Modal.

See https://drazik.github.io/cozy-ui/react/#modal

Edit : actually it seems there's a bug on the modal's cross placement on the online styleguide. When watching the doc locally, placement is OK. We already have this bug on https://docs.cozy.io/cozy-ui/react/#modal. I'm going to look at why we have this bug.


ptbrowne:
THe order of the CSS in the styleguidist seems to be the culprit, see https://github.com/cozy/cozy-ui/issues/335#issuecomment-483283437

drazik:
Thanks @ptbrowne. Since it's a known issue about the styleguide only, I'm in favor of merging this PR as is if it's OK for you, and have a look a this specific issue in a separate PR.

react/Modal/ModalBackButton.jsx
```diff
@@ -0,0 +1,37 @@
...
+  const {
+    breakpoints: { isMobile },
+    className,
+    ...rest
+  } = props
+
+  return (
+    <Button
+      icon="previous"
+      iconOnly
```
Crash--: Can you add a label plz ?

drazik: Did you mean accessibility instead of i18n? Actually since the props are spread, the underlying Button will receive the label that is passed to the ModalBackButton from the app. But I added a default "back" label.

Crash--: yes for accessibility sorry :/ Thanks

PR #1185, opened 2019-10-07
feat: Menu improvements

### LINKED ISSUE #1169: Menu space is limited by its opener
https://docs.cozy.io/cozy-ui/react/#!/Menu

image

I guess portaling the Menu would solve the problem.

### LINKED ISSUE #1170: Adapt black colors for Menu
Icon color and text should have slighty different shades of grey.

https://mattermost.cozycloud.cc/cozycloud/pl/93stn1zcu38wucd4z58zz6ub9r

Icon:  slate-grey
Text: charcoal-grey

ptbrowne:
- Fix the colors (fix #1170), charcoalGrey for text, slateGrey for icon
- Menu used in examples was the one from material-ui, styleguidist got confused when
  auto-providing the `Menu` variable inside the examples. Requiring explicitly fixes
  it (fix #1169).
- Have the menu automatically opened for Argos (provided the isArgos() util that can
  be used in examples for this kind of behavior. Based on user-agent)

Preview here : https://ptbrowne.github.io/cozy-ui/react/#/Menu

### DIFF
--- a/docs/utils.js
+++ b/docs/utils.js
@@ -27,7 +27,12 @@ export const fakeIntentCreate = (action, doctype, options) => {
       })
 
       ReactDOM.render(
-        React.createElement(IntentExample, { onComplete, action, doctype, options }),
+        React.createElement(IntentExample, {
+          onComplete,
+          action,
+          doctype,
+          options
+        }),
         iframe.contentDocument.body
       )
     }
@@ -37,6 +42,6 @@ export const fakeIntentCreate = (action, doctype, options) => {
   return p
 }
 
-export const t = (path) => {
+export const t = path => {
   return get(translations, path)
 }
--- a/react/Menu/index.jsx
+++ b/react/Menu/index.jsx
@@ -34,7 +34,12 @@ class MenuItem extends Component {
 }
 
 class Menu extends Component {
-  state = { opened: false }
+  constructor(props, context) {
+    super(props, context)
+    this.state = {
+      opened: props.initialOpen
+    }
+  }
 
   toggle = () => (this.state.opened ? this.close() : this.open())
 
@@ -156,7 +161,8 @@ Menu.defaultProps = {
   onSelect: null,
   onSelectDisabled: null,
   itemsStyle: {},
-  popover: false
+  popover: false,
+  initialOpen: false
 }
 
 Menu.propTypes = {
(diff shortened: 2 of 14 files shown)


react/Menu/index.jsx
```diff
@@ -124,10 +131,10 @@ class Menu extends Component {
         {...restProps}
       >
         {!component ? (
-          <MenuButton
+          <Button
             disabled={disabled}
             onClick={this.toggle}
-            label={text}
+            label={text || label}
             buttonClassName={buttonClassName}
```
Crash--: are we sure that we have in this `buttonClassName` the equal of : `className={cx('c-btn', styles['c-menu__btn'], buttonClassName)}` ? 🤔 

you say that there is no style but what about : 
```
.c-menu__btn
    font-size rem(16)
    border 0
    cursor pointer

.c-menu__btn:not([disabled]):not([aria-disabled=true]):hover
.c-menu__btn:not([disabled]):not([aria-disabled=true]):active
.c-menu__btn:not([disabled]):not([aria-disabled=true]):focus
    background-color transparent
```

react/Menu/index.jsx
```diff
@@ -175,8 +181,10 @@ Menu.propTypes = {
   onSelectDisabled: PropTypes.func,
   /** Global Styles for MenuItems */
   itemsStyle: PropTypes.object,
-  /** if you need fixed menu */
-  popover: PropTypes.bool
+  /** If you need fixed menu */
+  popover: PropTypes.bool,
+  /** Whether the menu should be initially opened */
+  initialOpen: PropTypes.bool
```
Crash--: SelectBox has a `menuIsOpen` prop. Should we keep the same name?

docs/utils.js
```diff
@@ -40,3 +40,7 @@ export const fakeIntentCreate = (action, doctype, options) => {
 export const t = (path) => {
   return get(translations, path)
 }
+
+export const isArgos = () => {
+  return navigator.userAgent.includes('Argos')
+}
```
Crash--: It seems that we have a `isTesting` variable to take "open" snapshot. Should we set `isTesting` to true when we are in a jest context but also when the userAgent is Argos? Like that no need to do : 

`menuIsOpen={isArgos() || isTesting}`

react/Menu/index.jsx
ptbrowne: I don't think so, menuIsOpen controls the component but here we do not control it. It is just the initial value.

docs/utils.js
ptbrowne: Correct, I made the change, thanks

react/Menu/index.jsx
ptbrowne: I removed the commit removing the MenuButton from this PR.

PR #1186, opened 2019-10-07
feat: Remove MenuButton

ptbrowne:
- [ ] ⚠️ Change base branch to master once menu-improvements is merged

BREAKING CHANGE
I think nobody used it, the style is very bare

Is there anybody using the MenuButton, otherwise I think we should remove it.


Crash--:
We're using it in Drive. https://github.com/cozy/cozy-drive/blob/master/src/components/Menu/Menuawaremobile.jsx#L31 

This MenuAwareMobile should be part of UI and maybe a props to the Menu itself.

ptbrowne:
I am closing since Drive is using it right now.

PR #1190, opened 2019-10-09
Composite row

ptbrowne:
Add to styleguide different variants of composite row layouts that Joel has on Sketch/Zeplin.

Visible on https://ptbrowne.github.io/cozy-ui/react/#/CompositeRow

- [x] Montant doit être aligné au milieu avec le Primary/Secondary text
- [x] Icône ne doit pas dépasser du cadre même quand la card est petite
- [x] Sans secondary text, le Primary text doit être aligné au milieu avec l'icône

~~I find the implementation a bit verbose, I will try to use utility classes instead of Bd/Img/Media to have less nesting.~~ With the help of utility classes, nesting is reduced, see https://github.com/cozy/cozy-ui/pull/1190/commits/4d76fc38209eb98d6b308431563e29de53d4949c

I also want to gather feedback on the following API: 

Proposition of names for the component (with my preference next to it):

- ComplexRow -
- RichRow +
- InfoRow -
- CompositeRow ++
- ListItem ++ (coherence with MaterialUI)

Proposition of props:

<p align='center'>image</p>

- I hesitate on "chips" since we maybe would want to have another kind of components there, do you have any thoughts on this @joel-costa  ? Alternatives

  - `chips` - 
  - `belowLeft` +
  - `extra` -
  - `belowRight` +

I think `belowLeft` and `belowRight` are actually better since they are more generic and at the same time, they are more descriptive.

Utility classes introduced:

- u-media, u-media-grow, u-media-fixed (Bd -> u-media-grow, Img -> u-media-fixed)

This is more flexible than Bd, Img since you can compose classes without introducing more divs

```patch
 <Media>
-  <Bd>
-    <Media>
-    </Media>
-  </Bd>
+  <div class='u-media-grow u-media'>
+  </div>
 </Media>
```

- u-row + variants (like u-stack but horizontal) cc @Crash-- ;)

Related work:

- MaterialUI ListItem : https://material-ui.com/components/lists/

### DIFF
--- a/docs/webpack.config.js
+++ b/docs/webpack.config.js
@@ -3,6 +3,9 @@ const webpack = require('webpack')
 
 module.exports = {
   resolve: {
+    alias: {
+      docs: __dirname
+    },
     extensions: ['.jsx', '.js', '.json', '.styl']
   },
   module: {
--- a/react/CompositeRow/Readme.md
+++ b/react/CompositeRow/Readme.md
@@ -1,0 +1,96 @@
+```js
+import { Media, Bd, Img } from '../Media'
+import UIChip from '../Chip'
+import { Bold, Caption } from '../Text'
+import Icon from '../Icon'
+import Circle from '../Circle'
+
+import Variants from 'docs/components/Variants'
+
+const fnacLogoURL = 'https://upload.wikimedia.org/wikipedia/commons/2/2e/Fnac_Logo.svg';
+
+const Chip = React.memo(({ children, ...props }) => (
+  <UIChip
+    variant="outlined"
+    className='u-mr-0 u-mb-0'
+    size='small'
+    children={children} {...props}
+  />
+))
+
+const ChipImage = React.memo(({ src }) => (
+  <img className='u-mr-half' src={src} height='50%' />
+))
+
+const compositeStyle = {
+  boxShadow: '0 0 10px rgba(0, 0, 0, 0.15)',
+  boxSizing: 'border-box', 
+}
+
+
+const CompositeImage = () => (
+  <Circle backgroundColor='var(--melon)'>
+    <Icon icon='wallet' />
+  </Circle>
+)
+
+const Amount = () => (
+  <Bold tag='span'>12,15€</Bold>
+)
+
+const onDotsClick = () => { alert('clicked dots !')}
+const Dots = () => (
+  <Icon className="u-c-pointer" color="var(--coolGrey)" icon="dots" onClick={onDotsClick}/>
+)
+
+const Chips = () => (
+  <Bd className='u-row-xs'>
+    <Chip>
+      <ChipImage src={fnacLogoURL} /> Fnac
+    </Chip>
+    <Chip>
+      <ChipImage src={fnacLogoURL} /> Fnac
+    </Chip>
+  </Bd>
+)
+
+const Actions = ({ extraInformation }) => (
+  <Media>
+    <Bd className='u-row-xs'>
+      <Chip>
+        <ChipImage src={fnacLogoURL} /> Fnac
+      </Chip>
+      <Chip>
+        <ChipImage src={fnacLogoURL} /> Fnac
+      </Chip>
+    </Bd>
+    { extraInformation ? <Shrink><Caption>Extra information to the right of the card</Caption></Shrink> : null }
+  </Media>
+)
+
+const Shrink = Img
+
+const initialVariants = [
+  { chips: true, secondaryText: true, dots: true, icon: true, extraInformation: true },
+  { chips: false, secondaryText: true, icon: true, dots: true },
+  { dense: true, chips: false, secondaryText: false, dots: false, icon: true },
+  { chips: false, secondaryText: false, dots: false, icon: false },
+  { dense: true, chips: false, secondaryText: false, dots: false, icon: false }
+];
+
+<Variants initialVariants={initialVariants}>{
+  variant => (
+    <CompositeRow
+      primaryText='Primary text'
+      secondaryText={variant.secondaryText ? 'Secondary Text' : null}
+      image={variant.icon ? <CompositeImage /> : null}
+      right={<>
+        <Shrink><Amount /></Shrink>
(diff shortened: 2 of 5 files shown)


y-lohse:
Haven't reviewed the code yet, but in terms of API

- I really like CompositeRow
- i'm not sure about `chips` and `right`. `right` is the only positional prop in the list. I would suggest `primaryAction` and `secondaryAction` instead of `right` and `chips` respectively, inspired by material UI.

react/CompositeRow/Readme.md
```diff
@@ -0,0 +1,85 @@
...
+];
+
+const onChangeVariant = (updatedVariant, i) => {
+  const newVariants = [
+    ...state.variants.slice(0, i),
+    updatedVariant,
+    ...state.variants.slice(i + 1)
+  ]
+  setState({ variants: newVariants })
+};
```
y-lohse: Not blocking, but I wonder if having only one interactive example wouldn't be better then a series of variants that are all just different default options. As a dev I think I'd rather have ust one example to play with, but curious to see what others think.

y-lohse:
> i'm not sure about chips and right. right is the only positional prop in the list. I would suggest primaryAction and secondaryAction instead of right and chips respectively, inspired by material UI.

After reviewing the code, I realize that the dot icon is sort of hardcoded. I think it may be better to leave the whole right hand area as a render prop, including the action icon. That way the icon can be customized, and it makes it easier to integrate a dropdown menu for example.

docs/webpack.config.js
```diff
@@ -3,6 +3,9 @@ const webpack = require('webpack')
 
 module.exports = {
   resolve: {
+    alias: {
+      docs: __dirname
```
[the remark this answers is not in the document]
ptbrowne: For https://github.com/cozy/cozy-ui/pull/1190/files#diff-7469a98ccbf86686a91f570acd329357R6

react/CompositeRow/Readme.md
ptbrowne: OK for one interactive example to play with, I was mainly worried about getting various examples into screenshots, and it was also easier when  developing to see all possible variants in one shot.

- For testing, it is possible to check `isTesting()` so we can show all variants in one go
- For developing, I think we should have a way to quickly toggle all variants (by modifyying the source code)

What I suggest is another `<Variants />` helper that can show either the first variant or either all variants depending on a prop.

react/Media/styles.styl
```diff
@@ -15,3 +15,22 @@
 .img
     line-height 0
     flex 0 0 auto
+
+:global
```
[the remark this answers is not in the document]
ptbrowne: Because I find the way utility classes are declared hard to do :) I put it like that to see if `.u-media-grow` would stick and it pleased other people.

ptbrowne:
Yes for hardcoded the dot, this is not supposed to be the final API, I did it quick to have feedback ;)

react/CompositeRow/Readme.md
ptbrowne: For information, I just saw that https://material-ui.com/components/lists/#interactive is showing 4 examples at once. Do you have any preference @drazik, @Crash-- ?

ptbrowne:
I just saw https://material-ui.com/components/lists/ , I should have gone there a bit sooner, I changed `shrinked` to `dense` to mirror their API.

Crash--:
I really like `ListItem` because what we can do to follow closer M-UI, we should do it

react/CompositeRow/Readme.md
Crash--: Personally, the more exemples I can have, the more exemples I take. It helps me to really think about what variante I should use instead of having to test them before

drazik: I also think we should have more than one example like MUI does.

ptbrowne:
I think we can remove `extra` and use `secondaryAction`.

react/CompositeRow/Readme.md
y-lohse: Ok let's leave it with multiple examples, looks like I'm in the minority here and it's mostly a matter of taste I guess :)

ptbrowne: Done

react/CompositeRow/Readme.md
```diff
@@ -0,0 +1,95 @@
...
+  <Media>
+    <Bd className='u-row-xs'>
+      <Chip>
+        <ChipImage src={fnacLogoURL} /> Fnac
+      </Chip>
+      <Chip>
+        <ChipImage src={fnacLogoURL} /> Fnac
+      </Chip>
+    </Bd>
+    { extraInformation ? <Shrink><Caption>Coucou ca va bien ?</Caption></Shrink> : null }
```
ptbrowne:
```suggestion
    { extraInformation ? <Shrink><Caption>Extra information to the right of the card</Caption></Shrink> : null }
```

react/CompositeRow/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  secondaryText: PropTypes.element,
+  /** Image to the left of the row */
+  image: PropTypes.element,
+  /**
+   * Actions are shown below primary and secondary texts. Pass fragment for multiple elements.
+   * Good to use with Chips.
+   */
+  actions: PropTypes.element,
+  /** Row height will be fixed to 48px */
+  dense: PropTypes.bool
```
y-lohse: The `right` prop is missing from the proptypes.

ptbrowne: Done

ptbrowne:
> LGTM, but we only end up with one right area in the end? My latest understanding was that we'd have a render prop for the bottom right area as well.

I think the bottom right area could be done in a higher up implementation I am not sure that we should do that in this component as I have never seen this in the wild. It can be done quite easily (see the example). I prefer to have this component fairly "low-level" and flexible.

react/Media/styles.styl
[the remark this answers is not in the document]
ptbrowne: I fixed this but put the commit in another PR #1213

PR #1192, opened 2019-10-09
fix(InputGroup): Now accept a classname

Crash--:
Currently there is no way to pass a classname to our `InputGroup` component. 

I needed it to customize the border radius of this component 

https://crash--.github.io/cozy-ui/react/#inputgroup


Crash--:
I'm working for the futur! In our new design, we're trying to work with a more rounded component (as I understand it). I think that this will be soon our default guideline ? (cc @joel-costa )

Capture d’écran 2019-10-09 à 16 32 47

Issue #1040, opened 2019-06-13
MUI-based components need better testing


GoOz:
As seen in PR #1039, MUI-based components testing is quite useless for now. It doesn't test much.

y-lohse:
We now have argos testing on these.

PR #1200, opened 2019-10-15
Update styleguidist

drazik:
This is an attempt at upgrading to react-styleguidist v9 and use some features it introduce (like the possibility to use `import`s).

The main problem with using imports is that our automatic snapshot test method breaks.

https://drazik.github.io/cozy-ui/react/


ptbrowne:
Thanks for your awesome work drazik, I checked the error a bit and after console logging it, it was a bit clearer that we could fix the error by configuring the bublé compiler. I checked the default config that react-styleguidist gives to bublé and copy pasted it. Then I had a lot of "Toggle is already defined", "Icon is already defined" etc... After removing all the magic we were doing to automatically provide components to examples, error disappeared from the most part, seems like react-styleguidist is going the way of explicit better than explicit (see https://github.com/styleguidist/react-styleguidist/releases?after=v9.0.0-beta2) for examples to be more easily copy/pastable. Seems like they also listened to Martin :)

We are left with 3 failing snapshots and still a lot of red in the console. Forward we go !

drazik:
It seems to be working well, but we had to disable all MUI components snapshots as they are failing. But the current ones were totally useless (see https://github.com/cozy/cozy-ui/blob/master/react/__snapshots__/examples.spec.jsx.snap#L3388-L3406 and https://github.com/cozy/cozy-ui/issues/1040). Empty divs vs nothing at all is quite the same, so I'm in favor of merging as is and look at this specific issue in another PR. But for now we have an up-to-date styleguidist and we can improve our styleguide with the latest features.

react/examples.spec.jsx
```diff
@@ -36,15 +36,15 @@ testComponent('Infos')
...
-testComponent('MuiCozyTheme/Buttons')
-testComponent('MuiCozyTheme/ExpansionPanel')
-testComponent('MuiCozyTheme/List')
-testComponent('MuiCozyTheme/Menus')
-testComponent('MuiCozyTheme/Grid')
+// testComponent('MuiCozyTheme/Buttons')
+// testComponent('MuiCozyTheme/ExpansionPanel')
+// testComponent('MuiCozyTheme/List')
+// testComponent('MuiCozyTheme/Menus')
+// testComponent('MuiCozyTheme/Grid')
```
y-lohse: We might as well remove the lines entirely I think :)

PR #1201, opened 2019-10-16
feat: Add StackIcon component

Crash--:
The goal of this component is to superimpose two components : one in the background and the other in the foreground. 

I needed something like that for the `Qualification` component, since I needed to superimpose two icons (like I do in the sample).

~~At the beginning I started with an alternative API~~ :
```jsx
<ComboIcon
 foreground={{
   icon:'car',
  size:'16'
}}
background={{
  icon:'home',
  size:'32'
}}
/>
```
~~But I ended up with the following API, WDYT?~~
```jsx
<SuperImpose
  background={<Icon icon="file-duotone" color="blue"  size={32} />}
  foreground={<Icon icon="bank" color="red" height={16} width={16} />}
/>
```

~~https://crash--.github.io/cozy-ui/react/#superimpose~~

After discussion let's use the final API : 
```jsx
<StackIcon
  background={<Icon icon="file-duotone" color="blue"  size={32} />}
  foreground={<Icon icon="bank" color="red" height={16} width={16} />}
/>
```

https://crash--.github.io/cozy-ui/react/#stackicon


ptbrowne:
Good, reminds me of https://fontawesome.com/how-to-use/on-the-web/styling/stacking-icons. I am not a fan of the SuperImpose name though. What do you think of IconStack ?

Crash--:
> Looks cool. It makes me think about https://docs.cozy.io/cozy-ui/react/#!/CipherIcon/1. It doesn't seem to be possible to achieve this with the current implementation, but maybe we can do something to make it possible ?


We can do something like : 

```
<StackIcon
foreground={<Comp />}
background={<Comp2 />}
align="center|bottom-right"
/>
```

with center = superimpose current implementation and bottom-right CipherIcon implem? 

WDY?

ptbrowne:
The CipherIcon makes me think of the Badge https://crash--.github.io/cozy-ui/react/#badge

drazik:
> > Looks cool. It makes me think about [docs.cozy.io/cozy-ui/react/#!/CipherIcon/1](https://docs.cozy.io/cozy-ui/react/#!/CipherIcon/1). It doesn't seem to be possible to achieve this with the current implementation, but maybe we can do something to make it possible ?
> 
> We can do something like :
> 
> ```js-jsx
> <StackIcon
> foreground={<Comp />}
> background={<Comp2 />}
> align="center|bottom-right"
> />
> ```
> 
> with center = superimpose current implementation and bottom-right CipherIcon implem?
> 
> WDYT?

Yes it can be a solution. But it can be added later since we don't really know what usecase and API we should use to do that.

> The CipherIcon makes me think of the Badge [crash--.github.io/cozy-ui/react/#badge](https://crash--.github.io/cozy-ui/react/#badge). Introducting `align` in StackIcon would mean that we would have to implement all possible alignments (?) and I am not sure we want to do that.

That's also a potential solution, but the Badge is not designed to do that for now.

Didn't want to create a not existing problem here, I think it's OK as is. I was just thinking out loud.

PR #1202, opened 2019-10-16
feat(GridItem): Expose experimental GridItem

Crash--:
- Create a `Labs` folder when we can write our experimental components without fearing any breaking change. 
- Added a `GridItem` currently only used in the "Qualification module" but can be used pretty soon in the new Home. At this time, we will work on the API to be able to do what we really want. Currently it's not very well defined yet. 

Also, I didn't use any utilities classes in the component since if we do that, it's pretty hard to overwrite because u-* === * !important


react/Labs/Readme.md
```diff
@@ -0,0 +1,4 @@
+# Labs
+
+In this directory, components you will find are experimentals and
```
ptbrowne:
```suggestion
In this directory, components you will find are experimental and
```

PR #1203, opened 2019-10-16
feat: Expose react-styleguidist screenshot binary

ptbrowne:
This is helpful for other applications that want to take
screenshots of their styleguide. Banks has a styleguide for
example.


ptbrowne:
The simplest would be directly on the readme I guess ?

PR #1039, opened 2019-06-12
feat: Add List component based on MUI

GoOz:
Added the `List` component based on Mui's `List`.

With it I updated the style of `ListItem`, `ListItemIcon`, `ListItemSecondaryAction` and `ListSubheader`
Thought I didn't change [ListItemAvatar](https://v3-9-0.material-ui.com/api/list-item-avatar/) because I didn't need it and it's kinda buggy.

[Preview](https://gooz.github.io/cozy-ui/react/#!/List)

PR #1204, opened 2019-10-17
feat(Labs/ExperimentalModal): New Modal on mobile

Crash--:
This is a first experiment for our new kind of modale on mobile. 

We're currently designing a new kind of Modal for our mobile usage. Some variantes are to be defined but @y-lohse and @joel-costa are working on it. So let's put this component in Experimental for now, and soon we now exactly what we do, we can change the behavior of our `<Modal />`

You can test it here : https://crash--.github.io/cozy-ui/react/#experimentalmodal (remember, mobile ;))


react/Labs/ExperimentalModal/Readme.md
```diff
@@ -0,0 +1,33 @@
+### ExperimentalModal
+
+This modal brings a new UI only on Mobile for now. It changes the
+size of the Header and its margins and add border-bottom.
+It also changes, our ModalFooter to something fixed in the bottom
```
y-lohse:
```suggestion
It also changes our ModalFooter to something fixed in the bottom
```

react/Labs/ExperimentalModal/index.jsx
```diff
@@ -0,0 +1,75 @@
...
+  ModalButtons
+} from 'cozy-ui/react/Modal'
+import { Button, Icon } from 'cozy-ui/react'
+import withBreakpoints from 'cozy-ui/react/helpers/withBreakpoints'
+import classNames from 'classnames'
+import styles from './styles.styl'
+
+class ExperimentalModal extends Component {
+  render() {
+    console.log('this.props', this.props)
```
y-lohse: Can be removed now :)

react/Labs/ExperimentalModal/Readme.md
```diff
@@ -0,0 +1,33 @@
...
+const hideModal = () => setState({ modalOpened: false });
+
+<div>
+  <button onClick={()=>setState({ modalOpened: !state.modalOpened })}>
+    Toggle modal
+  </button>
+  {state.modalOpened && <ExperimentalModal
+    title='Ada Lovelace'
+    description={content.ada.short}
+    mobileFullscreen
```
ptbrowne: This prop is not very understandable IMHO, if we have a new modal, it could be time to find an alternative that is more understandable like isFullscreenOnMobile ?

ptbrowne: This prop is not used in the implementation. I think that in general, modals should be always fullscreen on mobile.

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -3307,6 +3307,12 @@ exports[`Label should render examples: Label 3`] = `
 </div>"
 `;
 
+exports[`Labs/ExperimentalModal should render examples: Labs/ExperimentalModal 1`] = `
+"<div>
+  <div><button>Toggle modal</button></div>
```
ptbrowne: Can the modal be opened for snapshots ?

react/Labs/ExperimentalModal/Readme.md
Crash--: We tend to do that (ie always be fullscreen on mobile) but I'm waiting for a full "GO" of @joel-costa on this point. But I agree with you ;)

Crash--: In fact it's a copy paste error on my own for the exemple. We already force mobileFullScreen on this component

react/__snapshots__/examples.spec.jsx.snap
Crash--: We need to figure out why all our Path/Components are empty div. 

See https://github.com/cozy/cozy-ui/pull/1039

Will take a look tomorrow

PR #1218, opened 2019-10-28
feat: Add QuotaAlert component

Crash--:
Since we now have the premium system to production, we try to display this modal every time we want to upload something but don't have enough space to our Cozy. We use it for Drive and now for photo and will also be used in Scanner. 

https://crash--.github.io/cozy-ui/react/#!/QuotaAlert


helpers/withInstance.jsx
```diff
@@ -0,0 +1,38 @@
+import { useState, useEffect } from 'react'
+
+const withInstance = client => {
```
ptbrowne: Since this is a hook, and the official guideline for hooks is to use names starting with `use`, I would rename this to `useInstance`.

helpers/withInstance.jsx
```diff
@@ -0,0 +1,38 @@
...
+        if (!(instanceFetched instanceof Error)) {
+          setInstance(instanceFetched)
+        }
+        if (!(contextFetched instanceof Error)) {
+          setContext(contextFetched)
+        }
+      } catch (e) {} //eslint-disable-line
+    }
+    fetchData()
+  }, [])
```
ptbrowne: I think we should have a long cache for context and instance, inside cozy-client.

react/QuotaAlert/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const buildPremiumLink = (uuid, managerUrl) =>
+  `${managerUrl}/cozy/instances/${uuid}/premium`
+
+const QuotaAlert = ({ t, onClose, client }) => {
+  let uuid, managerUrl
+  /**
+   * We do the request only on the web since Apple
+   * and Google have a retriscted policy for the
```
ptbrowne:
```suggestion
   * and Google have a restricted policy for the
```

helpers/withInstance.jsx
Crash--: Yep, I think we should have a long cache for those. https://github.com/cozy/cozy-client/issues/551

react/QuotaAlert/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const QuotaAlert = ({ t, onClose, client }) => {
+  let uuid, managerUrl
+  /**
+   * We do the request only on the web since Apple
+   * and Google have a restricted policy for the
+   * inApp purchase...
+   */
+  if (!isMobileApp()) {
+    const instanceInfo = useInstance(client)
```
drazik: I'm no hooks expert, but it looks like we are breaking this rule here: https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

Are we sure we don't introduce potential bugs?

react/QuotaAlert/index.jsx
```diff
@@ -46,6 +37,19 @@ const QuotaAlert = ({ t, onClose, client }) => {
       dismissAction={onClose}
     />
   )
-}
+})
+
+const QuotaAlert = withClient(({ client, onClose }) => {
+  /**
+   * We do the request only on the web since Apple
```
ptbrowne: Which request ? AMHA the comment here is not very clear.

Crash--: Yep, now that I moved the call it's not clear. fixed

react/QuotaAlert/index.jsx
drazik: Yep :+1:

ptbrowne: I think there is no bug here but it is reliant on the fact that isMobileApp does not change. If it were to change, we would potentially have a bug here.

PR #1226, opened 2019-11-05
feat: Add TextField from MUI

ptbrowne:
Re-export component and add theme overrides.

I deployed on but it does not seem to have been received by GitHub pages. https://ptbrowne.github.io/cozy-ui/react/#/TextField

### DIFF
--- a/react/MuiCozyTheme/theme.js
+++ b/react/MuiCozyTheme/theme.js
@@ -20,6 +20,9 @@ export const theme = createMuiTheme({
       dark: getCssVariableValue('scienceBlue'),
       contrastText: getCssVariableValue('primaryContrastTextColor')
     },
+    error: {
+      main: getCssVariableValue('errorColor')
+    },
     secondary: {
       light: getCssVariableValue('monza'),
       main: getCssVariableValue('portage'),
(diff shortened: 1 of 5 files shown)


Crash--:
your link is not working 🤔

react/MuiCozyTheme/theme.js
```diff
@@ -40,6 +43,22 @@ export const theme = createMuiTheme({
     borderRadius: defaultValues.borderRadius
   },
   overrides: {
+    MuiOutlinedInput: {
+      root: {
+        '&$disabled': {
+          background: 'var(--paleGrey)'
+        },
+        '&:hover $notchedOutline': {
+          borderColor: 'var(--coolGrey)' // Does not seem to work, IDK why :/
```
Crash--: should you add a todo to fix it if needed? Or should we remove this config if it's not working and we don't want it?

ptbrowne: I removed this code from the PR and I will open another one with this code, I may need a pairing on this.

ptbrowne:
> your link is not working 🤔

Yes, GitHub does not seem to be picking up the changes that I deployed on the `gh-page` branch :/

The only recourse I see here is to `yarn watch:doc:react` locally. The argos build also shows the result: https://www.argos-ci.com/cozy/cozy-ui/builds/57082

PR #1228, opened 2019-11-05
fix: Change hover style of TextField

ptbrowne:
Previously, the border of the TextField component was black on hover instead of being in cool grey.


nono:
Ping @ptbrowne

ptbrowne:
I have to take a look at this after upgrade to MUI v4.

ptbrowne:
It works now 🎉 , thanks for the ping @nono, cc @JF-Cozy

PR #1232, opened 2019-11-06
feat: Add PercentageBar component

drazik:
It is based on PercentageLine, without modifying it.

See https://drazik.github.io/cozy-ui/react/#!/PercentageBar


react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -3425,6 +3425,13 @@ exports[`Menu should render examples: Menu 3`] = `
 </div>"
 `;
 
+exports[`PercentageBar should render examples: PercentageBar 1`] = `"<div></div><pre class=\\"rsg--root-0\\">Error: Cannot find module 'cozy-ui/transpiled/react/PercentageBar' from 'examples.spec.jsx'</pre>"`;
```
Crash--: hmmm. Can you fix this?

drazik: Yep, done. Thanks.

PR #1236, opened 2019-11-08
Handle flat and nested subdomains

y-lohse:
So far when using AppLinker, we only supported the flat subdomains configuration. We can now support both, by using this PR and https://github.com/cozy/cozy-stack/pull/2172

### DIFF
--- a/react/AppLinker/native.js
+++ b/react/AppLinker/native.js
@@ -12,13 +12,32 @@ const ensureFirstSlash = path => {
   }
 }
 
-export const generateWebLink = ({ cozyUrl, nativePath, slug }) => {
+/**
+ * generateWebLink - Construct a link to a web app
+ *
+ * @param {object} Options               Object of options
+ * @param {string}   options.cozyUrl       Base URL of the cozy, eg. cozy.tools or test.mycozy.cloud
+ * @param {string}   options.nativePath    Path inside the app, eg. /files/test.jpg
+ * @param {string}   options.slug          Slug of the app
+ * @param {string}   options.subDomainType Whether the cozy is using flat or nested subdomains. Defaults to flat.
+ *
+ * @return {string} Generated URL
+ */
+export const generateWebLink = ({
+  cozyUrl,
+  nativePath,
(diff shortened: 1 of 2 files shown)


react/AppLinker/native.js
```diff
@@ -12,28 +12,48 @@ const ensureFirstSlash = path => {
...
+/**
+ * generateWebLink - Conbstruct a link to a web app
+ *
+ * @param {object} Options               Object of options
+ * @param {string}   options.cozyUrl       Base URL of the cozy, eg. cozy.tools or test.mycozy.cloud
+ * @param {string}   options.nativePath    Path inside the app, eg. /files/test.jpg
+ * @param {string}   options.slug          Slug of the app
+ * @param {string}   options.subDomainType Whether the cozy is using flat or nested subdomains
+ *
+ * @return {type} Description
```
Crash--:
```suggestion
 * @return {String} Url
```

react/AppLinker/native.js
```diff
@@ -12,28 +12,48 @@ const ensureFirstSlash = path => {
...
 
-export const generateWebLink = ({ cozyUrl, nativePath, slug }) => {
+/**
+ * generateWebLink - Conbstruct a link to a web app
+ *
+ * @param {object} Options               Object of options
+ * @param {string}   options.cozyUrl       Base URL of the cozy, eg. cozy.tools or test.mycozy.cloud
+ * @param {string}   options.nativePath    Path inside the app, eg. /files/test.jpg
+ * @param {string}   options.slug          Slug of the app
+ * @param {string}   options.subDomainType Whether the cozy is using flat or nested subdomains
```
Crash--: We should add somewhere that `flag` is the default behavior

y-lohse:
> I'm wondering if generateWebLink should not read by itself cozy-data-domain to not having to pass this information at each call in our app

I think it's not the responsibility of this function to do that, just as it doesn't read the cozy URL or other information. These infos can be found in various ways and I think the caller is best suited to decide when and how to read them.

PR #1238, opened 2019-11-12
fix: Flex basis for tab names

y-lohse:
I'm still not sure *why*, but when specifying `flex 0 0`, the elements flex-basis is set to 0% instead of the default `auto`. This caused a problem when adding icons to tab names for example:

Capture d'écran 2019-11-12 14 10 45

Specifying `flex 0 0 auto` solves the issue.


ptbrowne:
TIL.

https://developer.mozilla.org/fr/docs/Web/CSS/flex

```css
/* Deux valeurs : flex-grow | flex-shrink */
/* flex-basis vaut alors 0 */
flex: 2 2;
```

PR #1239, opened 2019-11-12
fix: Cannot use var referencing another var

ptbrowne:
We have to use pomegranate right now instead of an indirection
since we cannot pass to mui a css variable.

### DIFF
--- a/stylus/settings/palette.styl
+++ b/stylus/settings/palette.styl
@@ -155,6 +155,4 @@
     --primaryColorLighter #4B93F7
     --primaryColorLightest #9FC4FB
     --primaryContrastTextColor var(--white)
-
-    --errorColor var(--pomegranate)
 // @stylint on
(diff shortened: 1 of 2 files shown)


edas:
Si la CSS passe avant, on ne peut pas récupérer les variables dans le DOM ?

ptbrowne:
Si on pourrait sûrement, dans getCSSVariableValue, continuer à "déréférencer" les variables CSS jusqu'à avoir une véritable valeur, mais là c'est plutôt la suppression d'un bug que j'ai introduit. Je peux créer une issue pour faire ça.

stylus/settings/palette.styl
```diff
@@ -155,6 +155,4 @@
     --primaryColorLighter #4B93F7
     --primaryColorLightest #9FC4FB
     --primaryContrastTextColor var(--white)
```
y-lohse: Don't we have the same problem here ? Although I guess white ends up being white anyway…

ptbrowne: I think I was a bit imprecise before. CSS vars do not work for MUI when MUI has to make computations on it (making a color lighter or darker, to build all the variants of the color). I think we do not have the problem here because `primaryContrastTextColor` is not used as a base color for other colors, contrary to previously used `error.main` where MUI has to compute all shades of the `error` color.

PR #1240, opened 2019-11-12
fix: Fix grid gap value

y-lohse:
Stylus compiles `grid-gap rem(1)` to… `grid-gap rem(1)`, which the browser obviously doesn't understand.


ptbrowne:
Good catch, it does not work since the import of the mixin is missing.

ptbrowne:
This advocates for a static checker on the transpiled CSS.

PR #1243, opened 2019-11-19
feat: AppIcon doesn't require secure and domain props anymore

y-lohse:
This change comes from the [Home app](https://github.com/cozy/cozy-home/blob/master/src/components/AppIcon.jsx) where I made a wrapper that avoided needing `domain` and `secure` props all the time, since they can be detected by cozy-client.

It's a breaking change because if `AppIcon` is not inside the cozy-client context, it won't work anymore. That said other components in Ui already rely on this behavior so it's not a big stretch.

### DIFF
--- a/react/AppIcon/Readme.md
+++ b/react/AppIcon/Readme.md
@@ -53,9 +53,6 @@ initialState = { fetchIcon: fetchIcon1 };
 </div>
 ```
 
-## `domain` and `secure` props
-If the `fetchIcon` is missing, the `<AppIcon />` component needs the `domain` prop (litteraly the cozy domain) and the secure prop, to know which protocol use between `http` or `https`.
-
 ### Provide `fallbackIcon`
 
 You can provide an `<Icon />` `icon` props to fallback when the AppIcon fetched is broken or the `fetchIcon` function errored.
--- a/react/AppIcon/index.jsx
+++ b/react/AppIcon/index.jsx
@@ -1,11 +1,11 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
+import { withClient } from 'cozy-client'
 import styles from './styles.styl'
 
-import Icon from '../Icon'
+import Icon, { iconPropType } from '../Icon'
 import palette from '../palette'
-import { iconPropType } from '../Icon'
 
 import { getPreloaded, preload } from './Preloader'
 import { AppDoctype } from '../proptypes'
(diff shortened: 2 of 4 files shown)


react/AppIcon/Readme.md
```diff
@@ -33,46 +34,43 @@ Is it also possible to provide a custom asynchronous `fetchIcon` which takes an
...
 const fetchIconBroken = () => 'blahblahblah'
 
-const brokeFetchIcon = () => setState({ fetchIcon: fetchIconBroken  })
+const brokeFetchIcon = () => setState({ fetchIcon: fetchIconBroken })
 
-initialState = { fetchIcon };
-
-<div>
+initialState = { fetchIcon }
+;<div>
```
ptbrowne: Why the colon changes ?

react/AppIcon/index.jsx
```diff
@@ -17,15 +17,27 @@ const FETCHING = 'fetching'
...
-      status: preloaded ? DONE : FETCHING
-    }
     this.isUnmounting = false
     this.handleError = this.handleError.bind(this)
+
+    try {
+      const { app, client } = props
+      const cozyURL = new URL(client.getStackClient().uri)
+      this.domain = cozyURL.host
+      this.secure = cozyURL.protocol === 'https:'
```
ptbrowne: AMHA the component should not store this.domain and this.secure.

1. It makes the constructor a bit bloated
2. It makes the component deal with low level stuff when it is pretty high level

Passing the client directly to loadFn would be cleaner amha.

If the backward compatibility is blocking here I think a wrapper to transform getPreloaded and loadFn so that they accept a client would be better as it would at some point completely disappear.

```js
const passDomainSecure = fnTakingDomainAndSecure => {
  if (fnTakingDomainAndSecure.length === 2) {
    return fnTakingDomainAndSecure // Already accepting the app + client
  }
  return function (app, client) => {
      const cozyURL = new URL(client.getStackClient().uri)
      const domain = cozyURL.host
      const secure = cozyURL.protocol === 'https:'
      return fnTakingDomainAndSecure.apply(this, [app, domain, secure])
  }
}

const loadFn = passDomainSecure(this.props.loadFn)
const result = await loadFn(this.props.client)

```

react/AppIcon/Readme.md
```diff
@@ -33,46 +34,43 @@ Is it also possible to provide a custom asynchronous `fetchIcon` which takes an
...
-initialState = { fetchIcon: fetchIcon1 };
-
-<div>
-  <button onClick={handleSwitch}>Switch</button>{' '}
-  fetchIcon: {state.fetchIcon.name}<br/>
+initialState = { fetchIcon: fetchIcon1 }
+;<div>
+  <button onClick={handleSwitch}>Switch</button> fetchIcon:{' '}
+  {state.fetchIcon.name}
+  <br />
```
ptbrowne: It seems the JSX code here went into a linter, it would be easier to see the real changs of the PR if these changes were in another commit.

y-lohse: Hmm yeah there are some actual changes that i want in there but I didn't realize the linter went overboard. I'll try to re-edit that file.

react/AppIcon/index.jsx
y-lohse: Yeah good point, i'll add a TODO for next time.

PR #1244, opened 2019-11-20
feat: New variation for experimental modal

y-lohse:
As the titles says, this is an alternative layout for the new modals. [Have a look at the documentation and let me know what you think.](https://yannick-lohse.fr/cozy-ui/react/#!/ExperimentalModal)

Eventually I think it would be good to rename `description` and `descriptionFooter` to maybe `content` and `contentFooter`, when I see `description` I tend to think of a simple string but it can be a lot more here.

### DIFF
--- a/react/Labs/ExperimentalModal/Readme.md
+++ b/react/Labs/ExperimentalModal/Readme.md
@@ -1,11 +1,8 @@
 ### ExperimentalModal
 
-This modal brings a new UI only on Mobile for now. It changes the
-size of the Header and its margins and add border-bottom.
-It also changes our ModalFooter to something fixed in the bottom
-and add a border-top.
+There are 2 variations of this modal that can be used.
 
-Not tested on "long content".
+The first one has a fixed header and footer that are separated from the body of the modal by a border.
 
 ```
 import ExperimentalModal from 'cozy-ui/transpiled/react/Labs/ExperimentalModal';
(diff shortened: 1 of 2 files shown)


ptbrowne:
First glance:

- The cross icon is still broken on the styleguide, I think it would be the time to fix it
- I cannot close the second modal after opening it
- In the first modal, I cannot see the borders that should separate the body from the header/footer

y-lohse:
> descriptionFooter: Don't you think we should use the actions for that ? I find descriptionFooter not to be a very good name. When I see the first modal on mobile, it has its actions buttons at the bottom. That's what we want right ? Only difference seems to be the box-shadow.

Mmh no, for example in https://github.com/cozy/cozy-libs/pull/871#issue-343306252 — it's not just the buttons that need to stick to the bottom of the modal, it's also the little gray text. Additionally, one of these buttons is a `ButtonLink` that is wrapped in a `AppLinker`, which isn't possible to do by simply passing the `primaryAction` etc props.

That being said I wouldn't mind unifying both areas when this component comes out of the experimental phase. i think a `StickyFooter` and a `FixedFooter` or something might be good, but then dev needs be able to control what goes in them.

____

Meantime, looking into the styleguidist issues.

y-lohse:
> The cross icon is still broken on the styleguide, I think it would be the time to fix it

Separate PR incoming since it's not related to this issue.

> I cannot close the second modal after opening it

Fixed, should be live soon.

> In the first modal, I cannot see the borders that should separate the body from the header/footer

Are you sure? Note that they only appear on mobile.

ptbrowne:
> Mmh no, for example in cozy/cozy-libs#871 (comment) — it's not just the buttons that need to stick to the bottom of the modal, it's also the little gray text. Additionally, one of these buttons is a ButtonLink that is wrapped in a AppLinker, which isn't possible to do by simply passing the primaryAction etc props.

Seems like the action props are not flexible enough. Maybe we could pass an `actions` property that could be used for that ?

drazik:
I think we should once again take material-ui as inspiration. See [this example](https://material-ui.com/components/dialogs/#AlertDialog.js). They have a `DialogActions` in which we can render anything we want. This is more flexible than our `primaryAction` and `secondaryAction`.

y-lohse:
Replaced by #1257

PR #1245, opened 2019-11-20
feat: Add HistoryRow component

Crash--:
Add a new `HistoryRow` component. 

Styleguidist should speak for itself ;) 

https://crash--.github.io/cozy-ui/react/


react/HistoryRow/styles.styl
```diff
@@ -0,0 +1,37 @@
+@require 'utilities/spaces.styl'
+    
+.HistoryRowCircleWrapper 
+    width: 2.6rem;
+    height: 2.6rem;
+    justify-content: center;
+    display: flex;
+    align-self: start;
```
ptbrowne: Nit: we normally do not use ";" and ":" in stylus files

react/HistoryRow/styles.styl
```diff
@@ -0,0 +1,37 @@
...
+
+.HistoryRowMedia
+    position relative
+.HistoryRowMedia::before
+    content ''
+    border-left 1px var(--coolGrey) dashed
+    position absolute
+    margin-left 20px
+    top: 35px
+    bottom: -15px
```
ptbrowne: Nit: These values seem magical, could you name them to make their purpose more explicit ?

react/HistoryRow/styles.styl
```diff
@@ -0,0 +1,40 @@
+@require 'utilities/spaces.styl'
+    
+.HistoryRowCircleWrapper 
+    width 2.6rem
+    height 2.6rem
+    justify-content center
+    display flex
+    align-self start
+
+.HistoryRowCirle
```
ptbrowne:
```suggestion
.HistoryRowCircle
```

react/HistoryRow/styles.styl
```diff
@@ -0,0 +1,40 @@
...
+    position relative
+.HistoryRowMedia::before
+    content ''
+    border-left 1px var(--coolGrey) dashed
+    position absolute
+    margin-left 20px
+    //circle is 1rem + 1px border => 18px
+    // 18 + 16px padding => 34
+    top 34px
+    //We have 16px from padding so the next item will be 16px lower
```
ptbrowne: from which padding ? I don't see a vertical padding in this file ?

I think we should restrain ourselves from commenting when we can have explicit variable names. Comments get easily old and out of date. Moreover variables lead to easier reuse.

```
CIRCLE_HEIGHT=1.125rem // If using this value in .HistoryRowCircle, its box-sizing should be set to border-box so that the border width is accounted in the height
VERTICAL_PADDING_FROM_??=1rem

top: CIRCLE_HEIGHT + VERTICAL_PADDING_FROM_??
```

I think that since the rem are converted implicitly into pixels by multiplying by 16, there will be a bug if the user changes its root font-size (from 16 to another value).

Crash--: You right. Let's see if this one is better ;)

Crash--:
>I just wonder if we could have separated the Timeline functionality (circles, dashed line) from its specific usage for files (showing version, download link etc..). 

At the beginning this is almost what I wanted: having a timeline component managing the vertical dashed line and letting the "HistoryRow" dealing with the circles and file version. 

At the end I only used one component because having a Timeline component will be much more work that only displaying a vertical line. So I told myself to wait before having an other case of timeline to start thinking about a good API

PR #1247, opened 2019-11-20
fix: Use a regular button for the modal cross

y-lohse:
We've had an issue for a while with the crosses in the modals of the styleguidist, more details can be found on [this thread](https://github.com/cozy/cozy-ui/issues/335#issuecomment-487582163) but basically it was a conflict of css classes.

With this fix, the cross doesn't use the `Button` component anymore, which means they don't try to override each others styles anymore.

My online preview repo is currently used by #1244 so I'll update this PR once the other one is merged.

### DIFF
--- a/react/Modal/ModalCross.jsx
+++ b/react/Modal/ModalCross.jsx
@@ -1,19 +1,15 @@
 import React from 'react'
 import styles from './styles.styl'
 import cx from 'classnames'
-import { Button } from '../Button'
 import Icon from '../Icon'
 import palette from '../palette'
 
 const ModalCross = ({ onClick, color, className }) => (
-  <Button
-    theme="close"
+  <button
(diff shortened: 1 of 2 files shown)


react/Modal/ModalCross.jsx
```diff
@@ -1,27 +1,22 @@
...
 const ModalCross = ({ onClick, color, className }) => (
-  <Button
-    theme="close"
+  <button
     className={cx(styles['c-modal-close'], className)}
     onClick={onClick}
-    extension="narrow"
     type="button"
-    label="Close"
-    iconOnly
```
Crash--: We lose accessibility here? Screen reader will not be able to tell the close button, right?

y-lohse: Ah right, forgot to add that back in.

PR #1249, opened 2019-11-20
feat: Implement ViewStack

ptbrowne:
Adds ViewStack that helps building UI with slides that can control an
animated stack of slides.

- [x] Test

View here: https://ptbrowne.github.io/cozy-ui/react/#!/ViewStack

### DIFF
--- a/react/ViewStack/index.jsx
+++ b/react/ViewStack/index.jsx
@@ -1,0 +1,66 @@
+import React, { useState, useContext } from 'react'
+import SwipeableViews from 'react-swipeable-views'
+
+// This is guessed, I could not see how to interface properly with
+// react-swipeable-views's onTransitionEnd
+const STEPPER_TRANSITION_DURATION_MS = 500
+
+export const ViewStackContext = React.createContext()
+export const useViewStack = () => useContext(ViewStackContext)
+
+const sleep = duration => new Promise(resolve => setTimeout(resolve, duration))
+
+const useStack = initialState => {
+  const [arr, setArray] = useState(initialState)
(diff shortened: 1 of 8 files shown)


react/ViewStack/index.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const [curIndex, setCurIndex] = useState(initialState.length - 1)
+
+  const push = item => {
+    const newArr = [...arr, item]
+    setArray(newArr)
+    setCurIndex(curIndex + 1)
+  }
+
+  const pop = async () => {
+    if (arr.length < 2) {
```
drazik: Nit: I find `if (arr.length <= 1)` more descriptive in that case. At first I thought « why 2? ». It may be just me :sweat_smile:

react/index.js
```diff
@@ -83,3 +83,8 @@ export { default as Popup } from './Popup'
 export { default as IconStack } from './IconStack'
 export { default as QuotaAlert } from './QuotaAlert'
 export { default as ThresholdBar } from './ThresholdBar'
+export {
+  default as ViewStack,
+  useViewStack,
+  ViewStackContext
```
drazik: We export both `useViewStack` and `ViewStackContext` to give the possibility to the user to use the hook or the context consumer? If that's it, shouldn't we expose only the hook so there is only one API?

ptbrowne: The hook is only usable in functional components, this is why I exported `ViewStackContext`.

react/ViewStack/index.jsx
ptbrowne: Done, I also thought it was weird.

PR #1250, opened 2019-11-20
feat: DateMonthPicker

ptbrowne:
Show a date month picker.

- [x] Test

image

⚠️ This is a very crude picker. We wanted to use [Material UI Datepicker](https://material-ui-pickers.dev/demo/datepicker), but this is only compatible with date-fns2 or moment. I propose that when we migrate, we will change this component to use Material UI date pickers.

### DIFF
--- a/react/DateMonthPicker/index.jsx
+++ b/react/DateMonthPicker/index.jsx
@@ -1,0 +1,111 @@
+import React, { useState } from 'react'
+import PropTypes from 'prop-types'
+import Button from '../Button'
+import Icon from '../Icon'
+import { translate } from '../I18n'
+import range from 'lodash/range'
+import { format } from 'date-fns'
+import styles from './styles.styl'
+import cx from 'classnames'
+
+const MonthButton = translate()(({ monthNum, f, onClick, isSelected }) => {
+  const d = new Date(2019, monthNum, 15)
+  const handleClick = () => {
+    onClick(monthNum)
+  }
+  return (
+    <Button
+      theme="secondary"
+      className={cx(
+        styles.DateMonthPicker__MonthButton,
+        isSelected ? styles.ValueSelected : null
+      )}
+      onClick={handleClick}
+      label={f(d, 'MMM')}
+    />
+  )
+})
+
+const useCounter = (initialValue, min, max) => {
+  const [state, setState] = useState(initialValue)
+  const increment = () => setState(Math.min(state + 1, max))
+  const decrement = () => setState(Math.max(state - 1, min))
+  return [state, decrement, increment]
+}
+
+const DateMonthPicker = ({ initialValue, onSelect }) => {
+  const [initialYear, initialMonth] = initialValue
+    .split('-')
+    .map(x => parseInt(x, 10))
+  const [year, decreaseYear, increaseYear] = useCounter(
+    parseInt(initialYear, 10),
+    1990,
+    Infinity
+  )
+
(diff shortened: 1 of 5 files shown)


Crash--:
I think we do not need ViewStack for this PR?

react/DateMonthPicker/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+  const [state, setState] = useState(initialValue)
+  const increment = () => setState(Math.min(state + 1, max))
+  const decrement = () => setState(Math.max(state - 1, min))
+  return [state, decrement, increment]
+}
+
+const DateMonthPicker = ({ initialValue, onSelect }) => {
+  const [initialYear, initialMonth] = initialValue
+    .split('-')
+    .map(x => parseInt(x, 10))
```
Crash--: We should add a check and warn the user if the initialValue doesn't match want we want.

ptbrowne: A PropType check right ?

Crash--: It can be a PropTypes check yes

ptbrowne: I added a Proptypes check

ptbrowne:
> I think we do not need ViewStack for this PR?

I rebased on master :)

react/DateMonthPicker/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+    return new Error(
+      'Invalid prop `' +
+        propName +
+        '` supplied to' +
+        ' `' +
+        componentName +
+        '`. Should be in the form YYYY-MM.'
+    )
+  }
+}
```
Crash--: Top 👏

react/DateMonthPicker/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+import Button from '../Button'
+import Icon from '../Icon'
+import { translate } from '../I18n'
+import range from 'lodash/range'
+import { format } from 'date-fns'
+import styles from './styles.styl'
+import cx from 'classnames'
+
+const MonthButton = translate()(({ monthNum, f, onClick, isSelected }) => {
+  const d = new Date(2019, monthNum, 15)
```
drazik: Are year and day hard coded here because we don't care about it in this component? Maybe we should add a comment?

react/DateMonthPicker/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  return (
+    <>
+      <div className={styles.DateMonthPicker__YearControls}>
+        <Button
+          className={styles.DateMonthPicker__YearButton}
+          theme="secondary"
+          size="small"
+          label=""
```
drazik: I think we should set a label and add the `iconOnly` prop if you just want to show an icon.

react/DateMonthPicker/index.jsx
```diff
@@ -0,0 +1,111 @@
...
+            year === initialYear && styles.ValueSelected
+          )}
+        >
+          {year}
+        </div>
+        <Button
+          className={styles.DateMonthPicker__YearButton}
+          theme="secondary"
+          size="small"
+          label=""
```
drazik: Same here with label and iconOnly

react/DateMonthPicker/index.jsx
ptbrowne: Done, thanks

PR #1251, opened 2019-11-21
feat: Use laptop Icon from UI icon

Crash--:
Since we have this icon in our Icon's palette, let's use it. 

Also removed the now unused icon

### DIFF
--- a/react/PushClientButton/index.jsx
+++ b/react/PushClientButton/index.jsx
@@ -3,8 +3,7 @@ import styles from './styles.styl'
 import React from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
-
-import DeviceIcon from 'cozy-ui/assets/icons/illus/device-laptop.svg'
+import { Icon } from 'cozy-ui/transpiled/react'
 
 const ButtonClient = props => {
   const { label, href, onClick, className } = props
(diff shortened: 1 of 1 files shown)


y-lohse:
But the illustration and the icon aren't the same though? And the new icon doesn't have a color, so it will be black?

Crash--:
Oh right I shouldn't have deleted this device-laptop.svg file since... I used this one.

react/PushClientButton/index.jsx
```diff
@@ -17,7 +16,7 @@ const ButtonClient = props => {
       onClick={onClick}
     >
       <figure>
-        <img width="32" height="32" src={DeviceIcon} />
+        <Icon icon="device-laptop" size="32" />
```
y-lohse: Soooo it should be `devices-laptop`, right?

Crash--: Nop `device-laptop` is fine ;) https://docs.cozy.io/cozy-ui/react/#!/Icon/1

drazik: I think it still works because the sprite was not re-generated, but it shouldn't, no?

Crash--: Oh come on! I didn't see my mistake in the filename :o. Thanks :)

ptbrowne:
Ping @Crash-- , is there still work to do ?

PR #1256, opened 2019-11-21
feat: Better styles for DateMonthPicker

ptbrowne:
Thanks @joel-costa !

Visible here : https://ptbrowne.github.io/cozy-ui/react/#/DateMonthPicker


stylus/utilities/shadows.styl
```diff
@@ -0,0 +1,4 @@
+box-shadow-12=0 4px 12px 0 rgba(0, 0, 0, .08)
+
+$raised-12
+    box-shadow box-shadow-12
```
y-lohse: I think it would be best to keep the utilities folder for files that output utility classes. There is a `settings` folder that is perfect for this sort of declaration :). Same thing for the radius declaration!

y-lohse:
Looks pretty good overall except for the utility variables. The `active` state could also use some love, but not blocking for me:

Capture d'écran 2019-11-22 10 37 24

ptbrowne:
> The active state could also use some love

You mean removing the `outline` right ?

y-lohse:
Yeah and ideally replacing it with something else so that there is still a visual indication that the element is active. Maybe a box-shadow or a border?

ptbrowne:
> replacing it with something else so that there is still a visual indication that the element is active. 

The background is already darker on click.

stylus/utilities/shadows.styl
ptbrowne: You're right, I changed it :)

PR #1257, opened 2019-11-21
feat: Material UI Dialog

y-lohse:
Allows us to use the `Dialog` component from Material UI. See the demo here: https://yannick-lohse.fr/cozy-ui/react/#!/Dialog

Still testing it out on a real use case so it's not 100% feature complete and the doc will need improvements, but this should be a good first step. I will remove the `ExperimentalModal` in this PR since everything it does can be accomplished with `Dialog`.

I'll request reviews a bit later, but feel free to have a look since there shouldn't be any major changes anymore.


y-lohse:
I've made a few tweaks and updated the demo. I'm now using the Dialog in Harvest, and it works!

Capture d'écran 2019-11-22 10 15 28

I expect there to be some more work, especially adjusting the margins between elements, but it will be for another PR.

Crash--:
Do we have a border below the title if there is a title? 

Buttons are on two lines , is it wanted and the default behavior ? ExpModal has the buttons only on one line.

y-lohse:
> Do we have a border below the title if there is a title?

Not by default, but it's as easy as

```
<Dialog>
  <DialogTitle />
  <Divider />
  <DialogContent />
</Dialog>
```

`Divider` is another MUI component, it's linked in the docs.

> Buttons are on two lines , is it wanted and the default behavior ? ExpModal has the buttons only on one line.

The 2 lines is not the default behavior, the default behavior is what you can see in the styleguidist. Putting them on 2 separate lines is a specific thing in this modal.

ptbrowne:
> Putting them on 2 separate lines is a specific thing in this modal.

Could we have this variant in the styleguide ?

y-lohse:
> Could we have this variant in the styleguide ?

I'd rather wait until I have discussed it further with @joel-costa before adding that to the styleguide. i don't think we're quite clear yet about the different button layouts and when to use them, desktop vs mobile, etc.

ptbrowne:
One thing to be careful about is the presence of the `has-body` class on document.body since it changes the z-index of alerts on mobile (I guess for alerts to appear from below the navigation bar when there are no modals, but on top of everything when a modal is mounted, thus hiding the navigation bar).

react/MuiCozyTheme/theme.js
```diff
@@ -42,183 +42,220 @@ export const theme = createMuiTheme({
...
     },
-    MuiListItemSecondaryAction: {
-      root: {
-        zIndex: 1
+    gutters: {
+      paddingLeft: 0,
+      paddingRight: 0
+    },
+    sticky: {
+      backgroundColor: 'var(--zircon)'
```
Crash--: So now we can just do `var(--zircon)` instead of `getCssVariableValue('zircon')` ?

y-lohse: It depends, see https://github.com/cozy/cozy-ui/pull/1239#discussion_r345648843 for more. Generally though it would probably be safer to always use `getCssVariableValue` going into the future.

PR #1258, opened 2019-11-21
ModalStack

ptbrowne:
Using the same mechanism as the ViewStack (that should be renamed SlideStack
now ?) we can have a stack of views that uses Modals.

This could be useful for example to change the style of stacking based on
whether we are on desktop or mobile.

See how the switch from a sliding stack to a modalling (?) stack : https://github.com/cozy/cozy-banks/pull/1680/commits/3eedbf390f8adb5f4ddba2ad255c9e4e392e7f27


ptbrowne:
Cc @y-lohse

y-lohse:
Looks pretty good to me, but could you add a little readme with some usage examples?

PR #1267, opened 2019-11-25
Feat/tracker

Crash--:
The idea is to have a method to do the check and just expose a `trackEvent([])` for the app. No need to check if there is a tracker anymore. 

https://github.com/cozy/cozy-drive/pull/1908

### DIFF
--- a/react/helpers/tracker.spec.js
+++ b/react/helpers/tracker.spec.js
@@ -1,0 +1,29 @@
+import * as trackerFile from './tracker'
+
+jest.mock('./tracker', () => ({
+  ...require.requireActual('./tracker'),
+  getTracker: jest.fn()
+}))
+describe('tracker / piwik action', () => {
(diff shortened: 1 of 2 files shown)


react/helpers/tracker.spec.js
```diff
@@ -0,0 +1,36 @@
...
+    return event
+  }
+}
+
+describe('tracker / piwik action', () => {
+  let mocked = new Tracker()
+  let getTrackerSpy, pushSpy
+  beforeEach(() => {
+    getTrackerSpy = jest.spyOn(trackerFile, 'getTracker')
+    getTrackerSpy.mockImplementation(() => mocked)
```
ptbrowne: IMHO, it'd maybe be simpler to remove the `Tracker` class and to do 

`getTrackerSpy.mockImplementation(() => ({ push: jest.fn() }))` ?

PR #1268, opened 2019-11-26
fix: Faulty babel configuration

y-lohse:
The babel configuration required a file in `docs`, but the `docs` folder is not published on npm, leading to build errors. I moved the file to the scripts folder where we already had other build related files.


ptbrowne:
Why do we need the babel config after the package has been published ?

y-lohse:
> Why do we need the babel config after the package has been published ?

That's a good point, we… shouldn't? But for some reason it is loaded, at the very least when running test on the home app. i'll create an issue to investigate further.

PR #1273, opened 2019-11-28
Add ContactsList and ContactsListModal

drazik:
For the link between bank accounts and contacts feature in banks, we needed to extract the contacts list from cozy-contacts. `ContactsList` is this new component. Strictly speaking, this is not 100% the same as cozy-contacts' list since the multi-selection feature via checkboxes is missing. We don't need it in banks' feature and it's implemented using redux in cozy-contacts. I didn't want to add this dependency. I plan to add this feature in near future so we can drop a `ContactsList` in place of the current list in cozy-contacts.

In banks, we also need a modal that allows the user to type something in an input to filter a contacts list and select on of them. The `ContactsListModal` does this.

See https://drazik.github.io/cozy-ui/react/#!/Contacts


react/ContactsList/ContactRow.spec.js
```diff
@@ -0,0 +1,73 @@
...
+
+import ContactRow from './ContactRow'
+import renderer from 'react-test-renderer'
+describe('ContactRow', () => {
+  test('should accept the strict minimum', () => {
+    const contact = { email: [{ address: 'johndoe@localhost' }] }
+    const contactRowInstance = <ContactRow contact={contact} />
+    const contactrow = mount(contactRowInstance)
+    const contactrowemail = contactrow.find('ContactEmail')
+    expect(contactrowemail).toBeDefined()
```
ptbrowne: I think it is unnecessary to use toBeDefined() as text() will fail if called on 0 or more than 1 node.

react/ContactsList/ContactsList.jsx
```diff
@@ -0,0 +1,50 @@
...
+  const { contacts, className, onItemClick, ...rest } = props
+
+  const sortedContacts = [...contacts].sort(sortLastNameFirst)
+  const categorizedContacts = sortedContacts.reduce((acc, contact) => {
+    const name = buildLastNameFirst(contact)
+    const header = name[0] || 'EMPTY'
+    acc[header] = acc[header] || []
+    acc[header].push(contact)
+    return acc
+  }, {})
```
ptbrowne: I think we should avoid sorting and building sorted contacts inside the render as it can be fairly expensive (we do not want to explode our 16ms budget ;))

Doing this via reselect would be better IMHO.

react/ContactsList/ContactsList.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <ol className={cx(styles['list-contact'], className)} {...rest}>
+      {Object.keys(categorizedContacts).map(header => (
```
ptbrowne: Nit: It is not guaranteed here than the keys will be returned in the order of insertion. I think sorting the keys would be better to ensure that we have no surprise.

react/ContactsListModal/DemoProvider.jsx
```diff
@@ -0,0 +1,33 @@
...
+  getChildContext() {
+    return {
+      client: mockClient
+    }
+  }
+
+  render() {
+    return <>{this.props.children}</>
+  }
+}
```
ptbrowne: It seems to achieve the same thing as createClientWithData, is there any difference ?

react/ContactsListModal/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+import Spinner from '../Spinner'
+import styles from './styles.styl'
+import Input from '../Input'
+import PropTypes from 'prop-types'
+import withBreakpoints from '../helpers/withBreakpoints'
+
+const thirtySeconds = 30000
+const olderThan30s = fetchPolicies.olderThan(thirtySeconds)
+
+const mkFilter = filter => contacts => {
```
ptbrowne: Nit: filter could be renamed filterStr ? I thought filter was a function at first.

react/ContactsList/ContactPropTypes.js
```diff
@@ -0,0 +1,59 @@
...
+    primary: PropTypes.bool
+  }),
+  birthday: PropTypes.string,
+  note: PropTypes.string,
+  metadata: PropTypes.shape({
+    version: PropTypes.number,
+    cozy: PropTypes.bool,
+    me: PropTypes.bool
+  })
+}
```
Crash--: I think we can use https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/contacts/Contact.js#L141

react/ContactsList/ContactRow.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+import contactPropTypes from './ContactPropTypes'
+
+import ContactPhone from './Contacts/ContactPhone'
+import ContactIdentity from './Contacts/ContactIdentity'
+import ContactCozy from './Contacts/ContactCozy'
+import ContactEmail from './Contacts/ContactEmail'
+import styles from './styles.styl'
+
+const getPrimaryOrFirst = (arr = [{}]) => arr.find(obj => obj.primary) || arr[0]
```
Crash--: Come on, we never used back our Contact model? :'( 

https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/contacts/Contact.js#L7

react/ContactsList/ContactRow.jsx
```diff
@@ -0,0 +1,63 @@
...
+    onClick,
+    ...rest
+  } = props
+  const { number: phone } = getPrimaryOrFirst(contact.phone) || {
+    number: undefined
+  }
+
+  const { address: email } = getPrimaryOrFirst(contact.email) || {
+    address: undefined
+  }
```
Crash--: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/contacts/Contact.js#L66

react/ContactsList/ContactRow.jsx
```diff
@@ -0,0 +1,63 @@
...
+  const name = contact.name || {}
```
Crash--: https://github.com/cozy/cozy-libs/blob/master/packages/cozy-doctypes/src/contacts/Contact.js#L136

react/ContactsListModal/index.jsx
```diff
@@ -0,0 +1,109 @@
...
+    return (
+      (contact.fullname && contact.fullname.toLowerCase().includes(f)) ||
+      (contact.name &&
+        contact.name.familyName &&
+        contact.name.familyName.toLowerCase().includes(f)) ||
+      (contact.name &&
+        contact.name.givenName &&
+        contact.name.givenName.toLowerCase().includes(f))
+    )
+  })
```
Crash--: We can have contact having only email's address or cozy-url (for instance if you have shared something with it) 

In drive we use a contact filter for the sharing process. Few methods are here: https://github.com/cozy/cozy-drive/blob/master/src/sharing/suggestionMatchers.js 

Since we're in the process of exporting `sharings` to cozy-libs, maybe we should also export search / filter to cozy-doctypes at least. 

Also, as @y-lohse will say : `I prefer using lodash.get instead of having to do several checks`

drazik: I'm filtering on `Contact.getDisplayName` now, which handles contacts with only email address. I added a TODO comment to extract matchers functions from drive later.

react/ContactsListModal/DemoProvider.jsx
drazik: Yes the goal is the same. I will keep it as is for now because I already used too much time on this.

react/ContactsList/ContactsList.jsx
Crash--: @drazik can you add a TODO with @ptbrowne's comment ? :)

PR #1275, opened 2019-11-28
Add ContactPicker component

drazik:
:warning: Base branch should be changed to master after https://github.com/cozy/cozy-ui/pull/1273 has been merged.

Add a `ContactPicker` components that renders like a select, but opens a `ContactsListModal`. Also add the possibility to use it with a `Field` with `type="contact"`.

See https://drazik.github.io/cozy-ui/react/#!/ContactPicker and https://drazik.github.io/cozy-ui/react/#

### DIFF
--- a/react/ContactPicker/index.jsx
+++ b/react/ContactPicker/index.jsx
@@ -1,0 +1,46 @@
+import React, { useState } from 'react'
+import ContactsListModal from '../ContactsListModal'
+import styles from './styles.styl'
+import cx from 'classnames'
+import { Contact } from 'cozy-doctypes'
+
+const SelectControl = props => {
+  const { className, children, ...rest } = props
+
+  return (
+    <button
+      type="button"
+      className={cx(styles.SelectControl, className)}
+      {...rest}
+    >
+      {children}
+    </button>
+  )
+}
+
+const ContactPicker = props => {
+  const { placeholder, onChange, selectedContact, ...rest } = props
+  const [showContactsList, setShowContactsList] = useState(false)
+
+  const handleChange = contact => {
+    onChange(contact)
+  }
+
+  return (
+    <>
+      <SelectControl {...rest} onClick={() => setShowContactsList(true)}>
+        {selectedContact
+          ? Contact.getDisplayName(selectedContact)
+          : placeholder}
+      </SelectControl>
+      {showContactsList && (
+        <ContactsListModal
+          dismissAction={() => setShowContactsList(false)}
+          onItemClick={handleChange}
+        />
+      )}
+    </>
+  )
+}
+
+export default ContactPicker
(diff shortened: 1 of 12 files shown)


stylus/components/forms.styl
```diff
@@ -428,6 +428,7 @@ $select
     appearance none
     background embedurl('../../assets/icons/illus/bottom-select.svg') right rem(16) center no-repeat
     background-size rem(14)
+    text-align left
```
ptbrowne: Why was this necessary ? Because otherwise in a context where the text-align is different (center for example), we want to reset it ?

react/ContactPicker/index.jsx
```diff
@@ -0,0 +1,46 @@
...
+  const { placeholder, onChange, selectedContact, ...rest } = props
+  const [showContactsList, setShowContactsList] = useState(false)
+
+  const handleChange = contact => {
+    onChange(contact)
+  }
+
+  return (
+    <>
+      <FakeSelect {...rest} onClick={() => setShowContactsList(true)}>
```
ptbrowne: SelectControl instead of FakeSelect ?

stylus/components/forms.styl
drazik: Because I'm using a `button` to render the fake select. And the default text align for buttons is center. But we always want the text to be aligned on the left by default when rendering something that looks like a select, no matter which underlying element is used.

react/ContactPicker/styles.styl
```diff
@@ -0,0 +1,5 @@
+@require 'tools/mixins'
+@require '../../stylus/components/forms'
```
Crash--: @require 'components/forms' should work?

react/Field/index.jsx
```diff
@@ -211,6 +222,17 @@ const Field = props => {
...
+  'date',
+  'email',
+  'password',
+  'select',
+  'textarea',
+  'text',
+  'url',
+  'number'
+]
+
```
Crash--: 👏

drazik:
> I think it'd be nice to have a small test
>
>   click on the field
>   click on a person
>   check that the person's name is in the field ?
>
> What do you think ?

Yep, I'm going to add it.

> Also in the styleguidist, the Field seems very large ? Is it expected ?

Actually, it seems the Field is the same size as other ones, but the SelectBox is slightly smaller. I think we should align the SelectBox on other field types?

ptbrowne:
>  I think we should align the SelectBox on other field types?

I am not sure but if they have two different custom sizes, I think it'd be ok yes to align one with the other.

drazik:
It had a 30rem max-width where other control types had 32rem. I set it to 32rem too.

ptbrowne:
Using a variable for both would be the best I guess, but LGTM anyway :+1: Thanks !

PR #1276, opened 2019-11-29
WIP feat: Add Pin WIP

ptbrowne:
The Pin component was coded in Banks but should be accessible from both
Banks and Drive.

- [x] Copy/pasted the code from Banks
- [x] Changed most of imports to work
- [ ] Make example in styleguidist work
- [ ] Make Pin component configurable (no need for queryConnect) so that we do not
  have to link it directly to a document that will change from app to app


y-lohse:
Is it worth reviewing now or will there be significant changes with "Make Pin component configurable"?

ptbrowne:
I am not waiting for a review right now, but I am not planning for significant changes even with "Make Pin component configurable", it will just accept the pin setting document so it should be pretty straightforward.

ptbrowne:
Will be superseded by the work of @flohhhh

PR #1277, opened 2019-11-29
Add CollectionField in labs

drazik:
This component is like a Field, but it handles an array of values.

I didn't extracted what's in cozy-contacts because it uses final-form, react-final-form, final-form-arrays and react-final-form-arrays, which weighs a total of 31kB of minified JS. Since this feature is needed for a single very simple form in banks, I thought it was not worth it at all.

I put it in the Labs because it's a first implementation for which I haven't fully tested all cases but only the one I'm interested of (collection of ContactPicker).

Also, I changed the ContactPicker's `selectecContact` prop to a more generic `value`. This is technically a breaking change, so I marked it as breaking change, even if nobody is using it for now.

See https://drazik.github.io/cozy-ui/react/#!/CollectionField


react/Labs/CollectionField/styles.styl
```diff
@@ -0,0 +1,13 @@
...
+    margin-right 0
+
+.CollectionField__addBtnIcon
+    margin-right 0.5rem
+
+.CollectionField__row
+    display flex
+    align-items center
+
+    > * + *
```
ptbrowne: TIL, I thought it wsa necessary to have & > * + *.

react/Labs/CollectionField/styles.styl
```diff
@@ -0,0 +1,13 @@
+.CollectionField__addBtn
+    margin-left 0
+    margin-right 0
+
+.CollectionField__addBtnIcon
+    margin-right 0.5rem
```
ptbrowne: Cna you use spacing_values from settings/spaces.styl ?

react/Labs/CollectionField/styles.styl
```diff
@@ -0,0 +1,13 @@
...
+
+.CollectionField__row
+    display flex
+    align-items center
+
+    > * + *
+        margin-left 1rem
```
ptbrowne: Cna you use spacing_values from settings/spaces.styl ?

react/Labs/CollectionField/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+              <div key={index} className={styles.CollectionField__row}>
+                <Component
+                  value={value}
+                  onChange={handleChange(index)}
+                  placeholder={placeholder}
+                />
+                <Button
+                  type="button"
+                  theme="secondary"
+                  label="delete"
```
ptbrowne: Nit: This is not translated.

react/Labs/CollectionField/index.jsx
```diff
@@ -0,0 +1,80 @@
+import React from 'react'
+import Button from '../../Button'
+import cx from 'classnames'
+import fieldStyles from '../../Field/styles.styl'
+import Label from '../../Label'
+import styles from './styles.styl'
+import Stack from '../../Stack'
+import Icon from '../../Icon'
+
+const CollectionField = props => {
```
ptbrowne: Can you add a docstring stating that this is a controlled component ?

react/Labs/CollectionField/index.jsx
```diff
@@ -0,0 +1,86 @@
...
+          }
+          onClick={handleAdd}
+          className={styles.CollectionField__addBtn}
+        />
+      </Stack>
+    </div>
+  )
+}
+
+export default CollectionField
```
Crash--: Even if it's an experimental component, can we have a snapshot ?

ptbrowne: Don't you think that Argos snapshots are enough ?

Crash--: Also, can we add the propTypes ?

PR #1286, opened 2019-12-03
Add empty message and add contact button to ContactsListModal

drazik:
Two things related to ContactsListModal in this PR:

* When no contact exists, we rendered an empty list. Now we can pass a message.
* A « add contact » button that redirects to contacts app when clicked has been added

See https://drazik.github.io/cozy-ui/react/#!/ContactsListModal and https://drazik.github.io/cozy-ui/react/#!/ContactPicker

### DIFF
--- a/react/ContactsListModal/AddContactButton.jsx
+++ b/react/ContactsListModal/AddContactButton.jsx
@@ -1,0 +1,36 @@
+import React from 'react'
+import { withClient } from 'cozy-client'
+import AppLinker, { generateWebLink } from '../AppLinker'
+import { ButtonLink } from '../Button'
+
+const DumbAddContactButton = props => {
+  const { client, ...rest } = props
+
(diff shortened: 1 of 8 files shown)


react/ContactsListModal/AddContactButton.jsx
```diff
@@ -0,0 +1,37 @@
...
+    cozyUrl: cozyURL.origin,
+    slug: contactsAppSlug,
+    subDomainType: cozySubDomainType
+  })
+
+  return (
+    <AppLinker slug={contactsAppSlug} href={contactsAppHref}>
+      {({ onClick, href }) => (
+        <Button
+          tag="a"
```
y-lohse: We could use `ButtonLink` instead no?

drazik: Yep. Done, thanks

PR #1287, opened 2019-12-03
fix: Don't trackEvent if no shouldEnableTracking

Crash--:
Call shouldEnableTracking in getTracker so we can't track if it's not enabled. 

shouldEnableTracking is memoized in order to avoir dom query


Crash--:
[Faire la PR sur Cozy-UI pour que le trackEvent respect le shouldEnable Tracking](https://trello.com/c/dAi2wS3D/176-faire-la-pr-sur-cozy-ui-pour-que-le-trackevent-respect-le-shouldenable-tracking)

PR #1288, opened 2019-12-03
feat: Use React component as icon

edas:
Let use `<Icon icon={Svg} />` where `Svg` is a React component
wrapping an `<svg>` tag.


edas:
@Crash-- Is it ok for you ?

PR #1289, opened 2019-12-03
fix(ContactsList): Show avatar even if name is long

drazik:
Avatar shrinked to 0px when the name was long. It doesn't shrink anymmore.

### DIFF
--- a/react/ContactsList/styles.styl
+++ b/react/ContactsList/styles.styl
@@ -59,6 +59,9 @@ $reset-list
 .contact-lastname
     font-weight bold
 
+.contact-avatar
+    flex-shrink 0
+
 +small-screen()
     .contact-identity
     .contact-phone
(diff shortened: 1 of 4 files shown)


react/ContactsList/styles.styl
```diff
@@ -59,6 +59,9 @@ $reset-list
 .contact-lastname
     font-weight bold
 
+.contact-avatar
+    flex-shrink 0
```
ptbrowne: It should be directly on the avatar right ? An avatar should never shrink IMHO.

ptbrowne: cc @y-lohse

y-lohse: Not sure. It should have a fixed size, but it already has a fixed width and height. I don't think the component itself should define how it behaves when it's put inside a flex layout. There are other ways to change the size of a component (grid layout, absolute positions, etc), imo it's the responsibility of the component user to handle those cases, like it's been done here.

drazik: Yes I don't think the Avatar should know how it should behave in all contexts in which it can be rendered.

ptbrowne: OK good point, thanks !

PR #1291, opened 2019-12-04
Handle contacts fullname in ContactsList

drazik:
When we don't have the name details but only the fullname, we show it. And
we take it into account when keying them on the name.

See https://drazik.github.io/cozy-ui/react/#!/ContactsList

### DIFF
--- a/react/ContactsList/data.json
+++ b/react/ContactsList/data.json
@@ -1,5 +1,15 @@
 [
   {
+    "_id": "013a6414-9727-4a4e-8b35-d3b9946a17c6",
+    "_type": "io.cozy.contacts",
+    "email": [
+      {
+        "address": "isabelle.durand@cozycloud.cc",
+        "primary": true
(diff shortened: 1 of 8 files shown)


react/ContactsList/data.json
```diff
@@ -1,4 +1,14 @@
...
+  {
+    "_id": "013a6414-9727-4a4e-8b35-d3b9946a17c6",
+    "_type": "io.cozy.contacts",
+    "email": [
+      {
+        "address": "isabelle.durand@cozycloud.cc",
+        "primary": true
+      }
+    ]
+  },
```
y-lohse: Why do we need this? I don't mind it but it makes the review a bit harder, especially the snapshots.

drazik: I wanted to add a contact with no fullname or name in the data, so we have this case in the styleguide.

PR #1294, opened 2019-12-05
WIP refactor: Remove the span in the button

ptbrowne:
I am pushing this PR not to let it rot on my computer and to check what
Argos says :)

- Makes it more easily usable when not using React
- Less DOM

Still some problems but I am confident they can be solved :)

https://www.argos-ci.com/cozy/cozy-ui/builds/58522


nono:
Ping @ptbrowne

ptbrowne:
I think we can close. The right path would be to use more the buttons from mui and not spend time on reimplementing or improving low level components.

PR #1296, opened 2019-12-10
fix: Removes useMemo in Icon

edas:
The missing useMemo should ne create too much noise.
This is an indirect fix for https://pad.cozycloud.cc/p/cozy-sharing
in case we do not find the root cause (yes, this is not satisfying).

Do not merge yet.


ptbrowne:
We can close right ?

edas:
yep

PR #1299, opened 2019-12-11
Update Infos component

y-lohse:
This PR updates the Infos component to suit new needs and fix old bugs, in particular:

- The line width should be capped
- More flexibility with the titles (adding subtitles, controlling their colors, …)
- Possibility to have a closing cross
- Compatibility with a InfosCarrousel component, which will be in a separate PR

The prop have changed, but there is a layer of compatibility, so no breaking changes. You can see the old components (and the new ones) in the [demo](https://yannick-lohse.fr/cozy-ui/react/#!/Infos), but I will remove the old examples before I merge this PR. The only changes are visual, the paddings and margins have slightly changed, but that is desired.

Just to be explicit about it: before this PR, we would pass a `title` prop, and the Infos component would render a `SubTitle` component. Now you are in charge of passing a `SubTitle` component to the `description` prop (or whatever other content you need, really).

### DIFF
--- a/react/Infos/index.jsx
+++ b/react/Infos/index.jsx
@@ -1,67 +1,117 @@
 import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
-import styles from './styles.styl'
-import Icon, { iconPropType } from '../Icon'
+import Icon from '../Icon'
 import Text, { SubTitle } from '../Text'
+import Stack from '../Stack'
+import palette from '../palette'
 
-const Infos = ({ actionButton, icon, isImportant, text, className, title }) => {
+import styles from './styles.styl'
+
+export const Infos = ({
+  description,
+  action,
+  dismissAction,
+  theme,
+  className
+}) => {
   return (
-    <div
-      className={cx(
-        styles['Infos'],
-        'u-stack-m',
-        isImportant ? 'u-bg-chablis' : 'u-bg-paleGrey',
-        className
+    <div className={cx(styles['Infos'], styles[`Infos--${theme}`], className)}>
+      <Stack spacing="m">
+        <div className={styles['Infos-description']}>
+          <Stack spacing="s">{description}</Stack>
+        </div>
+        {action && <div>{action}</div>}
+      </Stack>
+      {dismissAction && (
+        <button className={styles['Info-close']} onClick={dismissAction}>
+          <Icon icon="cross" color={palette['coolGrey']} />
+        </button>
       )}
-    >
-      {!!title && (
-        <SubTitle
-          className={cx({
-            'u-pomegranate': isImportant
-          })}
-        >
-          {title}
-        </SubTitle>
-      )}
-      <div className={cx('u-flex', 'u-w-100')}>
-        {icon && (
-          <Icon
-            icon={icon}
-            className={cx(
-              styles['Infos__icon'],
-              'u-w-1',
-              'u-h-1',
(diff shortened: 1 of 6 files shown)


react/Infos/index.jsx
```diff
@@ -1,73 +1,119 @@
...
+  /** The textual part of the info, including titles */
+  description: PropTypes.element.isRequired,
+  /** One or more call to actions */
+  action: PropTypes.element,
+  /** If supplied, displays a cross in the top right corner that calls the prop when clicked */
+  dismissAction: PropTypes.func,
+  /** Extra classnames to apply to the root element */
+  className: PropTypes.string,
+  /** Controls the background color of the component */
+  theme: PropTypes.oneOf(['primary', 'secondary', 'danger'])
```
y-lohse: I have my doubt about the name of this prop. In effect it really only controls the background color at the moment, though it could do more in the future..? It could impact the text color or the cross for example. And `theme`is coherent with the `Button` API. That said I could also add a default css color and leave the implementer to define the color they need. Any thoughts?

react/Infos/styles.styl
```diff
@@ -1,9 +1,16 @@
 @require 'tools/mixins'
+@require 'settings/spaces'
```
drazik: It looks like you wanted to use `spacing_values`, but you didn't. Was it your intention?

react/Infos/index.jsx
```diff
@@ -1,67 +1,117 @@
...
+    <div className={cx(styles['Infos'], styles[`Infos--${theme}`], className)}>
+      <Stack spacing="m">
+        <div className={styles['Infos-description']}>
+          <Stack spacing="s">{description}</Stack>
+        </div>
+        {action && <div>{action}</div>}
+      </Stack>
+      {dismissAction && (
+        <button className={styles['Info-close']} onClick={dismissAction}>
+          <Icon icon="cross" color={palette['coolGrey']} />
```
drazik: Is there a benefit for using `palette` over using `color="var(--coolGrey)"` ?

ptbrowne:
> Just to be explicit about it: before this PR, we would pass a title prop, and the Infos component would render a SubTitle component. Now you are in charge of passing a SubTitle component to the description prop (or whatever other content you need, really).

Is this a breaking change ?

react/Infos/styles.styl
y-lohse: Yes, good catch, thanks.

react/Infos/index.jsx
y-lohse: Not that I'm aware, it seems to be the more widely used form though. Don't mind changing it if we'd rather move away from it.

y-lohse:
> Is this a breaking change ? Why do we change the API ? Could a codemod be considered instead of keeping the old API ?

There is no breaking change in this PR. There will be one the day we remove the compatibility layer, and I'm in favor of providing a codemod at that point.

The API is changing because the old props wouldn't allow us enough flexibility, notably on title colors. It would have also required adding several new props, when we can rely on more composition instead :)

I'll ok into that extra div, thanks! I think it might just be the doc preview that is slightly out of date.

react/Infos/index.jsx
drazik: OK thanks. It was just for my culture ;)

y-lohse:
Cool! I'll remove the outdated examples from the docs and merge this.

PR #1300, opened 2019-12-11
Add tag prop on Stack and force column layout

drazik:
The tag prop enables us to change the tag used for the rendering of the Stack. It's useful to avoid rendering too many useless intermediary elements.

When some inline elements were children of a Stack, the margins were not
applied since vertical margins have no effect on inline elements. By adding
a flex display in column direction, we ensure that no matter which type of
element is rendered, it will have margins applied.

See https://drazik.github.io/cozy-ui/react/#!/Stack

### DIFF
--- a/react/Stack/index.jsx
+++ b/react/Stack/index.jsx
@@ -3,9 +3,9 @@ import cx from 'classnames'
 import PropTypes from 'prop-types'
 import styles from './styles.styl'
 
-const Stack = ({ spacing, ...props }) => {
+const Stack = ({ spacing, tag: Tag, ...props }) => {
   return (
-    <div
+    <Tag
       {...props}
       className={cx(props.className, spacing && styles['Stack--' + spacing])}
     />
@@ -17,7 +17,8 @@ Stack.propTypes = {
 }
 
 Stack.defaultProps = {
-  spacing: 'm'
+  spacing: 'm',
+  tag: 'div'
 }
 
 export default Stack


ptbrowne:
LGTM ? I am just curious about why you would not use only block elements in the Stack ?

drazik:
> I am just curious about why you would not use only block elements in the Stack ?

I have labels that look like titles.

So I used

```jsx
<Stack spacing="m">
  <SubTitle tag="label">Label title</SubTitle>
  <Input />
</Stack>
```

react/Stack/styles.styl
```diff
@@ -1,5 +1,9 @@
 @require 'utilities/spaces.styl'
 
+.Stack
+    display flex
+    flex-direction column
+
```
y-lohse: Switching to flexbox has some side effects that I'm not sure we want. If I'm not mistaken, elements will now `stretch` by default on the width (https://developer.mozilla.org/en-US/docs/Web/CSS/align-items), and all elements will be allowed to shrink.

IMO it's a bit risky, it can work if we set all the properties right but maybe applying a `display inline-block` on the `inline` elements is a better solution?

drazik: We can see via Argos that in the current state it effectively has undesirable effects (on CollectionField). I think I'm going to split the second commit in its own PR since it's the problematic one.

drazik: See https://github.com/cozy/cozy-ui/pull/1301

PR #1301, opened 2019-12-11
feat(Stack): Force column rendering

drazik:
When some inline elements were children of a Stack, the margins were not
applied since vertical margins have no effect on inline elements. By
adding a flex display in column direction, we ensure that no matter
which type of element is rendered, it will have margins applied.

This was originally submitted in https://github.com/cozy/cozy-ui/pull/1300, but it causes some visual regression in CollectionField (see https://github.com/cozy/cozy-ui/pull/1300#pullrequestreview-330530348). So we may need to add some flex properties to avoid problems.

### DIFF
--- a/react/Stack/index.jsx
+++ b/react/Stack/index.jsx
@@ -7,7 +7,11 @@ const Stack = ({ spacing, ...props }) => {
   return (
     <div
       {...props}
-      className={cx(props.className, spacing && styles['Stack--' + spacing])}
+      className={cx(
+        props.className,
+        styles.Stack,
+        spacing && styles['Stack--' + spacing]
+      )}
     />
   )
 }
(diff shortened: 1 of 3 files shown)


y-lohse:
So I think we need at least `align-items: flex-start`. The other tricky situation is if the Stack container has a limited height, in that case flexbox will probably start messing with elements heights.
Another option we can consider is explicitly stating that Stack children need to be `block` or `inline-block`. It makes `Stack` less practical but I don't know how often we really have inline children in a stack.

Crash--:
@drazik What is the status of this PR?

drazik:
I finally didn't need it. If nobody else had this problem, I think we can close, as it seems it will add more problems than it solves.

ptbrowne:
OK for me :)

PR #1302, opened 2019-12-12
Infos Carrousel

y-lohse:
This PR goes with #1299, I'll update the docs preview once it's merged. It adds a `InfosCarrousel` component which can be used to cycle between several `Infos` components.

The API is basically

```
<InfosCarrousel>
  <FirstInfo />
  <SecondInfo />
</InfosCarrousel>
```

One thing that I'm not super happy with is the `theme` prop. Both `Infos` and `InfosCarrousel` have a `theme` prop, which controls the background color for the Infos and the arrows color for the Carrousel. But at this point you have to specify both separately, which means you could have a `<InfosCarrousel theme="danger"><Infos theme="primary" /><Infos theme="secondary" /></InfosCarrousel>`… But it's hard to say if we want that or not, at the moment our use case is to have the same theme everywhere.


y-lohse:
Demo visible here: https://yannick-lohse.fr/cozy-ui/react/#!/InfosCarrousel

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <div
+      className={cx(
+        styles['InfosCarrousel'],
+        styles[`InfosCarrousel--${theme}`],
+        className
+      )}
+    >
+      <SwipeableViews index={index} animateHeight {...swipleableProps}>
+        {children}
```
ptbrowne: To remove the need to repeat the theme property on the children, we can do `React.Children.map(children x => React.clone(x, { theme }))`

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        </button>
+      </div>
+    </div>
+  )
+}
+
+InfosCarrousel.propTypes = {
+  /** A collection of Infos component that will be rendered, one per carrousel slide */
+  children: PropTypes.element,
+  /** COntrols the color scheme of the carrousel */
```
ptbrowne:
```suggestion
  /** Controls the color scheme of the carrousel */
```

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+        >
+          <Icon icon="right" />
+        </button>
+      </div>
+    </div>
+  )
+}
+
+InfosCarrousel.propTypes = {
+  /** A collection of Infos component that will be rendered, one per carrousel slide */
```
ptbrowne:
```suggestion
  /** Infos components that will be rendered, one per carousel slide */
```

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+  children: PropTypes.element,
+  /** COntrols the color scheme of the carrousel */
+  theme: PropTypes.oneOf(['primary', 'secondary', 'danger']),
+  /** Extra classes to apply to the main wrapper */
+  className: PropTypes.string,
+  /** Props to pass to the underlying react-swipeable-views component */
+  swipleableProps: PropTypes.object
```
ptbrowne:
```suggestion
  swipeableProps: PropTypes.object
```

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+  return (
+    <div
+      className={cx(
+        styles['InfosCarrousel'],
+        styles[`InfosCarrousel--${theme}`],
+        className
+      )}
+    >
+      <SwipeableViews index={index} animateHeight {...swipleableProps}>
```
ptbrowne:
```suggestion
      <SwipeableViews index={index} animateHeight {...swipeableProps}>
```

react/InfosCarrousel/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+import SwipeableViews from 'react-swipeable-views'
+
+import styles from './styles.styl'
+import Icon from '../Icon'
+
+const InfosCarrousel = ({ children, theme, className, swipleableProps }) => {
+  const count = React.Children.count(children)
+  const [index, setIndex] = useState(0)
+  const nextInfos = useCallback(() => setIndex(index + 1), [index])
+  const previousInfos = useCallback(() => setIndex(index - 1), [index])
```
ptbrowne: TIL useCallback

ptbrowne: IMHO, nextInfos and previousInfos, since they are functions should be named with verbs : goToPrevious && goToNext.

react/InfosCarrousel/styles.styl
```diff
@@ -0,0 +1,31 @@
...
+
+.InfosCarrousel
+    position relative
+
+    &.InfosCarrousel--primary .InfosCarrousel-arrow,
+    &.InfosCarrousel--secondary .InfosCarrousel-arrow
+        color var(--coolGrey)
+
+    &.InfosCarrousel--danger .InfosCarrousel-arrow
+        color var(--pomegranate)
```
ptbrowne: We have to keep this pattern in mind : theme dependent buttons. I think this could be extracted at some point.

react/InfosCarrousel/index.jsx
y-lohse: I thought about that, but it feels like a bad practice to forcefully inject props to child components… or is it just me?

ptbrowne: I think there a times when it is necessary, mainly when the coupling between the parent and the child is very visible, and expected.

I use the same strategy for example with the Slides inside a Swiper component. They receive props like isActiveSlide.

> it feels like a bad practice to __forcefully__ inject props to child components

A check not to override an existing theme prop would be good I think.

ptbrowne: Another possible technique would be to use a themeContext, a provider inside themeInfos, an a consumer inside Infos.

```
const Infos = () => {
  let contextTheme = useContext(ThemeContext)
  let resolvedTheme = props.theme || contextTheme
   ....
}
```

react/InfosCarrousel/styles.styl
ptbrowne: Those buttons could use a ThemeContext, like discussed in https://github.com/cozy/cozy-ui/pull/1302#discussion_r359225774

react/InfosCarrousel/index.jsx
ptbrowne: Another way to implement this `context` need is to use the cascading styles of CSS.

```
.themePrimary .InfosSlide
  background var(--primaryColor)

.themePrimary .Infos
  background var(--primaryColor)
```

y-lohse: Cascading css isn't possible I think because of css modules — `Infos` and `InfosCarrousel` are 2 separate modules. The context would work, but I'll go ahead with the `clone` technique, checking that there is no theme prop we would override 👍

ptbrowne: In this case, I wonder if Infos and InfosCarrousel should not share the same stylus module. Anyway, this goes beyond the scope of the PR.

> with the clone technique, checking that there is no theme prop we would override 👍

+1

PR #1303, opened 2019-12-12
feat: Add Em component

drazik:
~~This component wraps its children in a slight padding and a border. It's extracted from the new cozy-settings passphrase update form. (see https://github.com/cozy/cozy-settings/pull/269)~~

~~The name come from the discussion I had with Joël about the extraction of this component. He said « c'est comme un highlight ». If you have a better idea, please tell me :)~~

~~See https://drazik.github.io/cozy-ui/react/#!/HighlightedText/0~~

See https://github.com/cozy/cozy-ui/pull/1303#issuecomment-565116979

### DIFF
--- a/react/Text/Readme.md
+++ b/react/Text/Readme.md
@@ -53,3 +53,11 @@ import { Caption } from 'cozy-ui/transpiled/react/Text';
 
 <Caption>This a caption text</Caption>
 ```
+
+#### Emphasis text
+
+```jsx
+import { Em } from 'cozy-ui/transpiled/react/Text';
+
+<Em>This is an emphasis text</Em>
+```
--- a/react/Text/index.jsx
+++ b/react/Text/index.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
+import styles from './styles.styl'
 
 export const BaseText = props => {
   const { className, children, tag, ellipsis, ...restProps } = props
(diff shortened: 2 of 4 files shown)


ptbrowne:
Browsing on a thesaurus, I eventually came across Emphasis

image


I would prefer to have a Noun for the name of the components.

- [Emphasis](https://en.wikipedia.org/wiki/Emphasis_(typography)) seems good ?

<Em variant='bordered' /> ?

drazik:
`<Em>` looks OK to me. @y-lohse any thoughts on this?

drazik:
I just added `Em` to the Text components. See https://drazik.github.io/cozy-ui/react/#!/BaseText

ptbrowne:
cc @joel-costa

y-lohse:
Not against it and the implementation LGTM, but we also have Bold that serves a similar purpose… are we sure it's a highlight component? Do we have other use cases?

ptbrowne:
> There is already Bold

This is why I think we can combine the two: `<Em variant='bold' />`

The composant is semantic and the variant is more about the design, like the Buttons.

y-lohse:
Just checked with @joel-costa , the Settings Em blip should be a [`InlineCard`](https://docs.cozy.io/cozy-ui/react/#inlinecard), it's not a new component :)

ptbrowne:
This could maybe be clearer in the styleguide. Emphasis components : Bold, InlineCard

ptbrowne:
What should we do about this PR ?

y-lohse:
IMO we can close it. @drazik ?

> This could maybe be made clearer in the styleguide. Emphasis components : Bold, InlineCard ?

If `inlineCard` is indeed used to highlight text, it makles sense to group it with the other text components 👍

drazik:
I'm closing. I used InlineCard

PR #1306, opened 2019-12-12
feat: Use hook for i18n

ptbrowne:
- I18n provider implemented through new context API
- I18n provider still provider backward compatibility with old context API
- translate() is kept retro compatible
- useI18n() hook provides access to t, f

- A codemod provides basic auto migration

```js
import { translate } from "cozy-ui/transpiled/react"
const Component = ({t, f}) => {}
export default translate()(Component)
```

becomes

```js
import { translate } from "cozy-ui/transpiled/react"
import { useI18n } from "cozy-ui/transpiled/react"

const Component = ({}) => {
 const { t, f } = useI18n()
}

export default Component
```

See https://github.com/cozy/cozy-banks/pull/1723

It can be improved:

- [x] import of useI18n could be merged with the one from cozy-ui
- [x] Empty props object could be removed if t was the only property
- [x] translate could be removed from the imports if it is no longer used

https://astexplorer.net/#/gist/dbda709790627b25a6b23a5b51564d09/3db7786b30d8d1c8be0953c06893ae07dce89e51


y-lohse:
Completely in favor of this, I'll review later this morning \o/

codemods/use-i18n-hook.js
```diff
@@ -0,0 +1,128 @@
+// Press ctrl+space for code completion
```
y-lohse: I don't think we need this? 😅

codemods/use-i18n-hook.js
```diff
@@ -0,0 +1,128 @@
...
+
+export default function transformer(file, api) {
+  const j = api.jscodeshift
+  const root = j(file.source)
+
+  const isI18nProp = prop => {
+    return prop.key && (prop.key.name === 't' || prop.key.name === 'f')
```
y-lohse: There's also the `lang` prop that is injected by translate, it would be good to migrate it as well.

codemods/utils.js
```diff
@@ -1,46 +1,123 @@
-/* eslint new-cap: 0 */
+// Press ctrl+space for code completion
```
y-lohse: Not needed either I believe :)

codemods/utils.js
```diff
@@ -1,46 +1,123 @@
...
 
-module.exports = function (j) {
-  const strip = s => s.replace(/^[\s\n]+/g, "").replace(/[\s\n]+$/g, "");
+const groupBy = (arr, fn) => {
```
y-lohse: How does this differ from lodash.groupBy?

codemods/utils.js
```diff
@@ -1,46 +1,123 @@
...
+    return false
+  }
+  return reactSpecificIdentifiers.includes(spec.imported.name)
+}
 
-  const isJsxElementOfClass = (element, klass) => {
-    return element.openingElement && element.openingElement.name.name === klass;
-  };
+export default function transformer(file, api) {
+  console.log('hello')
```
y-lohse: Can be removed now :)

codemods/use-i18n-hook.js
ptbrowne: It was there originally but I saw than in Banks it would migrate by error a component that was not included in the I18n provider (the top level App component, that takes a lang). So I removed it.

I could maybe make this dynamic so that the user of the codemod can choose whether to migrate components using lang.

ptbrowne:
> there seems to be a lot of duplication between utils.js and use-i18n-hook.js, I don't think it was intended?

😱 I am sorry, I first copy pasted by error the content of the codemod inside utils, and then I committed it by error. This is fixed now, `utils.js` does not change.

codemods/use-i18n-hook.js
y-lohse: Yep I think that would be great! Otherwise I think we risk ending up with leftover `lang` props that won't work.

y-lohse:
No problem, didn't take to long to realize something was wrong 😛. Changes LGTM, just that `lang` prop migration needs adressing imo.

codemods/use-i18n-hook.js
ptbrowne: > Otherwise I think we risk ending up with leftover lang props that won't work.

No, we will just end up with components using `lang` that will just not use useI18n and stay with translate(). This is not a big problem IMHO.

ptbrowne:
I refactored just a bit to extract helpers. I think it is important before merging that I address the checklist in the description.

codemods/use-i18n-hook/example-before.js
```diff
@@ -0,0 +1,28 @@
...
+import OtherImports from 'other/imports'
+
+const ComponentUsingT = translate()(({ t }) => (
+  <Padded>
+    <Title>{t('Title1')}</Title>
+    <Text>{t('Text1')}</Text>
+  </Padded>
+))
+
+const ComponentUsingTAndF = ({ t, f }) => (
```
edas:
```suggestion
const DumbComponentUsingTAndF = ({ t, f }) => (
```

react/I18n/index.jsx
```diff
@@ -25,11 +27,13 @@ export class I18n extends Component {
     this.translator =
       polyglot || initTranslation(lang, dictRequire, context, defaultLang)
     this.format = initFormat(lang, defaultLang)
+    this.t = this.translator.t.bind(this.translator)
+    this.contextValue = this.getContextValue()
   }
 
-  getChildContext() {
```
edas: En renommant ici tu casses la compatibilité avec l'ancienne API de contexte, non ?

ptbrowne: Oui mais normalement cette API de contexte est un détail d'implémentation vu que les utilisateurs utilisent `translate()` et `I18n`. Après ca peut quand même casser des tests/des endroits ou on utilise directement l'ancienne API de contexte directement au lieu d'I18n.

Tu penses que je devrais laisser getChildContext pour utiliser les 2 APIs en même temps ?

codemods/use-i18n-hook/example-before.js
ptbrowne: fix: Example had the wrong variable name

react/I18n/index.jsx
Crash--: Dans drive j'ai encore du : 
```
 static contextTypes = {
    t: PropTypes.func.isRequired
  }
```

Donc si ça joue là-dessus j'imagine que oui il faudrait garder les deux.

ptbrowne: OK j'ai remis le getChildContext et rajouté des tests là dessus. 

https://github.com/cozy/cozy-ui/pull/1306/commits/75d907297b90f779f9ccffd9b143789c0bc827f4

react/AppSections/components/SmallAppItem.jsx
```diff
@@ -61,7 +61,6 @@ export const SmallAppItem = ({
 }
 
 SmallAppItem.propTypes = {
-  t: PropTypes.func.isRequired,
```
y-lohse: Why is the proptype removed?

codemods/use-i18n-hook/example-after.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <Padded>
+      <Title>{t('Title2')}</Title>
+      <Text>{f(new Date())}</Text>
+    </Padded>
+  );
+}
+
+const EnhancedComponentUsingTAndF = translate()(DumbComponentUsingTAndF)
```
edas: Are you sure this line is left after codemod ?

codemods/use-i18n-hook/example-after.js
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <Padded>
+      <Title>{t('Title1')}</Title>
+      <Text>{t('Text1')}</Text>
+    </Padded>
+  );
+}
+
+const ComponentUsingTAndF = () => {
```
edas: Shouldn't this be EnhancedComponentUsingTAndF or DumbComponentUsingTAndF ?

package.json
```diff
@@ -48,7 +48,9 @@
     "screenshots": "node scripts/screenshots.js --screenshot-dir ./screenshots --styleguide-dir ./build/react",
     "test": "yarn test:jest",
     "test:jest": "jest --verbose",
-    "transpile": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --verbose && babel helpers/ --out-dir transpiled/helpers --verbose",
+    "transpile:react": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --verbose",
+    "transpile:helpers": "babel helpers/ --out-dir transpiled/helpers --verbose",
```
edas: Don't we need the `env BABEL_ENV=transpilation` here  ?

ptbrowne: Maybe but since it was not present in the first line (in the second command), I think it should not be put here.

codemods/use-i18n-hook/example-after.js
ptbrowne: The translate() is removed but the line is left.

codemods/use-i18n-hook/example-after.js
ptbrowne: Fixed when re-ran the codemod.

react/AppSections/components/SmallAppItem.jsx
ptbrowne: I don't know, I guess it was to remove a warning during tests. Since it is not part of the PR, I removed the removal.

PR #1307, opened 2019-12-13
fix: Always use transpiled cozy-ui

edas:
Some files did use an absolute path for cozy-ui
loading from `cozy-ui/react/…`. This breaks
if we use the components from a transpiled path
(it will escape the transpiled to go into non transpiled)

----

In early versions of B2B overrides, we needed to have absolute path for the files we wanted to overide. Files modified in this PR were using absolute path for that motive.

Today, our babel configuration does rewrite all path in order to use absolute path pointing at the transpiled version of each file. We don't need to write absolute path ourself in cozy-ui anymore (and we shouldn't).

### DIFF
--- a/react/AppSections/Sections.jsx
+++ b/react/AppSections/Sections.jsx
@@ -2,8 +2,8 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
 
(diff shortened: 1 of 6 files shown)


ptbrowne:
This was done for customisation purposes. For overrides.

Issue #1304, opened 2019-12-12
Extract new password input from cozy-settings


drazik:
The input password from cozy-settings can replace the current Field with password type we have here. IMHO it doesn't seem very straightforward to achieve this and keep the current Field API (secondaryLabels). It may require a breaking change, or implement a totally different component.

See https://github.com/cozy/cozy-settings/pull/269

![image](https://user-images.githubusercontent.com/1606068/70734563-c3cc9880-1d0c-11ea-8125-45aeddf08a7b.png)

PR #1311, opened 2019-12-16
feat: Add PasswordInput to Labs components

drazik:
This is a new form of input field. But it needs to stabilized and maybe merged with current Input/Field. Labs seems to be a good place for it at the moment.

Related to https://github.com/cozy/cozy-ui/issues/1304

See https://github.com/cozy/cozy-settings/pull/269#discussion_r357225768

See https://drazik.github.io/cozy-ui/react/#!/PasswordInput

### DIFF
--- a/react/Labs/PasswordInput/Readme.md
+++ b/react/Labs/PasswordInput/Readme.md
@@ -1,0 +1,44 @@
+This component takes the same props as the `Input` component.
+`Input` props are spread to the underlying `Input`.
+
+### Original
+
+```jsx
+import PasswordInput from 'cozy-ui/transpiled/react/Labs/PasswordInput';
+const initialState = { password: '' };
+
+<PasswordInput
+  value={state.password}
+  onChange={e => setState({ password: e.target.value })}
+  placeholder="Enter your password"
+/>
+```
+
+### With strength
+
+```jsx
+import PasswordInput from 'cozy-ui/transpiled/react/Labs/PasswordInput';
+const initialState = { password: '' };
+
+<PasswordInput
+  value={state.password}
+  onChange={e => setState({ password: e.target.value })}
+  placeholder="Enter your password"
+  showStrength
+/>
+```
+
+### Errored
+
+```jsx
+import PasswordInput from 'cozy-ui/transpiled/react/Labs/PasswordInput';
+const initialState = { password: '' };
+
+<PasswordInput
+  value={state.password}
+  onChange={e => setState({ password: e.target.value })}
+  placeholder="Enter your password"
+  showStrength
+  error
+/>
+```
--- a/react/Labs/PasswordInput/helpers.js
+++ b/react/Labs/PasswordInput/helpers.js
@@ -1,0 +1,62 @@
+const charsets = [
+  // upper
+  { regexp: /[A-Z]/g, size: 26 },
+  // lower
+  { regexp: /[a-z]/g, size: 26 },
+  // digit
+  { regexp: /[0-9]/g, size: 10 },
+  // special
+  { regexp: /[!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]/g, size: 30 }
+]
+
+/*
+ * Returns the strength (percentage and label) of a given password
+ *
+ * @param {String} password
+ * @returns {Object} An object with percentage and label properties
+ * representing the strenght of the password
+ */
+export const getStrength = password => {
+  if (!password && password !== '') {
+    throw new Error('password parameter is missing')
+  }
+  if (!password.length) {
+    return { percentage: 0, label: 'weak' }
+  }
+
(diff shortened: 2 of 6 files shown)


react/Labs/PasswordInput/Readme.md
```diff
@@ -0,0 +1,44 @@
+This component also takes same props as the `Input` component.
```
ptbrowne:
```suggestion
This component takes the same props as the `Input` component.
```

react/Labs/PasswordInput/helpers.js
```diff
@@ -0,0 +1,56 @@
+// Return given password srength as an object {percentage, label}
```
ptbrowne: Please use a proper docstring.

s/Return/Return

react/Labs/PasswordInput/helpers.js
```diff
@@ -0,0 +1,56 @@
...
+  const charsets = [
+    // upper
+    { regexp: /[A-Z]/g, size: 26 },
+    // lower
+    { regexp: /[a-z]/g, size: 26 },
+    // digit
+    { regexp: /[0-9]/g, size: 10 },
+    // special
+    { regexp: /[!@#$%^&*()_+\-=[\]{};':"\\|,.<>/?]/g, size: 30 }
+  ]
```
ptbrowne: Those could be extracted from the function. This way we do not create new regexps at each run.

react/Labs/PasswordInput/helpers.js
```diff
@@ -0,0 +1,56 @@
...
+  ) {
+    // between 33% and 66%
+    strengthPercentage = (passwordStrength * 66) / _at66percent
+    strengthLabel = 'moderate'
+  } else {
+    // passwordStrength > 192
+    strengthPercentage = (passwordStrength * 100) / _at100percent
+    if (strengthPercentage > 100) strengthPercentage = 100
+    strengthLabel = 'strong'
+  }
```
ptbrowne: Seems like this logic is repeated and could be improved.

- strength labels could be put in a mapping linking label to password strength.

const strengthLabels = {
  weak: 0,
  moderate: 50,
  strong: 100
}

- levels could be extracted

- 100PercentStrength = strengthLabels.strong


I don't think there's the need to have _at33percent and _at66percent if the 100percent is known.

react/Labs/PasswordInput/helpers.js
```diff
@@ -0,0 +1,56 @@
...
+    return possibleChars
+  }, 0)
+
+  const passwordStrength =
+    Math.log(Math.pow(possibleChars, password.length)) / Math.log(2)
+
+  // levels
+  const _at33percent = 50
+  const _at66percent = 100
+  const _at100percent = 150
```
y-lohse: I think I recognize this from zxcvbn, is it the same extracted from there? If so did you consider straight up using zxcvbn as a dependency?

react/Labs/PasswordInput/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+      value={strength.percentage}
+      className={cx(
+        styles.PasswordField__strength,
+        styles[`PasswordField__strength--${strength.label}`],
+        className
+      )}
+      {...rest}
+    />
+  )
+}
```
y-lohse: Not sure how it looks, but would it be possible to use https://docs.cozy.io/cozy-ui/react/#percentageline by any chance?

react/Labs/PasswordInput/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+const PasswordInput = props => {
+  const { className, showStrength, error, ...rest } = props
+  const [hidden, setHidden] = useState(true)
+
+  return (
+    <div className={cx(styles.PasswordField, className)}>
```
y-lohse: Nit, but we should avoid mixing `styles.PasswordField` and `styles['PasswordField']` 😁

react/Labs/PasswordInput/index.jsx
y-lohse: After looking at the css, I guess it's not similar. Might still be worth extracting it into a standalone component eventually.

react/Labs/PasswordInput/helpers.js
drazik: Actually I extracted it from [cozy-settings](https://github.com/cozy/cozy-settings/blob/master/src/lib/passwordHelper.js). I don't know if that's code extracted from zxcvbn, so I never considered using zxcvbn as a dependency.

react/Labs/PasswordInput/helpers.js
drazik: I added tests and created https://github.com/cozy/cozy-ui/issues/1315 so your suggestion can be done later with the confidence of the tests. I need to move forward quickly at the moment.

PR #1314, opened 2019-12-16
feat: Add OrderedList component

drazik:
This is used in the password manager onboarding app. This is the ordered version of the previously added UnorderedList.

image


react/OrderedList/styles.styl
```diff
@@ -0,0 +1,19 @@
+counter-name = cozy-ui-ordered-list
+
+.OrderedList
+    list-style none
+    counter-reset counter-name
+    padding-left 0
+    line-height 1.5
+
+.ListItem
+    counter-increment counter-name
```
y-lohse: Any reason for not using `list-style-type: decimal` and generating the numbers ourselves?

drazik: Because we want to style the decimals. When I looked at how to style these, I found https://css-tricks.com/custom-list-number-styling/ that uses the counters.

y-lohse: Ah right we want them to be bold 👍

PR #1316, opened 2019-12-18
WIP feat(TextField): Add inverted prop

drazik:
This prop is used to show an « inverted » version of the TextField. E.g.
a version that looks good on a primary color background.

MUI doesn't have a prop to manage different « looks » of the TextField,
so I cheated by wrapping the component in a `div` with a particular
class name when `inverted` is true.

I'm not fully happy with this, but I didn't see another way to achieve it. If you have something better in mind, please tell me.

See https://drazik.github.io/cozy-ui/react/#!/TextField

### DIFF
--- a/react/MuiCozyTheme/TextField/Readme.md
+++ b/react/MuiCozyTheme/TextField/Readme.md
@@ -1,4 +1,6 @@
-Cozy themed MUI TextField. It is only a re-export of the component from MUI.
+Cozy themed MUI TextField. See
+[MUI V3 documentation](https://v3.material-ui.com/api/text-field/) to find the
+props.
 
 ```
 import MuiCozyTheme from 'cozy-ui/transpiled/react/MuiCozyTheme';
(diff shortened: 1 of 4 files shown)


ptbrowne:
Could we use a theme for this ?

`<ThemeProvider theme={invertedTheme}>`

drazik:
> Could we use a theme for this ?
> 
> > You can nest multiple theme providers.
> 
> [material-ui.com/customization/theming/#nesting-the-theme](https://material-ui.com/customization/theming/#nesting-the-theme)
> 
> `<ThemeProvider theme={invertedTheme}>`

Nice! I didn't see these piece of documentation. I checked on the MUI v3 version (since we don't use v4 yet), and there's the same concept: https://v3.material-ui.com/css-in-js/advanced/#theme-nesting

So I propose to update [`MuiCozyTheme`](https://github.com/cozy/cozy-ui/blob/master/react/MuiCozyTheme/index.jsx) to take a `variant` prop that can be `normal` or `inverted` (`primary` maybe?), and in [theme.js](https://github.com/cozy/cozy-ui/blob/master/react/MuiCozyTheme/theme.js) build two themes and have a function that return one or the other according to the given variant.

Then in the apps, we can do

```jsx
<MuiCozyTheme>
  <SomeComponent />
  <MuiCozyTheme variant="inverted">
    <SomeOtherComponent />
  </MuiCozyTheme>
</MuiCozyTheme>
```

y-lohse:
Using themes sounds good, much better than adding a prop. In terms of names, we thought of `light` and `dark` with @joel-costa when he first showed me these fields. `primary` and `inverted` also works imo.

ptbrowne:
> So I propose to update MuiCozyTheme to take a variant prop that can be normal or inverted (primary maybe?), and in theme.js build two themes and have a function that return one or the other according to the given variant.

OK for me, normal/inverted.

I think light/dark do not convey the same concepts and could be mistaken for something else. There could be a normal/light and a normal/dark as well as an inverted/light and a inverted/dark.

drazik:
See https://github.com/cozy/cozy-ui/pull/1317

PR #1317, opened 2019-12-19
feat(MuiCozyTheme): Add inverted theme for TextField

drazik:
The inverted theme is made to look fine on non-white backgrounds. For
now we only have the TextField look for this theme, but with this
commit, everything should be ready to implement the theme for other MUI
components.

The discussion leading to this new theme was on https://github.com/cozy/cozy-ui/pull/1316. I decided to open another PR instead of amend the whole content of the previous one to avoid losing the context.

See https://drazik.github.io/cozy-ui/react/#!/TextField

### DIFF
--- a/react/MuiCozyTheme/theme.js
+++ b/react/MuiCozyTheme/theme.js
@@ -5,7 +5,7 @@ const defaultValues = {
   borderRadius: 6
 }
 
-export const theme = createMuiTheme({
+export const normalTheme = createMuiTheme({
   typography: {
     useNextVariants: true,
     fontFamily: getCssVariableValue('primaryFont'),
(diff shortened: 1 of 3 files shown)


react/MuiCozyTheme/theme.js
```diff
@@ -258,7 +259,61 @@ theme.overrides = {
...
-theme.shadows[8] =
+normalTheme.shadows[8] =
   '0rem 0.125rem 0.375rem 0rem rgba(50, 54, 63, .19), 0rem 0.375rem 1.25rem 0rem rgba(50, 54, 63, .19)'
 
-export default theme
+export const invertedTheme = {
+  ...normalTheme,
+  palette: {
+    ...normalTheme.palette,
+    type: 'dark',
```
ptbrowne: Nice we were discussing this with Joel, I had the feeling that the dark color was a "coincidence" for the inverted palette (you could have a normal palette that was dark, and its inverted would be light) and that the theme should contain this information (see https://github.com/cozy/cozy-ui/pull/1316, and my awkward `useDarkBackground`.

I am glad that material UI support this 🎉

ptbrowne:
Real nice, thanks :tada

Issue #1318, opened 2019-12-19
What should we do with console.warnings emitted in tests


ptbrowne:
Problem:

We have a lot of warnings during our test suites and we have learned to
ignore them. This is problematic since some warnings are indicative of
problems in code. The problem right now is that those problematic
warnings are drown into a see of non problematic one.

My proposition is to disable output of warnings that are not problematic
during CI tests. We can have a list of non problematic warnings that we
know of and plan to solve later.

List of warnings that are deemed not problematic

- Lifecycle warnings: sometimes they are due to external libs that have
  not updated their code. In my implementation, I disable those warning
  only for some components, to be more fine grained and only disable
  those warnings for the components we know there is no problem in.
- Aria label not there: I actually don't know too much what to here, it
  is not a big problem per say, but they are still adding a lot of noise.

The end goal is to cleanup all warnings from tests and then to be much
more strict on warnings. Typically I think we should transform them into
exceptions so that we are cleaner in tests.

PR #1319, opened 2019-12-19
Less warnings

ptbrowne:
- Removed expected warnings
- Ignore non problematic warnings

See [issue](https://github.com/cozy/cozy-ui/issues/1318) for more information

### DIFF
--- a/react/Field/index.jsx
+++ b/react/Field/index.jsx
@@ -246,7 +246,8 @@ const allowedTypes = [
   'textarea',
   'text',
   'url',
-  'number'
+  'number',
+  'contact'
 ]
 
 Field.propTypes = {
--- a/react/__snapshots__/examples.spec.jsx.snap
+++ b/react/__snapshots__/examples.spec.jsx.snap
@@ -5090,7 +5090,20 @@ exports[`InputGroup should render examples: InputGroup 6`] = `
 </div>"
 `;
 
-exports[`InputGroup should render examples: InputGroup 7`] = `"<div></div><pre class=\\"rsg--root-0\\">ReferenceError: InputGroup is not defined</pre>"`;
+exports[`InputGroup should render examples: InputGroup 7`] = `
+"<div>
+  <form>
+    <div>
+      <div class=\\"styles__c-inputgroup___12OVJ u-bdrs-3\\">
+        <div class=\\"styles__c-inputgroup-side___60v0v\\">
+          <div class=\\"u-title-h4 u-pl-1\\">text</div>
+        </div>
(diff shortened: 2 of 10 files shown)


react/Field/index.jsx
```diff
@@ -246,7 +246,8 @@ const allowedTypes = [
   'textarea',
   'text',
   'url',
-  'number'
+  'number',
+  'contact'
```
y-lohse: Is that a real input type? I've never heard of it and a quick search doesn't return anything.

test/jestsetup.js
```diff
@@ -5,3 +5,56 @@ configure({ adapter: new Adapter() })
...
+
+const ignoreOnConditions = (originalWarn, ignoreConditions) => {
+  return function(...args) {
+    const msg = args[0]
+    if (ignoreConditions.some(condition => condition(msg))) {
+      return
+    }
+    throw new Error(
+      'Code in tests should not emit any warning. Warning emitted:' + msg
+    )
```
y-lohse: Hardcore… but ok I think, it's not too difficult to add exceptions and promotes good practice.

ptbrowne: In fact it was not meant to be in this commit. I prefer to remove it for now and add it later.

react/Field/index.jsx
ptbrowne: Yes it is real :) https://github.com/cozy/cozy-ui/blob/master/react/Field/index.jsx#L182

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -5754,7 +5767,7 @@ exports[`Spinner should render examples: Spinner 5`] = `
     <div class=\\"styles__c-spinner___1snK7\\"><svg class=\\"styles__icon___23x3R styles__icon--spin___ybfC1\\" style=\\"fill: var(--primaryColor);\\" width=\\"80\\" height=\\"80\\">
         <use xlink:href=\\"#spinner\\"></use>
       </svg>
-      <p>loading.hi</p>
+      <p>loading.helloWorld</p>
```
Crash--: We should have `Hello World !` with your modification, no? (https://github.com/cozy/cozy-ui/pull/1319/files#diff-aae4f4155d6342669aceedd221607d8eR58)

ptbrowne: Yes, this is weird, the Argos screenshot had the change. Looking into it.

ptbrowne: https://github.com/cozy/cozy-ui/pull/1325, thanks good catch

PR #1321, opened 2019-12-19
feat: Allow `ref` in text and titles

edas:
This is needed in cozy-notes for `<MainTitle />`

### DIFF
--- a/react/Text/index.spec.jsx
+++ b/react/Text/index.spec.jsx
@@ -1,0 +1,35 @@
+import React from 'react'
+import {
+  Text,
+  MainTitle,
+  Title,
+  SubTitle,
+  Bold,
+  Uppercase,
+  Caption,
+  ErrorMessage
+} from './index'
+
+describe('Text/*', () => {
+  it('forwards ref to the HTML tag', () => {
+    const components = [
+      Text,
+      MainTitle,
+      Title,
+      SubTitle,
+      Bold,
+      Uppercase,
+      Caption,
+      ErrorMessage
+    ]
+    components.forEach(Component => {
+      const ref = React.createRef()
+      mount(
+        <div>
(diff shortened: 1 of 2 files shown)


edas:
Failing tests doesn't seem related to my changes. Any ideas on why ?

ptbrowne:
I haven't understood the root cause, but the truth seems to come from Travis, the snapshot for InfosCarrousel has only 1 slide where it should have two. https://github.com/cozy/cozy-ui/pull/1302/files#diff-9298567b32c47c61874c3eea102bcd60R5011

ptbrowne:
Can you add a test for the forwardRef ?

edas:
> Do you think it would be a good idea to add a test for the forward ref ?

will do

edas:
Done for the tests. Is it ok for you ?

react/Text/index.spec.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+import { Text, MainTitle, Title, SubTitle, Bold, Uppercase, Caption, ErrorMessage } from './index'
+
+describe('Text/*', () => {
+  it('forwards ref to the HTML tag', () => {
+    const components = [ Text, MainTitle, Title, SubTitle, Bold, Uppercase, Caption, ErrorMessage ]
```
ptbrowne: Nit: Putting the for loop just outside the `it` would allow for more precise test failures.

```
const components = [ Text, MainTitle, Title, SubTitle, Bold, Uppercase, Caption, ErrorMessage ]

for (const component of components) {
  test('${component.name} forwards ref to the HTML tag', () => {
    ...
  })
}
```

ptbrowne:
Some linting to do ;)

PR #1323, opened 2019-12-20
feat: Size of button icons is dependent of the button size

ptbrowne:
Previously icons inside buttons were not dependent on the size of the button, which made small / tiny buttons with icons look weird.

See https://ptbrowne.github.io/cozy-ui/react/#/Button > Subtle Buttons (deploying...)


ptbrowne:
Merging since the failing test has been cfixed by #1324

PR #1326, opened 2019-12-20
Contacts list realtime

drazik:
This adds realtime to the contacts list. There are two features:

* If the contacts app is not installed, then in the ContactsListModal, the « add a contact » button links to the contacts' store install page. When the app has been installed, we update the button so it links to the contact page instead of the store
* When the user clicks on « add a contact », and creates or updates a contact in the contacts app, we update the contacts list, so when he comes back to the original app, he sees the new or updated contacts.

There is a thing that bothers me, it's that by adding cozy-realtime and using it in the ContactsListModal, which is exported by the main entry point (index.js), it will require apps to install cozy-realtime. We already have the problem with cozy-client sometimes. I don't know how to fix this without removing the export of this component in the main entry point. Do you have any ideas?

Also, snapshots are failing and I don't really know why :thinking: 

Edit : the app can use a cozy-realtime plugin in its client, see https://mattermost.cozycloud.cc/cozycloud/pl/41c6hhyhob8wdpybpri8k1416r

### DIFF
--- a/react/ContactsListModal/AddContactButton.jsx
+++ b/react/ContactsListModal/AddContactButton.jsx
@@ -1,22 +1,66 @@
-import React from 'react'
-import { withClient } from 'cozy-client'
-import AppLinker, { generateWebLink } from '../AppLinker'
+import React, { useEffect } from 'react'
+import { withClient, models, queryConnect } from 'cozy-client'
+import AppLinker from '../AppLinker'
 import { ButtonLink } from '../Button'
+import compose from 'lodash/flowRight'
 
 const DumbAddContactButton = props => {
-  const { client, ...rest } = props
+  const { client, apps, ...rest } = props
 
-  const cozyURL = new URL(client.getStackClient().uri)
-  const { cozySubdomainType } = client.getInstanceOptions()
-  const contactsAppSlug = 'contacts'
-  const contactsAppHref = generateWebLink({
-    cozyUrl: cozyURL.origin,
-    slug: contactsAppSlug,
-    subDomainType: cozySubdomainType
-  })
+  const wantedApp = { slug: 'contacts' }
+  const installedApp = models.applications.isInstalled(apps.data, wantedApp)
+
+  let href
+
+  if (installedApp) {
+    href = models.applications.getUrl(installedApp)
+  } else {
+    href = models.applications.getStoreInstallationURL(apps.data, wantedApp)
+  }
+
+  useEffect(() => {
+    const refreshApps = () => {
+      apps.fetch()
+    }
+
+    const subscribeRealtime = async () => {
+      try {
+        await client.plugins.realtime.subscribe(
+          'created',
+          'io.cozy.apps',
+          refreshApps
+        )
+      } catch (err) {
+        console.error(err)
+        console.warning(
+          'Impossible to subscribe to io.cozy.apps creation in realtime. Your app should have io.cozy.apps permission and your client should have realtime initialized.'
+        )
+      }
+    }
+
+    subscribeRealtime()
+
+    return async () => {
+      try {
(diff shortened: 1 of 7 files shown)


react/ContactsListModal/AddContactButton.jsx
```diff
@@ -1,22 +1,50 @@
...
-  const contactsAppHref = generateWebLink({
-    cozyUrl: cozyURL.origin,
-    slug: contactsAppSlug,
-    subDomainType: cozySubdomainType
+  const { client, apps, ...rest } = props
+
+  const wantedApp = { slug: 'contacts' }
+  const installedApp = models.applications.isInstalled(apps.data, {
+    slug: 'contacts'
   })
```
Crash--:
```suggestion
const installedApp = models.applications.isInstalled(apps.data, wantedApp)
```

react/ContactsListModal/AddContactButton.jsx
```diff
@@ -1,22 +1,50 @@
...
+  let href
+
+  if (installedApp) {
+    const cozyURL = new URL(client.getStackClient().uri)
+    const { cozySubdomainType } = client.getInstanceOptions()
+    href = generateWebLink({
+      cozyUrl: cozyURL.origin,
+      slug: installedApp.attributes.slug,
+      subDomainType: cozySubdomainType
+    })
```
Crash--: Is not the responsability of this PR but https://github.com/cozy/cozy-ui/issues/1327

react/ContactsListModal/AddContactButton.jsx
```diff
@@ -1,22 +1,50 @@
...
+    href = models.applications.getStoreInstallationURL(apps.data, wantedApp)
+  }
+
+  useEffect(() => {
+    const realtime = new CozyRealtime({ client })
+    const refreshApps = () => {
+      apps.fetch()
+    }
+
+    realtime.subscribe('created', 'io.cozy.apps', refreshApps)
```
Crash--: We should try / await / catch this subscribe and put a warning to tell the dev that to use this feature the app should have the rights on `io.cozy.apps`

react/ContactsListModal/AddContactButton.jsx
Crash--: Since you have the `installedApp` you can simply use `models.applications.getUrl(installedApp)` since we'll have the `app.links.related` from the stack ;)

react/ContactsListModal/AddContactButton.jsx
drazik: :+1:

react/ContactsListModal/AddContactButton.jsx
drazik: :+1: thankss

react/ContactsListModal/AddContactButton.jsx
drazik: :+1:

react/ContactsListModal/index.jsx
```diff
@@ -61,6 +64,77 @@ const ContactsListModal = props => {
...
+  const filteredContacts = filterContacts(contacts.data)
+
+  useEffect(() => {
+    const refreshContacts = () => {
+      contacts.fetch()
+    }
+
+    const subscribeRealtime = async () => {
+      try {
+        await client.plugins.realtime.subscribe(
```
Crash--: Should we test if client.plugins.realtime exists and if no tell the dev to register the plugin to get a full featured component?

drazik: I added it, but thought it was becoming a lot of logic inside the component. So I extracted all of this in a hook. See https://github.com/cozy/cozy-ui/pull/1326/commits/fab16305aac90a08418866b2481764f882179f15

PR #1336, opened 2020-01-03
feat(ContactsListModal): Double realtime with mobile app resume…

drazik:
When using a mobile app, when the user wants to add a contact, the app
will be put in the background while the user is in a browser to add the
contact. During this time, the realtime events are not handled by the
app. To make sure the user sees fresh data, we have to also subscribe to
the cordova's resume event to re-fetch it when the app is put to
foreground again.


react/hooks/useMobileAppResume.js
```diff
@@ -0,0 +1,18 @@
...
+      document.addEventListener('resume', cb, false)
+    }
+
+    return () => {
+      if (isMobileApp()) {
+        document.removeEventListener('resume', cb, false)
+      }
+    }
+  }, [cb])
+}
```
ptbrowne: Nit: I am not sure it is necessary to check whether it is a mobile app before subscribing to the event. AMHA subscribing to the resume event is fairly cheap and does not consume a lot of resource.

Nit: A useEventListener hook could be used here to have more flexibility in the future.

drazik: I changed the useMobileAppResume hook to useEventListener. This addresses both nits: we don't check the platform, and we have the flexibility.

PR #1341, opened 2020-01-08
Fix/app linker don't crash

Crash--:
We encountered the issue on Notes. If you mount an `<AppLinker>` with an empty `href`, `generateUniversalLink` throws an error but we don't catch it and we finished by having a blank screen. 

This PR fix the behavior by catching the error and logging it with an console.error(). It's better to have a link that doesn't work instead of the all app.

### DIFF
--- a/react/AppLinker/index.jsx
+++ b/react/AppLinker/index.jsx
@@ -81,7 +81,7 @@ export class AppLinker extends React.Component {
           href = generateUniversalLink({ slug, nativePath, fallbackUrl: href })
         } catch (err) {
           console.error(err)
-          href = ''
+          href = '#'
         }
       }
     }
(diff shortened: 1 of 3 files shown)


react/AppLinker/index.jsx
```diff
@@ -75,7 +75,14 @@ export class AppLinker extends React.Component {
...
         onClick = AppLinker.openNativeFromWeb.bind(this, props)
       } else {
-        href = generateUniversalLink({ slug, nativePath, fallbackUrl: href })
+        //Since generateUniversalLink can rise an error, let's catch it to not crash
+        //all the page.
+        try {
+          href = generateUniversalLink({ slug, nativePath, fallbackUrl: href })
+        } catch (err) {
+          console.error(err)
+          href = ''
```
ptbrowne:
```suggestion
          href = '#'
```

?

Crash--: Yep better. Thanks :)

PR #1342, opened 2020-01-08
feat: Restore displayNames for text components

edas:
One recent commit made Text components loose their display name. This change fixes that.


edas:
Superseded by a change from Patrick

PR #1347, opened 2020-01-13
Responsive ActionMenu

y-lohse:
At the moment we have 2 types of menus:

- [`Menu`](https://docs.cozy.io/cozy-ui/react/#menu), suitable for desktop
- [`ActionMenu`](https://docs.cozy.io/cozy-ui/react/#actionmenu) suitable for mobile

A common pattern we have is to show the same menu as a `Menu` on desktop and an `ActionMenu` on mobile. But the two component have quite different APIs, making it more difficult to implement our menus than necessary.

This PR solves the problem by introducing a new prop for `ActionMenu` which changes the menu style from mobile to desktop. This means `ActionMenu` can be used for mobile like before, but also for desktop as a replacement for `Menu` (with a better API), and it can be changed from mobile to desktop automatically by hooking it up with `withBreakpoints` for example. [The component demo is here](https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu).

To accomplish this I split the old `ActionMenu` component in two, so we now have a new [`BottomDrawer` component](https://yannick-lohse.fr/cozy-ui/react/#!/BottomDrawer) as well.


react/ActionMenu/index.jsx
```diff
@@ -1,166 +1,43 @@
...
 import styles from './styles.styl'
-import Overlay from '../Overlay'
-import once from 'lodash/once'
 import { Media, Bd, Img } from '../Media'
+import BottomDrawer from '../BottomDrawer'
+import ClickAwayListener from '@material-ui/core/ClickAwayListener'
+
+const ActionMenuWrapper = ({ inline, onClose, children }) =>
+  inline ? (
+    <ClickAwayListener onClickAway={onClose}>{children}</ClickAwayListener>
```
y-lohse: I'm not 100% sure about this. It works as intended, but also makes this component dependent on material-ui. I think it's ok, but we ca achieve the same thing with a transparent `Overlay` component if we'd rather not have the dependency.

react/BottomDrawer/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+  close = () => {
+    this.setState({ closing: true })
+    this.props.onClose && this.props.onClose()
+  }
+
+  render() {
+    const { children } = this.props
+    const { closing } = this.state
+    this.menuNode = React.createRef()
+    this.wrapperNode = React.createRef()
```
y-lohse: This whole component is copy and pasted from `ActionMenu` with the exception of the `ref`s — they are now created with `React.createRef` instead of the old `ref` prop.

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -18,6 +18,12 @@ exports[`ActionMenu should render examples: ActionMenu 3`] = `
 </div>"
 `;
 
+exports[`ActionMenu should render examples: ActionMenu 4`] = `
+"<div>
+  <div><button>Show action menu</button></div>
```
Crash--: you should have an initialOpen or something similar when testing ;)

react/ActionMenu/index.jsx
Crash--: IMO not a big deal to have a dep to MUI. It's now almost mandatory anyway

y-lohse:
> can you add a link to a styleguidist plz?

They are in the first message 😁 but here we go:

- https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu
- https://yannick-lohse.fr/cozy-ui/react/#!/BottomDrawer

ptbrowne:
Does this mean that using Menu is deprecated now ?

y-lohse:
I updated the component, docs and redeployed the examples. Could you take a look and let me know if it looks ok? If yes I will deprecate the `Menu` component and handle the add infos about the migration.

react/ActionMenu/index.jsx
```diff
@@ -1,160 +1,20 @@
...
-    )
-  }
-}
+const ActionMenu = ({ children, className, onClose }) => (
+  <div className={className}>
+    <BottomDrawer onClose={onClose}>
+      <div className={styles['c-actionmenu']}>{children}</div>
+    </BottomDrawer>
+  </div>
+)
```
ptbrowne: It seems we have no control on the inner actionMenu className.

Why is the wrapping div necessary ? Could the className be put on the BottomDrawer for the structure to be simpler ?

react/BottomDrawer/index.jsx
```diff
@@ -0,0 +1,151 @@
...
+    return (
+      <div ref={this.wrapperNode}>
+        <Overlay
```
ptbrowne: Nit: I am wondering if we could simplify the structure and have the ref directly on the Overlay. It would necessitate a forwardRef on the Overlay though.

react/BottomDrawer/styles.styl
```diff
@@ -0,0 +1,15 @@
...
+    transition transform .1s ease-out
+
+.BottomDrawer-content
+    z-index        $file-action-menu
+    position       fixed
+    bottom         rem(9)
+    bottom         env(safe-area-inset-bottom) // @stylint ignore
+    left           0
+    width          'calc(100% - %s)' % rem(16)
+    margin         0 rem(8)
```
Crash--: Nit: You can use spacing_values :-°

y-lohse: In my defense, it was a copy and paste :)

react/ActionMenu/index.jsx
y-lohse: You may have seen that in the more recent code, the inner menu can be wrapped either in `BottomDrawer` or `ClockAwayListener`. We could pass the props to `BottomDrawer`, but `ClickAwayListener` absolutely needs a wrapper. I think injecting the classname in 2 different ways would make the structure more complicated.

react/BottomDrawer/styles.styl
ptbrowne: :/ We should always take a look at the code we copy and paste

y-lohse:
I've now added the deprecation warning on Menu and a migration guide [here](https://yannick-lohse.fr/cozy-ui/react/#!/Menu). Unless there are issues with any of these I plan to merge this PR soon.

I realized the `ActionMenu` doesn't support left/right positions as `Menu` does, I will add it in a separate PR.

Crash--:
Menu also supports `Popover` and I don't think `ActionMenu` supports it yet. 

Since we're deprecating `Menu` here , we should provide a 100% working alternative.

y-lohse:
> Menu also supports Popover and I don't think ActionMenu supports it yet.

Yep, will do at the same time as I add support for something like `position`. Maybe they can even be combined, not sure yet.

PR #1348, opened 2020-01-13
fix: Adjust Infos UI

y-lohse:
This PR makes some small UI tweaks for the Infos component, as seen with @joel-costa .

My preview branch is busy with #1347 but I think we can do without it here, it's just changing some spacing values.

### DIFF
--- a/react/InfosCarrousel/styles.styl
+++ b/react/InfosCarrousel/styles.styl
@@ -1,4 +1,5 @@
 @require 'settings/palette'
+@require 'settings/breakpoints'
 
 .InfosCarrousel
     position relative
@@ -12,16 +13,20 @@
 
 .InfosCarrousel-navigation
     position absolute
-    bottom 0
-    right 0
+    bottom 10px
+    right 8px
     display flex
     align-items center
+
+    +mobile()
+        bottom 4px
+        right 0
 
 .InfosCarrousel-arrow
     background none
     border 0
     min-width 32px
-    padding 8px 16px
+    padding 16px
 
     &:disabled
         opacity .24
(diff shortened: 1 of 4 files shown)


react/InfosCarrousel/styles.styl
```diff
@@ -12,16 +13,20 @@
 
 .InfosCarrousel-navigation
     position absolute
-    bottom 0
-    right 0
+    bottom 10px
+    right 8px
```
drazik: This can be replaced by `spacing_values.xs`

react/InfosCarrousel/styles.styl
```diff
@@ -12,16 +13,20 @@
...
     align-items center
 
+    +mobile()
+        bottom 4px
+        right 0
+
 .InfosCarrousel-arrow
     background none
     border 0
     min-width 32px
```
drazik: This can be replaced by `spacing_values.xl`

react/InfosCarrousel/styles.styl
```diff
@@ -12,16 +13,20 @@
...
-    padding 8px 16px
+    padding 16px
```
drazik: This can be replaced by `spacing_values.m`

react/InfosCarrousel/styles.styl
```diff
@@ -12,16 +13,20 @@
 
 .InfosCarrousel-navigation
     position absolute
-    bottom 0
-    right 0
+    bottom 10px
```
drazik: Can we use `rem(10)` here?

react/InfosCarrousel/styles.styl
```diff
@@ -12,16 +13,20 @@
...
+    right 8px
     display flex
     align-items center
 
+    +mobile()
+        bottom 4px
```
drazik: Can we use `rem(4)` here?

react/InfosCarrousel/styles.styl
ptbrowne: Are we sure that a min-width should be implemented with a spacing value ?

PR #1354, opened 2020-01-16
fix(Modal): Use overflow auto

drazik:
The modal wrapper was using `overflow-y: scroll`. The problem is that
the `scroll` value makes scrollbar appear even if it's disabled. Using
the `auto` value fixes it.

Before 
image

After
image


ptbrowne:
overflow-y scroll was introduced in https://github.com/cozy/cozy-ui/pull/117. I don't think it was necessary to have overflow-y scroll and think auto is ok for the purpose of this PR. LGTM.

PR #1355, opened 2020-01-17
InfosCarrousel improvements

### LINKED ISSUE #1352: InfosCarrousel improvements
Infos arrows do not have cursor pointer.

ptbrowne:
Fix #1352

- [x] Infos arrows do not have cursor pointer.
- [x] When there is only one Info, we don't need to display the arrows
- [x] When swipping between errors, the arrows are not updated
- [x] Added tests

https://ptbrowne.github.io/cozy-ui/react/#/InfosCarrousel


react/InfosCarrousel/index.jsx
```diff
@@ -18,6 +18,17 @@ const InfosCarrousel = ({ children, theme, className, swipeableProps }) => {
...
+  const onChangeIndex = useMemo(
+    () =>
+      function(index) {
+        const { onChangeIndex: onChangeIndexProp } = swipeableProps
+        if (onChangeIndexProp) {
+          onChangeIndexProp(index)
+        }
+        setIndex(index)
+      }
+  )
```
drazik: The [`useMemo` docs](https://reactjs.org/docs/hooks-reference.html#usememo) says:

> If no array is provided, a new value will be computed on every render.

It seems here the function will be redefined on every render, so useMemo may be useless.

Since we are more and more using hooks, I think we should add [eslint-plugin-react-hook](https://www.npmjs.com/package/eslint-plugin-react-hooks#installation) to our eslint config, so the deps of our hooks are automatically checked. See https://github.com/cozy/cozy-libs/issues/916

react/InfosCarrousel/styles.styl
```diff
@@ -26,6 +26,7 @@
 .InfosCarrousel-arrow
     background none
     border 0
+    cursor pointer
```
drazik: Don't we want to reset it to default cursor on disabled arrows?

react/InfosCarrousel/index.jsx
y-lohse: Agree with the above, but I believe it's a situation for `useCallback` here?

react/InfosCarrousel/styles.styl
```diff
@@ -14,24 +14,26 @@
 
 .InfosCarrousel-navigation
     position absolute
-    bottom rem(10)
-    right spacing_values.xs
+    bottom (spacing_values.xs)
```
y-lohse: The `rem(10)` was intended, but perhaps it needs a comment or a variable name - it's there to be visually aligned with buttons inside the `Infos` component.

react/InfosCarrousel/styles.styl
```diff
@@ -14,24 +14,26 @@
...
+    right (spacing_values.xs)
     display flex
     align-items center
 
     +mobile()
         bottom rem(4)
-        right 0
+        right rem(4)
```
y-lohse: I don't think we want the extra offset on mobile?

react/InfosCarrousel/index.jsx
ptbrowne: Ah yes, useCallback is better here as we do not have to have a `() => function () {}` but only `function () {}`. Thanks !

react/InfosCarrousel/styles.styl
ptbrowne: Seen in Zoom ;)

react/InfosCarrousel/styles.styl
ptbrowne: Seen in Zoom

ptbrowne:
J'ai rajouté un exemple pour IconButton et mis a jour la preview sur mon fork ;) https://ptbrowne.github.io/cozy-ui/react/#/InfosCarrousel
https://ptbrowne.github.io/cozy-ui/react/#/IconButton

Issue #1206, opened 2019-10-17
Argos / Component viewPort mobile


Crash--:
Some component need to be tested on mobile. ATM argos only set the viewport to desktop resolution.

This issue is here to discuss how we can implement automatic screenshot on mobile viewport for a few components.  (for instance https://www.argos-ci.com/cozy/cozy-ui/builds/55995) 

We currently go to the page : https://github.com/cozy/cozy-ui/blob/master/scripts/screenshots.js#L30 and take the screenshot 

If we put an information in this page saying "take a mobile screenshot", we should be able to read it from `page` and set the viewport accordingly. 

We can have for instance:
Readme.md : 
``
#### MobileOnly
```jsx
...
```
```

OR we can have a argos/config/ComponentName.json to tell which viewport we want to use for each component?

ptbrowne:
Fixed in https://github.com/cozy/cozy-ui/pull/1356

PR #1356, opened 2020-01-17
Screenshots in mobile viewport

ptbrowne:
Take screenshots in mobile viewport additionally to the desktop viewport.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ script:
   - yarn build:css:all
   - yarn build:doc:react
   - yarn screenshots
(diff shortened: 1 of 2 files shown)


Crash--:
https://github.com/cozy/cozy-ui/issues/1206 for posterity

ptbrowne:
Ah thanks, I had forgotten about the possibility of doing this at component level. I think it's simpler like this. And since we do not have to add mobile viewport request at component level, we detect unknown problems like the one with the tabs.

Crash--:
Yeah I agree with you! It's simpler this way. My comment was just a reminder to close the issue once this PR is merged :D

PR #1359, opened 2020-01-20
feat: Support ActionMenu placement and scroll

y-lohse:
The `Menu` supported `position` and `popover` props. this PR brings similar features for `ActionMenu`, carried by Popover and under the names `placement` and `preventOverflow`.

[Here is the preview.](https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu) As you may see, the menus don't move along when you scroll the page. We had the same problem with the `<Menu popover />`, unfortunately I couldn't find a workaround. It works as expected outside of styleguidist though.

____

The PR also fixes the margins on the `BottomDrawer` component, which affects the `ActionMenu`. [The preview is here](https://yannick-lohse.fr/cozy-ui/react/#!/BottomDrawer).


react/Menu/Readme.md
```diff
@@ -94,3 +94,12 @@ const showItem = item => {
...
 ```
+
+#### `position` and `popover`
+
+`ActionMenu` relies on the Material-UI (Popover)[https://v3.material-ui.com/utils/popper/] component to handle both `position` and `popover`.
+
+- `position='left'` is now `placement='bottom-start'`, and is still the default behavior.
+- `position='right'` is now `placement='bottom-end'`.
+- All other Popover.js placement options are also supported
+- The Menu `popover` prop is now called `preventOverflow`
```
Crash--: It should be considered as a Breaking changes (and have this description attached to the commit)

Crash--: My bad!

PR #1360, opened 2020-01-20
feat: Adds useConfirmExit React hook

edas:
This hook was needed for cozy-notes. It allows to hook the onBeforeUnloadevent to warn about unsaved changes.

### DIFF
--- a/react/hooks/useConfirmExit.js
+++ b/react/hooks/useConfirmExit.js
@@ -1,0 +1,30 @@
+import { useCallback, useRef } from 'react'
+import useEventListener from './useEventListener'
+
+/**
+ * When provided a message, will warn the user before exiting the page
+ *
+ * Warning: The confirmation message is required but will usually
+ * be replaced by an internal message by the browser, and never displayed.
+ *
+ * A falsy error message would deactivate to confirmation popup.
+ *
+ * @param {string|null} message - Confirmation message
+ * @param {function} callback - will be executed before returning
+ */
+export default function useConfirmExit(message, callback) {
+  // Using a ref in order to have an event listener that does not
+  // need to be deregistered, recreated and registered again at each
+  // message or callback change. If not, the lag introduced by the
+  // useEffect inside useEventListener may create wrong behaviours
+  // for fast changing calls to useConfirmExit.
+  const state = useRef()
+  state.current = { message, callback }
+  const beforeunload = useCallback(event => {
+    state.current.callback && state.current.callback()
+    const returnValue = state.current.message || null
+    if (returnValue) event.returnValue = returnValue
+    return returnValue
+  }, [])
+  useEventListener(window, 'beforeunload', beforeunload)
+}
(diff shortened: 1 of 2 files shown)


react/hooks/useConfirmExit.js
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * A falsy error message would deactivate to confirmation popup.
+ *
+ * @param {string|null} message - Confirmation message
+ * @param {function} callback - will be executed before returning
+ */
+export default function useConfirmExit(message, callback) {
+  // Using a ref in order to have an event listener that does not
+  // need to be deregistered, recreated and registered again at each
+  // message or callback change. If not, the lag introduced by the
```
ptbrowne: I don't really understand why we we have use a ref here ? How does it work ?

edas: Without the ref, the `beforeunload` event handler would change each time you change the message or the callback. Then each time it changes, it would call the `useEventListener` that would deregister the previous event handler and register the new one.

In cozy-notes, we may change the parameters very often (on a box with low latency, it may be every 20ms). It would trigger useless attachEvent and removeEvent that many times.

Furthermore, useEffet inside useEventListener is not immediate. It should be very quick but when the state change so often, I had occurence where the confirmation did not occur  because the corresponding effect still hasn't run (and an old eventhandler with obsolete parameters was still attached).

Using a stable event handler and a reference allows us to always have an handler that will use the most recent values

react/hooks/useConfirmExit.js
```diff
@@ -0,0 +1,30 @@
+import { useCallback, useRef } from 'react'
+import useEventListener from 'cozy-ui/transpiled/react/hooks/useEventListener'
```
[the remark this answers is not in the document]
edas: J'invoque @Crash-- qui me demandait de faire le contraire pour permettre les surcharges B2B

react/hooks/useConfirmExit.js
ptbrowne: After reading in more detail the useRef doc, it became clear.

react/hooks/useConfirmExit.js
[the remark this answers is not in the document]
y-lohse: Dans les apps on met un chemin absolu pour la surcharge, mais pas dans cozy-ui. Et il faut éviter d'importer un composant depuis `transpiled` dans la partie non-transpilée de cozy-ui.

react/hooks/useConfirmExit.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+describe('useConfirmExit', () => {
+  it('should subscribe to window.onbeforeunload at mount', async () => {
+    const cb = jest.fn()
+    const { unmount } = renderHook(() => useConfirmExit('message', cb))
+
+    triggerBeforeUnload()
+    expect(cb).toHaveBeenCalledTimes(1)
+
+    unmount()
```
ptbrowne: C'st pas fait automatiquement ca ? cc drazik

react/hooks/useConfirmExit.js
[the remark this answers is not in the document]
edas: Erreur de ma part, je corrige

PR #1361, opened 2020-01-21
Merge pull request #1361 from cozy/page-par-section

ptbrowne:
Since we start to have a lot of examples, performance starts
to degrade when loading the styleguidist. Having only one page
per section could help.

This is done to be able to add more complex examples ("organisms" in atomic design parlance) that would
showcase for example how we want to render a settings page
(discussed with @joel-costa).

### DIFF
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -3,6 +3,7 @@ const webpackMerge = require('webpack-merge')
 
 module.exports = {
   title: 'Cozy UI React components',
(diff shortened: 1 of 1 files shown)


y-lohse:
Ok for me, but does it have an impact on screenshots?

ptbrowne:
Because in the perPageSection mode, we are focused on the first section ("Basics") when opening the page. Since we are focused, the focus buttons are toggled.

PR #1364, opened 2020-01-22
Hook for breakpoints

ptbrowne:
- useBreakpoints to provide the same breakpoints predicate as withBreakpoints
- Perf change : only one window resize handler (app needs to be wrapped in BreakpointsProvider)
- withBreakpoints does not change, retrocompat is assured

We should be able to code a codemod to do the transition from withBreakpoints to useBreakpoints, but this is not in this PR since there is no deprecation on withBreakpoints for now.


react/hooks/useBreakpoints/Readme.md
```diff
@@ -0,0 +1,62 @@
+You can use the `useBreakpoints` hook when you need access to whether
```
y-lohse:
```suggestion
You can use the `useBreakpoints` hook to know whether
```

ptbrowne:
Yes, this was intentional since we do not use this feature much.

PR #1366, opened 2020-01-23
feat: Adds a user-space modal for useConfirmExit

edas:
This is a BREAKING CHANGE but the hook is only used by me on cozy-notes and I will update cozy-notes accordingly

The hook now returns a `requestToLeave` function and a `exitConfirmationModal` React object. When called with an URL or a function, requestToLeave will display a confirmation modal in `exitConfirmationModal` before going to this url or executing the function.

This allows to have a confirmation on exit when we leave the current view (in addition to when we leave the window or change the domain, which was previously implemented)


react/locales/en.json
```diff
@@ -0,0 +1,14 @@
...
+        "useConfirmExit": {
+          "back": "Back",
+          "leave": "Leave",
+          "title": "Do you want to leave?",
+          "message": "Some changes are not saved yet. Do you really want to leave and loose these changes?"
+        }
+      }
+    }
+  }
+}
```
Crash--: ATM we are storing the locales in the component's folder (ex: react/viewer/locales, react/AppSection/locales). Personally I find it easier to maintain

react/locales/en.json
```diff
@@ -0,0 +1,14 @@
+{
+  "UI": {
+    "React": {
+      "hooks": {
+        "useConfirmExit": {
+          "back": "Back",
+          "leave": "Leave",
+          "title": "Do you want to leave?",
+          "message": "Some changes are not saved yet. Do you really want to leave and loose these changes?"
```
Crash--:
```suggestion
          "message": "Some changes are not saved yet. Do you really want to leave and lose these changes?"
```

react/locales/en.json
Crash--: @edas Does it miss the commit to move the files to the right folder? Or do you want to keep this structure?

PR #1367, opened 2020-01-27
feat: Adds useRerender

edas:
This changes adds a hook `useRerender` which returns
a simple function that cause the component to be
rerendered when called.

### DIFF
--- a/react/hooks/useRerender.js
+++ b/react/hooks/useRerender.js
@@ -1,0 +1,12 @@
+import { useState, useCallback } from 'react'
+
+/**
+ * Very simple React hook to trigger a rerender when needed
+ *
+ * @returns {function} rerender the hook & calling component when called
+ */
+export default function useRerender() {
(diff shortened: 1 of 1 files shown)


ptbrowne:
I also think it is an anti pattern. Could you point to an example of code using this?

edas:
I had this in my last two previous hooks where I had to use references instead of states. Doing so, I missed a way to forcea rerender and used the same trick (a useless state that is only used to trigger a rerender). This PR was to mutualize the trick

edas:
Example in : https://github.com/cozy/cozy-notes/blob/master/src/hooks/useCollaborationState.js#L36

Is is using a ref instead of a state because it's changing very frequently and if we use the state a basis for our values, we may have wrong behaviours.

nono:
What is the status on this PR?

edas:
We choosed to discard it. We'll see ifwe really need it later

PR #1368, opened 2020-01-27
feat: Distribute es5 files

ptbrowne:
- "transpiled" contains files with import directive, useful for tree shaking
- "es5" directory contains files without import directories, this way we do not
  have to babel those files during tests


y-lohse:
Nothing wrong with the changes, but how do you plan to use the new es5 output? Import them instead of the `transpiled` files in our other repos test suites?

ptbrowne:
> Import them instead of the transpiled files in our other repos test suites?

Yes.

package.json
```diff
@@ -41,17 +41,19 @@
...
     "semantic-release": "semantic-release",
     "sprite": "scripts/make-icon-sprite.sh",
     "jest": "yarn test:jest",
     "screenshots": "node scripts/screenshots.js --screenshot-dir ./screenshots --styleguide-dir ./build/react",
     "test": "yarn test:jest",
     "test:jest": "jest",
-    "transpile:react": "env BABEL_ENV=transpilation babel react/ --out-dir transpiled/react --verbose",
-    "transpile:helpers": "babel helpers/ --out-dir transpiled/helpers --verbose",
-    "transpile": "yarn transpile:react && yarn transpile:helpers",
+    "transpile:react": "babel react/ --out-dir $BABEL_OUTPUT_DIR/react --verbose",
```
Crash--: @ptbrowne Do we need now to declare a $BABEL_OUTPUT_DIR var in our env?

ptbrowne: No, since the `transpile` method does that for you.

Crash--: Oh OK. I'm not used to `transpile`, I mainly use `transpile:react` directly to have less work to do when watching. Let's change my habit then

ptbrowne: See https://github.com/cozy/cozy-ui/pull/1368#discussion_r372209679

PR #1372, opened 2020-01-29
Feat/confirm exit labels

edas:
Change to the exit confirmation modal:
-  Translations
- The confirmation button is now a "danger"

### DIFF
--- a/react/hooks/useConfirmExit/index.js
+++ b/react/hooks/useConfirmExit/index.js
@@ -22,8 +22,8 @@ function ConfirmModal({
   message,
   onConfirm,
   onCancel,
-  cancel,
-  confirm
+  cancelLabel,
+  confirmLabel
 }) {
   return (
     <Modal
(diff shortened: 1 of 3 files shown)


react/hooks/useConfirmExit/index.js
```diff
@@ -142,6 +152,8 @@ export default function useConfirmExit({
       title={title}
       onCancel={onCloseModalRequest}
       onConfirm={onConfirm}
+      confirm={leave}
+      cancel={cancel}
```
Crash--: `cancel` or `confirm` are too generic and we don't know that they're string and used for primaryText / secondaryText. 

Instead we should have: 
`confirmLabel` and `cancelLabel`. 

but I'm not sure about creating a new API in fact. Why don't we use the existing Modal API with primaryText and secondaryText?

Crash--: and primaryAction / dismissAction. Since we have `Modal` in the component name, I expect the same API as our modal.

WDYT?

edas: I would prefer not.
In this flavour, they are not generic primary and secondary buttons. They have a dedicated functional role.

* We may put cancel as primary and confirm as secondary in the future (we hesitated yesterday)
* We do not want to let people think it can change the primary  and secondary role themselves because  it would mess with default labels.

edas: (I am changing the names for the labels)

edas: :+1: for confirmLabel and cancelLabel, I made the change

react/hooks/useConfirmExit/index.js
```diff
@@ -16,27 +16,38 @@ import fr from './locales/fr.json'
...
 }
-ConfirmModal.PropTypes = {
+ConfirmModal.propTypes = {
   message: PropTypes.string,
   title: PropTypes.string,
   onConfirm: PropTypes.func.isRequired,
-  onCancel: PropTypes.func.isRequired
+  onCancel: PropTypes.func.isRequired,
+  cancelLabel: PropTypes.func.string,
+  confirmLabel: PropTypes.func.string
```
Crash--:
```suggestion
  confirmLabel: PropTypes.string
```

react/hooks/useConfirmExit/index.js
```diff
@@ -16,27 +16,38 @@ import fr from './locales/fr.json'
...
   )
 }
-ConfirmModal.PropTypes = {
+ConfirmModal.propTypes = {
   message: PropTypes.string,
   title: PropTypes.string,
   onConfirm: PropTypes.func.isRequired,
-  onCancel: PropTypes.func.isRequired
+  onCancel: PropTypes.func.isRequired,
+  cancelLabel: PropTypes.func.string,
```
Crash--:
```suggestion
  cancelLabel: PropTypes.string,
```

PR #1375, opened 2020-01-29
feat: Adds useInterval

edas:
Allows to rerender the calling component by a given interval

### DIFF
--- a/react/hooks/usePeriodicRender.js
+++ b/react/hooks/usePeriodicRender.js
@@ -1,0 +1,23 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Force a new render each `duration` ms
+ *
+ * @param {number} duration - in millisecond
+ * @returns {Date} last trigger date for a rerender
+ */
+export default function usePeriodicRender(duration) {
+  const [lastExecuted, setLastExecuted] = useState(undefined)
+  useEffect(() => {
+    if (duration) {
+      const interval = window.setInterval(() => {
+        setLastExecuted(new Date())
+      }, duration)
+      return () => {
+        window.clearInterval(interval)
+      }
+    }
+  }, [duration])
+
+  return lastExecuted
+}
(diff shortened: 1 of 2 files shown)


ptbrowne:
No tests :) ?

edas:
> No tests :) ?

Fixed, thank  you

react/hooks/useInterval.js
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from 'react'
+
+/**
+ * Force a new render each `duration` ms
+ *
+ * @param {number} duration - in millisecond
+ * @returns {Date} last trigger date for a rerender
+ */
+export default function useInterval(duration) {
```
y-lohse: Have you considered a name like `setIntervalRerender` or something similar? To make it explicit that the hook is about re-rendering and not so much about intervals.

react/hooks/useInterval.spec.js
```diff
@@ -0,0 +1,68 @@
+import useInterval from './useInterval'
+import { renderHook } from '@testing-library/react-hooks'
+
+async function wait(duration) {
+  console.log('setTimeout', duration)
```
y-lohse: We probably don't want the extra log here :)

react/hooks/useInterval.spec.js
```diff
@@ -0,0 +1,68 @@
...
+  return new Promise(resolve => {
+    window.setTimeout(() => {
```
y-lohse: Could we use [timer mocks here](https://jestjs.io/docs/en/timer-mocks.html)?

react/hooks/useInterval.js
ptbrowne: usePeriodicRender

react/hooks/useInterval.spec.js
edas: Vu ensemble, pas adapté pour ce cas

PR #1376, opened 2020-01-30
fix: InfosCarrousel handles removing children

y-lohse:
Fixes a bug with InfosCarrousel: when we navigate to the last slide, and remove a child, the internal `index` points to a slide that no longer exists. We now force the index back one slide when that happens.

This component can still "break" if it ends up with no children at all, but I don't think that the component itself should guard against that..?

### DIFF
--- a/react/InfosCarrousel/index.jsx
+++ b/react/InfosCarrousel/index.jsx
@@ -1,15 +1,28 @@
-import React, { useState, useCallback } from 'react'
+import React, { useState, useCallback, useEffect } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
+import clamp from 'lodash/clamp'
 import SwipeableViews from 'react-swipeable-views'
 
 import styles from './styles.styl'
 import Icon from '../Icon'
 import IconButton from '../IconButton'
 
+const useClampedValue = (initialValue, min, max) => {
+  const [value, setValue] = useState(initialValue)
+
+  const setClampedValue = newVal => setValue(clamp(newVal, min, max))
+
+  useEffect(() => {
+    setClampedValue(value)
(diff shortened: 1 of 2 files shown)


react/InfosCarrousel/index.jsx
```diff
@@ -15,6 +15,8 @@ const InfosCarrousel = ({ children, theme, className, swipeableProps }) => {
   const hasPreviousInfos = index === 0
   const hasNextInfos = index === count - 1
 
+  if (index >= count) setIndex(count - 1) // can happen when children are removed after navigating
+
```
ptbrowne: I am not really comfortable with setting state inside the render.

I think it would be clearer if the index could *never* go below 0 or higher than the index.

What do you think of the following approach where the value is completely constrained by a hook. No double setState are necessary, and internal state is never wrong.

```
const clamp = (value, min, max) => Math.max(min, Math.min(value, max)))

const useClampedCounter = (start, min, max) => {
  const [counter, setCounterRaw] = useState(start) 
  // maybe a useCallback to use here ?
  const setCounter = (index) => setCounterRaw(clamp(value, min, max))
  return [counter, setCounter]
}

const [index, setIndex] = useClampedCounter(0
```

y-lohse: Very good point and elegant solution, I'll give it a try!

ptbrowne: useClampedValue instead of useClampedCounter since no specific methods like increment or decrement are returned by the hook.

y-lohse: I renamed a few other things, and also didn't use `useCallback` on `setCounter` since the function is not directly passed to child components. But it works great, thanks!

PR #1384, opened 2020-02-17
feat: Add ExpModal again

Crash--:
Adding a new Experimental Modal based this time on MUI Dialog and MUI theme. 

Dealed specially with mobile style: 
- fullscreen on mobile 
- small cross on mobile 
- 2 types of action: default (100% width if one button, 50% if 2 ) or below (100% even if several buttons) 

See: https://crash--.github.io/cozy-ui/react/#/Labs?id=experimentalmodal

- [X] Deployed the styleguidist
- [ ] Did you think of ARIA attributes if you are coding new components ?


react/Labs/ExperimentalModal/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+import DialogActions from '@material-ui/core/DialogActions'
+
+import withMobileDialog from '@material-ui/core/withMobileDialog'
+
+import AppTitle from '../../AppTitle'
+
+export const ExperimentalModal = props => {
+  const { dismissAction, fullScreen, children, open } = props
+  return (
+    <MuiCozyTheme>
```
y-lohse: Actually I don't think the component should come with a MuiCozyTheme wrapper. That wrapped is supposed to be provided once, at a higher level in the app. Not including it here also means that the modal's styles can be overriden by a different Theme provider, for example the inverted theme.

react/Labs/ExperimentalModal/index.jsx
```diff
@@ -0,0 +1,38 @@
...
+export const ExperimentalModalTitle = ({ children }) => {
+  return (
+    <DialogTitle disableTypography>
+      <AppTitle>{children}</AppTitle>
+    </DialogTitle>
+  )
+}
+
+export const ExperimentalModalActions = ({ type, children }) => {
+  const className = type === 'below' ? { className: 'modal_actions_below' } : {}
```
y-lohse: `type` is a bit generic, maybe `layout` instead?
And with `below` I wonder "below what?", so I would suggest `stack` or `column` instead.

react/Labs/ExperimentalModal/index.jsx
Crash--: Fair enough!

react/Labs/ExperimentalModal/index.jsx
Crash--: row / column, no?

nono:
@Crash-- it looks like some snapshots should be added/updated to fix the tests on travis

y-lohse:
Looking at the preview for now, not the code yet:

- Could you remove the docs about Dialog ?
- In the new experimental modals, clicking on the backdrop/overlay should close the modal
- I agree with @drazik , mixing `Modal` and `Dialog` is a bit confusing. I think we should use `Dialog` for this one, so that we can phase out `Modal` gradually without running int name conflicts.

react/Labs/ExperimentalModal/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+import AppTitle from '../../AppTitle'
+
+export const ExperimentalModal = props => {
+  const { dismissAction, fullScreen, children, open } = props
+  return (
+    <Dialog
+      open={open}
+      onClose={dismissAction}
+      fullScreen={fullScreen}
+      aria-labelledby="dialog-title"
```
y-lohse: This shouldn't stay like that.

react/Labs/ExperimentalModal/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+      scroll={'paper'}
```
y-lohse: Generally speaking we could probably spread whatever props we get here, couldn't we? Having default prop values is a great addition, especially the `open`, but I don't see the value in renaming `onClose` to `dismissAction` for example.

react/MuiCozyTheme/theme.js
```diff
@@ -243,17 +255,58 @@ normalTheme.overrides = {
...
+  MuiDivider: {
+    root: {
+      [normalTheme.breakpoints.down('md')]: {
+        width: `calc(100% + ${defaultValues.dialog.sm.padding}*2px)`,
+        marginLeft: `-${defaultValues.dialog.sm.padding}px`
+      },
+      [normalTheme.breakpoints.up('md')]: {
+        width: `calc(100% + ${defaultValues.dialog.md.padding}*2px)`,
+        marginLeft: `-${defaultValues.dialog.md.padding}px`
+      }
```
y-lohse: I think this is worth a comment explaining why we do this.

react/Labs/ExperimentalModal/index.jsx
Crash--: Why that? `aria-labelledby`makes a reference to the node id = dialog-title. And this node is the modal title

y-lohse: Oh right, my bad!

react/Labs/ExperimentalModal/index.jsx
Crash--: What it's the goal? Should we stay as close as possible to our actual modal? Or do we embrace fully the API from MUI? If yes, I can remove Modal* reference to only get Dialog (ExperimentalDialog for instance) and use the same props as Modal

y-lohse: The goal is to have a modal with an API and a look that better suits or needs. I don't see a reason to force ourselves staying close to the previous API, in fact it may be quite limiting.

react/Labs/ExperimentalDialog/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const ExperimentalDialog = props => {
+  const { onClose, fullScreen, children, open, scroll } = props
+  return (
+    <Dialog
+      open={open}
+      onClose={onClose}
+      fullScreen={fullScreen}
+      aria-labelledby="dialog-title"
+      scroll={scroll}
```
y-lohse: Can we spread the `props` here? That way if we want to add other props supported by Dialog, we don't need to add them here.

PR #1385, opened 2020-02-18
chore: Let's screenshots all the components 

Crash--:
Since https://github.com/cozy/cozy-ui/pull/1361/files we now have a perSection styleguidist. 

But our current argos script get the links from the displayed page and not from the side menu. This PR fixes that. 

In order to not screenshot both the section page and the component page, I filter the section but looking for a `?` in the URL. Then, I get the text (name) of the component in order to build the right url

### DIFF
--- a/scripts/screenshots.js
+++ b/scripts/screenshots.js
@@ -64,16 +64,19 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
   })
 
   console.log('Extracting links')
-  const links = await page.evaluate(() => {
-    return Array.from(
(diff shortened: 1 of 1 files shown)


Crash--:
Since https://github.com/cozy/cozy-ui/pull/1361/files we now have a perSection styleguidist.

But our current argos script get the links from the displayed page and not from the side menu. This PR fixes that.

In order to not screenshot both the section page and the component page, I filter the section but looking for a ?id= in the URL. Then, I get the text (name) of the component in order to build the right url

PR #1388, opened 2020-02-19
test: Take individual screenshots for the examples

Crash--:
Before this PR we were taking screenshot for a component page. But this page could contain several examples and sometimes it doesn't work well since examples can override each other. For instance, if we had a page with several modals, only the latest will appear on the screenshot.

This PR fix this behavior by creating a screenshot for every example we have in the styleguidist

### DIFF
--- a/scripts/screenshots.js
+++ b/scripts/screenshots.js
@@ -65,7 +65,7 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
   })
 
   console.log('Extracting links')
-  //We want to take screenshot for individual exemple, so we : 
+  //We want to take screenshot for individual example, so we : 
   //-extract categories (link from the side menu with no ?id=)
   //- go to category's page
   //- look for `.rsg--controls-40 a` which is the open isolated for exemples
--- a/scripts/screenshots.js
+++ b/scripts/screenshots.js
@@ -68,7 +68,7 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
   //We want to take screenshot for individual example, so we : 
   //-extract categories (link from the side menu with no ?id=)
   //- go to category's page
-  //- look for `.rsg--controls-40 a` which is the open isolated for exemples
+  //- look for `.rsg--controls-40 a` which is the open isolated for examples
   //- look for its closest data-testid to get the name
   const categoriesName = await page.evaluate(() => {
     return Array.from(document.querySelectorAll('.rsg--sidebar-4 a'))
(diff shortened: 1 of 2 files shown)


scripts/screenshots.js
```diff
@@ -64,22 +65,42 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
...
+  const sortedCategoriesNames = sortBy(
+    categoriesName.map(catName => ({
+      link: styleguideIndexURL + '#/' + catName,
+      name: catName
     })),
     x => x.name
   )
+  const allLinks = [];
+  for (cate of sortedCategoriesNames){
+      console.log('Go to category page', cate.link)
```
y-lohse: We can remove this now :)

Crash--:
> A shame that the examples are called example-1, example-2 

I think I can do better, let me try ;)

scripts/screenshots.js
```diff
@@ -64,22 +65,42 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
   })
 
   console.log('Extracting links')
-  //Categories have link like : #/Labs. Components' link: Labs?id=component
-  //So we filter to not have the categories, and get the component's name from
-  //the text attribute of the a. Like that we can create the right url
-  const names = await page.evaluate(() => {
+  //We want to take screenshot for individual exemple, so we :
```
drazik:
```suggestion
  //We want to take screenshot for individual example, so we : 
```

scripts/screenshots.js
```diff
@@ -64,22 +65,42 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
...
 
   console.log('Extracting links')
-  //Categories have link like : #/Labs. Components' link: Labs?id=component
-  //So we filter to not have the categories, and get the component's name from
-  //the text attribute of the a. Like that we can create the right url
-  const names = await page.evaluate(() => {
+  //We want to take screenshot for individual exemple, so we : 
+  //-extract categories (link from the side menu with no ?id=)
+  //- go to category's page
+  //- look for `.rsg--controls-40 a` which is the open isolated for exemples
```
drazik:
```suggestion
  //- look for `.rsg--controls-40 a` which is the open isolated for examples
```

scripts/screenshots.js
```diff
@@ -65,7 +65,7 @@ const fetchAllComponents = async (page, styleguideIndexURL) => {
   })
 
   console.log('Extracting links')
-  //We want to take screenshot for individual exemple, so we : 
+  //We want to take screenshot for individual example, so we : 
   //-extract categories (link from the side menu with no ?id=)
   //- go to category's page
   //- look for `.rsg--controls-40 a` which is the open isolated for exemples
```
ptbrowne: Nit: spaces are not uniform here. There should be space after the dash in the third list item. Also, it seems the convention to have a space after the double slash indicating the start of the comment.

PR #1389, opened 2020-02-21
fix(I18n): Remove deprecated method warning

drazik:
On the I18n component, we kept preact compatibility by implementing both
`UNSAFE_componentWillMount` and `componentWillMount` methods. But when
using with react, this leads to deprecated methods calls warnings in
console. To avoid this, we can export a component that only implements
`UNSAFE_componentWillMount` by default, and another one that implements
`componentWillMount` when `USE_PREACT` is defined.


drazik:
Because we have some apps (mostly for demo purposes) that are still using preact.

PR #1390, opened 2020-02-28
feat: Add Variant to IconStack

Crash--:
Add a new variant for the IconStack component : bottom-right alignment. 

I created a container of 24x24 following @joel-costa 's advice. In this container, we will have either a 20x20 Circle (I can't use circle, since 20x20 is not one of our size) or a 24x24 rectangle.

At the end of the implementation, I'm not sure that it's still an IconStack :/ 

https://crash--.github.io/cozy-ui/react/#/

### DIFF
--- a/react/IconStack/Readme.md
+++ b/react/IconStack/Readme.md
@@ -1,6 +1,6 @@
 ### IconStack
 
-A component to put something in the background and center an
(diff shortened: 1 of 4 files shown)


ptbrowne:
Isn't it the same as Badge ?

PR #1392, opened 2020-03-02
feat: Add new Badges

Crash--:
Follow up of https://github.com/cozy/cozy-ui/pull/1390

We decided that it was more a Badge than an IconStack. 

I created a container of 24x24 following @joel-costa 's advice. In this container, we have a medium circle or a small circle 

https://crash--.github.io/cozy-ui/react/#/Basics?id=badge


react/Badge/Readme.md
```diff
@@ -1,5 +1,7 @@
 `<Badge />` component wraps the element on which you want to apply a badge. In the following examples, it is applied to compact `<ButtonAction />`.
 
+You should use `<Badge />` when we want to add additionnal informations on the wrapped component
```
y-lohse:
```suggestion
You should use `<Badge />` to provide additional information related to the wrapped content.
```

Just a suggestion, both are correct but I don't like the you/we switch :)

react/Badge/Readme.md
```diff
@@ -28,3 +30,26 @@ import ButtonAction from 'cozy-ui/transpiled/react/ButtonAction';
   </p>
 </div>
 ```
+
+
+#### Alignment
+
+Only support `bottom-right`
```
y-lohse:
```suggestion
Only supports `bottom-right` 
```

react/Badge/Readme.md
```diff
@@ -28,3 +30,26 @@ import ButtonAction from 'cozy-ui/transpiled/react/ButtonAction';
...
+import Badge from 'cozy-ui/transpiled/react/Badge'
+import Icon from 'cozy-ui/transpiled/react/Icon';
+
+
+
+<div>
+  <p>
+    <Badge 
+      content={ <Icon icon="link" size={10} />} type="normal" alignment="bottom-right" size='medium'>
+      <Icon icon="cloud" size="32" />
```
y-lohse: Could we use a `ButtonAction ` here as well? I find it's a bit easier to understand which one the badge is and which one the other component is if we don't use Icon for both :)

Crash--: It was to have a really life example since we currently don't use Badge with a ButtonAction

y-lohse: Mmh ok fair enough!

react/Badge/Readme.md
```diff
@@ -28,3 +30,26 @@ import ButtonAction from 'cozy-ui/transpiled/react/ButtonAction';
...
+    </Badge>
+  </p>
+  
+</div>
+```
```
ptbrowne: I am not sure why we add doc for `alignment` since it can only have one value ?

Crash--: For now, but we plan to add top-right pretty soon

PR #1393, opened 2020-03-03
fix: Move isTesting to helpers

Crash--:
Since we're now using the isTesting method in the MUI theme creation (to disable animation), we need to export it. 

I just moved the file to helpers since it's used that way now


ptbrowne:
I am a bit surprise this was not detected by tests, is there a way to add a test for this ?

PR #1396, opened 2020-03-04
Feat/add bar context provider

Crash--:
Sometimes, we need to display a component from an
 app to the Cozy Bar. In that case, we want to be
 able to provide to this component the "context"
 of the app and not the one from the bar.
This is the case for :
- store
- client
- lang
- f
- t


react/ContactsListModal/DemoProvider.jsx
```diff
@@ -41,19 +40,11 @@ mockClient.plugins = {
...
-    }
-  }
-
   render() {
-    return <>{this.props.children}</>
+    return (
+      <CozyProvider client={mockClient}>{this.props.children}</CozyProvider>
+    )
   }
 }
```
ptbrowne: Nit: This can be transformed into a function component.

react/BarContextProvider/index.jsx
```diff
@@ -0,0 +1,23 @@
...
+          <I18nContext.Provider
+            value={{ f: this.props.f, t: this.props.t, lang: this.props.lang }}
+          >
+            {this.props.children}
+          </I18nContext.Provider>
+        </CozyProvider>
+      </Provider>
+    )
+  }
+}
```
ptbrowne: Nit: This can be transformed into a function component.

react/BarContextProvider/indes.spec.jsx
```diff
@@ -0,0 +1,64 @@
...
+  const store = useStore()
+  return (
+    <MockedBar>
+      <BarContextProvider t={t} f={f} lang={lang} client={client} store={store}>
+        <DumbHelloWorld />
+      </BarContextProvider>
+    </MockedBar>
+  )
+}
+describe('BarContextProvider', () => {
```
ptbrowne: Nit: Please add a space between component declarations and the surrounding code.

react/BarContextProvider/indes.spec.jsx
```diff
@@ -0,0 +1,64 @@
...
+  useClient
+} from 'cozy-client'
+import configureStore from 'redux-mock-store'
+import { Provider, useStore } from 'react-redux'
+
+import BarContextProvider from '.'
+import I18n, { useI18n, translate } from '../I18n'
+const locales = { helloworld: 'Hello World !' }
+const localesBar = { a: 'b' }
+const DumbHelloWorld = translate()(({ t, f, lang }) => (
```
ptbrowne: Nit: Please add a blank line between component declarations and the surrounding code.

PR #1398, opened 2020-03-04
fix: Help icon viewbow

y-lohse:
The existing help icon had a problem, it might be that it had no viewbox or it might be something else. At any rate, displaying it at small sizes would lead to the icon being cut-off:

Capture d'écran 2020-03-04 15 25 07

This PR fixes that.


ptbrowne:
> it might be that it had no viewbox or it might be something else.

I don't understand, what did you do to fix the problem if the root cause is unknown ?

y-lohse:
> I don't understand, what did you do to fix the problem if the root cause is unknown ?

I re-exported the SVG icon from Zeplin. In the final diff, the `path` is the same, but there is an extra viewbox and the path is drawn right away instead of defined first and then printed with a `use` tag.

ptbrowne:
Thanks for the explanation

PR #1399, opened 2020-03-05
fix: Default to align items center

Crash--:
I migrated to Chromium (82.0.4078.0) again and the button's content were not correctly centered. 

It seems that the default user agent stylesheet for button in Chromium is `align-items: flex-start;`. I don't know if this behavior has been changed recently? 
Capture d’écran 2020-03-05 à 11 43 19

exemple: 
Capture d’écran 2020-03-05 à 11 39 14

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -76,6 +76,7 @@ $button
     text-transform   uppercase
     text-decoration  none
     cursor           pointer
+    align-items      center
 
     svg
         fill         currentColor


Crash--:
(argos failed is not about the buttons but an animation ;))

Crash--:
In fact, this default is there since a long time : https://chromium.googlesource.com/chromium/blink/+blame/master/Source/core/css/html.css#587 and nothing changed in this file since a while https://chromium.googlesource.com/chromium/blink/+log/master/Source/core/css/html.css

But it works fine on every other browser. I just tried with Chrome and it works as expected whiteout this change

PR #1401, opened 2020-03-06
feat: Add a specific viewer for shortcut file

Crash--:
Here is an exemple: 

Capture d’écran 2020-03-06 à 11 12 00

wip: need hook from cozy-client


react/Viewer/ShortcutViewer.jsx
```diff
@@ -0,0 +1,36 @@
...
+const ShortcutViewer = ({ file }) => {
+  const client = useClient()
+  const { t } = useI18n()
+  const { shortcutInfos } = useFetchShortcut(client, file.id)
+  let url = ''
+  if (shortcutInfos) {
+    url = new URL(shortcutInfos.data.attributes.url)
+  }
+  return (
+    <NoViewer
```
ptbrowne: Nit: A Viewer rendering NoViewer ? It's a bit surprising, is it intended ?

react/Viewer/ShortcutViewer.spec.jsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <CozyProvider client={client}>
+      <I18n lang="en" dictRequire={() => locales['en']}>
+        <ShortcutViewer file={{ id: '1' }} />
+      </I18n>
+    </CozyProvider>
+  )
+}
+
+describe('ShortCutviewer', () => {
```
ptbrowne:
```suggestion
describe('ShortcutViewer', () => {
```

react/Viewer/ShortcutViewer.spec.jsx
```diff
@@ -0,0 +1,27 @@
...
+  it('test the component', async () => {
+    //const Comp = setup()
```
ptbrowne:
```suggestion
```

react/Viewer/ShortcutViewer.jsx
Crash--: Yep, I think the NoViewer should be renamed to something else, like DefaultViewer ;)

react/Viewer/ShortcutViewer.jsx
```diff
@@ -0,0 +1,36 @@
...
+import NoViewer from './NoViewer'
+import { FileDoctype } from '../proptypes'
+
+const ShortcutViewer = ({ file }) => {
+  const client = useClient()
+  const { t } = useI18n()
+  const { shortcutInfos } = useFetchShortcut(client, file.id)
+  let url = ''
+  if (shortcutInfos) {
+    url = new URL(shortcutInfos.data.attributes.url)
```
y-lohse: nit
```suggestion
    url = new URL(get(shortcutInfos, 'data.attributes.url'))
```

just in case :)

react/Viewer/ShortcutViewer.spec.jsx
```diff
@@ -0,0 +1,26 @@
...
+      </I18n>
+    </CozyProvider>
+  )
+}
+
+describe('Shortcutviewer', () => {
+  it('test the component', async () => {
+    const { container } = render(setup())
+    expect(container).toMatchSnapshot()
+    await wait()
```
y-lohse: Why is it needed?

react/Viewer/ShortcutViewer.spec.jsx
```diff
@@ -0,0 +1,26 @@
...
+    <CozyProvider client={client}>
+      <I18n lang="en" dictRequire={() => locales['en']}>
+        <ShortcutViewer file={{ id: '1' }} />
+      </I18n>
+    </CozyProvider>
+  )
+}
+
+describe('Shortcutviewer', () => {
+  it('test the component', async () => {
```
y-lohse:
```suggestion
  it('renders the component', async () => {
```

react/Viewer/ShortcutViewer.spec.jsx
Crash--: because we have the a promise (hooks) that resolve. So we need to wait the end of the cycle if we don't want any warning ;)

PR #1406, opened 2020-03-09
fix: Viewer Warning and accessibility

Crash--:
We display a button icon only for the PDF viewer, but we have to provide a label to make it mare accessible. It also fixes a few warnings in the console ;)

### DIFF
--- a/react/Viewer/PdfJsViewer.jsx
+++ b/react/Viewer/PdfJsViewer.jsx
@@ -7,6 +7,7 @@ import { Spinner } from '../Spinner'
 import withFileUrl from './withFileUrl'
 import ToolbarButton from './PdfToolbarButton'
 import NoViewer from './NoViewer'
+import { translate } from '../I18n'
 import styles from './styles.styl'
 
 export const MIN_SCALE = 0.25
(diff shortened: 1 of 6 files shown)


react/Viewer/PdfJsViewer.jsx
```diff
@@ -192,4 +198,4 @@ PdfJsViewer.propTypes = {
   renderFallbackExtraContent: PropTypes.func
 }
 
-export default withFileUrl(PdfJsViewer)
+export default translate()(withFileUrl(PdfJsViewer))
```
y-lohse: We can use `useI18n` no?

react/Viewer/PdfJsViewer.jsx
```diff
@@ -172,11 +176,13 @@ export class PdfJsViewer extends Component {
                 icon="dash"
                 onClick={this.scaleDown}
                 disabled={scale === MIN_SCALE}
+                label={t('Viewer.scaledown')}
               />
               <ToolbarButton
                 icon="plus"
                 onClick={this.scaleUp}
                 disabled={scale === MAX_SCALE}
+                label={t('Viewer.up')}
```
y-lohse:
```suggestion
                label={t('Viewer.scaleup')}
```

PR #1407, opened 2020-03-12
feat: Load more component

y-lohse:
This PR adds a new LoadMore component. [See the demo for more infos](https://yannick-lohse.fr/cozy-ui/react/#!/LoadMore)

### DIFF
--- a/react/LoadMore/index.jsx
+++ b/react/LoadMore/index.jsx
@@ -1,0 +1,51 @@
+import React, { useState, useEffect, useRef, useCallback } from 'react'
+import PropTypes from 'prop-types'
+import Button from '../Button'
+import Spinner from '../Spinner'
+
+const LoadMore = ({ fetchMore, label }) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const elementRef = useRef()
+
+  const startFetchMore = useCallback(async () => {
+    if (!isLoading) {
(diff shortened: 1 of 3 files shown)


react/LoadMore/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+const LoadMore = ({ fetchMore, label }) => {
+  const [isLoading, setIsLoading] = useState(false)
+  const elementRef = useRef()
+
+  const startFetchMore = useCallback(async () => {
+    if (!isLoading) {
+      setIsLoading(true)
+      await fetchMore()
+      setIsLoading(false)
```
y-lohse: I considered adding a try/catch here… but we wouldn't be able to do anything with the error except maybe logging it. I think it's better to let it propagate.

react/LoadMore/index.jsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }, [isLoading, fetchMore])
```
ptbrowne: setIsLoading is also a dep right ?

y-lohse:
I've fixed the bug that prevented the button from being displayed, and I've changed the sizes of the blocks so that the spinner isn't visible on page load (which wasn't useful to understand how things work anyway). So the visual tests should be stable now!

PR #1409, opened 2020-03-16
feat: Import the tooltip component from MUI

edas:
Direct import from MUI for a first use in cozy-notes.
We only add a cozy style but the API doesn't change.

### DIFF
--- a/react/Tooltip/Readme.md
+++ b/react/Tooltip/Readme.md
@@ -1,0 +1,8 @@
+### Simplest usage
+
+```js
+import Tooltip from 'cozy-ui/transpiled/react/Tooltip';
+<Tooltip title={'This is an explaination'}><u>hover over me</u></Tooltip>
+```
+
+This component is imported from MUI and has the same API.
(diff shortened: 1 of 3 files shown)


react/Tooltip/index.jsx
```diff
@@ -0,0 +1,18 @@
+import { withStyles } from '@material-ui/core/styles'
+import { Tooltip } from '@material-ui/core'
+
+const styles = {
```
ptbrowne: Can you add a reason why this had to be managed in CSS-in-JS ?

react/Tooltip/Readme.md
```diff
@@ -0,0 +1,8 @@
+### Simplest usage
+
+```js
+import Tooltip from 'cozy-ui/transpiled/react/Tooltip';
+<Tooltip title={'This is an explaination'}><u>hover over me</u><Tooltip>
```
Crash--:
```suggestion
<Tooltip title={'This is an explaination'}><u>hover over me</u></Tooltip>
```

react/Tooltip/index.jsx
y-lohse: Initially it was because the component was used in cozy-notes, but now that we add it in UI I think we could put this css in the MUI theme no?

edas: done

edas:
> Can you add a link to the styleguidist?

done, thank you

PR #1410, opened 2020-03-19
fix: Popover position

Crash--:
Before, the popover was portaled. It worked well
in a normal context, but not when we only have
a portal like a Modal.

The idea, is to display the popover in the
component calling it and put it in a fixed
position (we can't have an absolute one
since Modals for instance are fixed...).

We set the right zIndex to this popover to be well
displayed.

The drawback of this method, is that the Popover
doesn't follow the scroll anymore. But we already
have this behavior in our old Menu. I think we can
deal with it, the time we find a solution.

We need to use an inline style since the MUI override 
of popperjs is setting the position absolute as inline style 
(https://github.com/mui-org/material-ui/blob/v3.x/packages/material-ui/src/Popper/Popper.js#L176) 

Styleguidist 
https://crash--.github.io/cozy-ui/react/

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -7,7 +7,7 @@ import { Media, Bd, Img } from '../Media'
 import BottomDrawer from '../BottomDrawer'
 import withBreakpoints from '../helpers/withBreakpoints'
 import Popper from '@material-ui/core/Popper'
-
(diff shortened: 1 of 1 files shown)


y-lohse:
Changs LGTM but can you deploy the styleguide preview please?

PR #1411, opened 2020-03-20
feat: Adds useBrowserOffline

edas:
This hook is extracted from cozy-notes, to simplify
the offline indicator there. It should return if the browser
is online (default to 'yes' if we have no indicator, like in
Node.js) and react to online and offline events.


Crash--:
I don't know if UI is the right place to put this kind of hook? 

It's hard to know if the hook is in UI or in cozyclient or somewhere else... 

I think having this logic in CozyClient makes more sense since offline / inline is not only about displaying UI but can be used also to do cozy client related stuff.

ptbrowne:
Personally, I think it belongs more in cozy-ui since it does not use anything from the client, I see cozy-ui as our general toolbox. This is not blocking for me.

I agree for the need for documentation :).

react/hooks/useBrowserOffline.js
```diff
@@ -0,0 +1,22 @@
+import { useState, useCallback } from 'react'
+import useEventListener from './useEventListener'
+
+export default function useBrowserOffline() {
+  function isOfflineNow() {
+    return window && window.navigator && window.navigator.onLine === false
+  }
+
+  const [isOffline, setOfflineSince] = useState(isOfflineNow)
```
y-lohse:
```suggestion
  const [isOffline, setOfflineSince] = useState(isOfflineNow())
```

… at least I think? But then why didn't the test catch that?

edas: Yu can pass a function to useState, it will be executed at the first initialization

edas:
> Also, if we keep it in UI, we should add it to the documentation

done

> I don't know if UI is the right place to put this kind of hook?

I have no strong opinion about this. I've used UI because there is no interaction with the server or the data. We already have useEventlistener there.

react/hooks/useBrowserOffline.js
y-lohse: TIL!

PR #1412, opened 2020-03-23
feat: Upload Queue

ptbrowne:
- Extracted component from Drive
- Locales are also extracted and can be tree-shaken

Attributes are added for extra flexibility:

- `popover` can be set to false to display the UploadQueue in the normal
   flow (useful for styleguide)
- `app` is added to customize the name of the app in the header

Visible on https://ptbrowne.github.io/cozy-ui/react/#!/UploadQueue

### DIFF
--- a/react/UploadQueue/index.jsx
+++ b/react/UploadQueue/index.jsx
@@ -1,0 +1,226 @@
+import React, { Component } from 'react'
+import classNames from 'classnames'
+import { translate } from 'cozy-ui/react/I18n'
+import Icon from '../Icon'
+import Spinner from '../Spinner'
+import Stack from '../Stack'
+import withLocales from '../I18n/withLocales'
+import ThresholdBar from '../ThresholdBar'
+import palette from '../../stylus/settings/palette.json'
+import styles from './styles.styl'
+
+import localeEn from './locales/en.json'
+import localeEs from './locales/es.json'
+import localeFr from './locales/fr.json'
+
+const locales = {
+  en: localeEn,
+  es: localeEs,
+  fr: localeFr
+}
+
+const CANCEL = 'cancel'
+const PENDING = 'pending'
+const LOADING = 'loading'
+const CREATED = 'created'
+const UPDATED = 'updated'
+const FAILED = 'failed'
+const CONFLICT = 'conflict'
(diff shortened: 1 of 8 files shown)


react/UploadQueue/styles.styl
```diff
@@ -0,0 +1,161 @@
...
+    width      100%
+    height     .125rem
+    margin     0
+
+progress.upload-queue-progress::-webkit-progress-bar
+    background-color var(--silver)
+progress.upload-queue-progress::-webkit-progress-value
+    background-color var(--dodgerBlue)
+progress.upload-queue-progress::-moz-progress-bar
+    background-color var(--dodgerBlue)
```
y-lohse: Do we still need this style since we're using Tresholdbar?

react/UploadQueue/index.jsx
```diff
@@ -0,0 +1,264 @@
...
+          [styles['upload-queue-item--done']]: done,
+          [styles['upload-queue-item--error']]: error
+        })}
+      >
+        <div
+          className={classNames(
+            styles['item-file'],
+            'u-flex',
+            'u-flex-items-center',
+            'u-p-1'
```
y-lohse: Since we have a stylus declaration for this element, can we move the utility classes there as well?

react/UploadQueue/styles.styl
ptbrowne: Yes, because they are the styles for the overall progress bar.

react/UploadQueue/index.jsx
```diff
@@ -0,0 +1,264 @@
...
+const DONE_STATUSES = [CREATED, UPDATED]
+const ERROR_STATUSES = [CONFLICT, NETWORK, QUOTA]
+
+const splitFilename = filename => {
+  const parts = filename.split('.')
+  return {
+    filename: parts.slice(0, -1).join('.') + '.',
+    extension: parts[parts.length - 1]
+  }
+}
```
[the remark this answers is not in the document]
ptbrowne: This is part of cozy-libs's cozy-doctypes which cannot be tree-shaken since it's transpiled to es5. I think it is small enough to be here.

PR #1414, opened 2020-03-26
feat: Allows to push modals ModalStack

edas:
This change is useful to push a QuotaAlert in the stack, and avoid
the ModalStack to wrap a Modal to a component which already has one.

### DIFF
--- a/react/ViewStack/ModalStack.jsx
+++ b/react/ViewStack/ModalStack.jsx
@@ -12,6 +12,9 @@ import { useStack } from './hooks'
  * - Several modals can be stacked in this manner
  * - The <Modal /> element wrapping <View /> can receive props via the 2nd
  * argument of stackPush : `stackPush(<SmallView />, { size: 'xsmall' })`
+ * Passing `{ noModalStackWrapper: true }` will avoid wrapping your
+ * component inside a `<Modal />`, if the component already embed one.
+ * When doing so, it's up to you to use the `stackPop` method.
  *
  * Can be used as a replacement of a <ViewStack /> on mobile
  */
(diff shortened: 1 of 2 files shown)


ptbrowne:
Another solution would be to make it possible for QuotaAlert not to show its Modal ? I think it would make it more flexible ? This way QuotaAlert is just in charge of showing things related to Quota and the ModalStack takes care of the Modals ? It seems more orthogonal this way and does not expand the API of ModalStack.

On another note, if this is really necessary, I would prefer if the API used only positive attributes : instead of `noModalStackWrapper`, I suggest `wrap: false` ? And `wrap` would be considered `true` unless specified.

edas:
> Another solution would be to make it possible for QuotaAlert not to show its Modal ?

It would be difficult IMHO, as most of QuotaAlert is defining the props of Modal.

> On another note, if this is really necessary, I would prefer if the API used only positive attributes

Yep, I've thought about that but this would means that not all falsy values would mean `false`. For a boolean attribute, I find this error prone.

> instead of noModalStackWrapper, I suggest wrap

The props are normaly sent to modal. We need a special value that could never be interpreted as a Modal prop, hence the special namespaced name.

ptbrowne:
> Yep, I've thought about that but this would means that not all falsy values would mean false. For a boolean attribute, I find this error prone.

I think a boolean property with a default value to true is not error prone.

react/ViewStack/ModalStack.jsx
```diff
@@ -29,11 +32,16 @@ const ModalStack = ({ children }) => {
...
+        const shouldWrap = props ? !props.noModalStackWrapper : true
+        if (shouldWrap) {
+          return (
+            <Modal mobileFullscreen key={i} dismissAction={stackPop} {...props}>
+              {child}
+            </Modal>
+          )
+        } else {
+          return React.cloneElement(child, { key: i })
+        }
```
Crash--: Can't we do a bit of magic and detect that the first child is a Modal / Dialog ? And add a forceModalWrapper if this magic/ default behavior is not the one we want ?

JF-Cozy:
Since we don't use Modal anymore, maybe we should close this ? @Crash--

PR #1415, opened 2020-03-27
Fix background color UploadQueue on mobile

ptbrowne:
Due to a bug, the background was transparent.

### DIFF
--- a/react/UploadQueue/Readme.md
+++ b/react/UploadQueue/Readme.md
@@ -10,7 +10,11 @@ the upload queue:
 - underneath the top-bar on mobile  
 
 ```
-const state = {
+const initialState = {
+  popover: false
+}
+
+const data = {
   queue: [{
     file: { name: 'Photo.jpg' },
     status: 'created'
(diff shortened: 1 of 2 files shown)


react/UploadQueue/Readme.md
```diff
@@ -18,13 +33,16 @@ const state = {
...
+  <UploadQueue
+    lang='fr'
+    app='Cozy Drive'
+    getMimeTypeIcon={() => 'file'}
+    queue={data.queue}
+    doneCount={data.doneCount}
+    successCount={data.successCount}
+    popover={state.popover}
+  />
+</>
```
Crash--: I'm not sure that having an interactive exemple is a good idea for our Argos test. If I'm not mistaken we're not screenshoting when popover={true}, right?

ptbrowne: > I'm not sure that having an interactive exemple is a good idea for our Argos test

Why ?

The checkbox is for the developer working on the UploadQueue.

Crash--: Because we are not screenshoting the component when popover={true} and I think it's not a good idea to no doing it.

ptbrowne: Since this is not the main point of this PR, I prefer to improve the screenshots in another PR https://github.com/cozy/cozy-ui/pull/1416, this way I can merge this fix earlier.

PR #1419, opened 2020-03-30
Add remaining time to individual uploads

ptbrowne:
Add remaining time next to progress bars.

- Only shown on desktop
- This component is only used in Drive but it should possible to use in Photo
- Speed is not shown, I chose this not to clutter the interface too much

https://ptbrowne.github.io/cozy-ui/react/#/UploadQueue

image

cc @joel-costa

### DIFF
--- a/react/ThresholdBar/index.jsx
+++ b/react/ThresholdBar/index.jsx
@@ -11,13 +11,17 @@ import styles from './styles.styl'
  * - Otherwise, the bar is red an a tick indicates the threshold amount
  *   so that we see how far the threshold has been exceeded
  */
-function ThresholdBar({ threshold, value }) {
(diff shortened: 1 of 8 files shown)


y-lohse:
The changes LGTM but visually, if we don't use the remaining time, the treshold bar doesn't go all the way to the left. Is that intended? Are we sure that we'll always have the download speed displayed?

Capture d'écran 2020-03-30 10 30 47

ptbrowne:
It's intended on my part but I did not check with joel yet, this is why I pinged him :). I used a narrower progress bar not to have too much visual clutter.

y-lohse:
@ptbrowne can you let me know when the code is ready for a second review?

ptbrowne:
It's ok now.

PR #1420, opened 2020-03-31
feat: Uses useClient in QuotaAlert

edas:
The small refactor should help having the good displayName
for the component


ptbrowne:
Can you please put a "BREAKING CHANGE:" mention in your commit since QuotaAlert now needs an up-to-date cozy-client ? This could be reflected in the peer dependencies that cozy-ui needs.

PR #1423, opened 2020-04-01
feat: Added DropdownButton component

y-lohse:
This component is used with the `ActionMenu` component, but not only — it will also be used with drive's Breadcrumb component, and more.

Capture d'écran 2020-04-01 11 25 28

- [DropdownButton demo](https://yannick-lohse.fr/cozy-ui/react/#!/DropdownButton)
- [Updated ActionLMenu example](https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu)

### DIFF
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -15,6 +15,7 @@ module.exports = {
         '../react/Badge/index.jsx',
         '../react/Button/index.jsx',
         '../react/ButtonAction/index.jsx',
+        '../react/DropdownButton/index.jsx',
         '../react/Card/index.jsx',
         '../react/CompositeRow/index.jsx',
         '../react/InlineCard/index.jsx',
--- a/react/ActionMenu/Readme.md
+++ b/react/ActionMenu/Readme.md
@@ -4,6 +4,7 @@ Use an ActionMenu to show a list of actions. ActionMenus automatically switch th
 
 ```
 import ActionMenu, { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu';
+import DropdownButton from 'cozy-ui/transpiled/react/DropdownButton';
 import Icon from 'cozy-ui/transpiled/react/Icon';
 
 initialState = { menuDisplayed: isTesting() };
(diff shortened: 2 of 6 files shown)


ptbrowne:
Some questions :

- It seems like the angle quote is too close from the text in the first example
- It seems like the angle quote is too bold in the last example
- Should the weight of the angle charater follow the weight of the text ?
- Should the API really allow to put any component in there ? (follow up of question 1) ? AMHA maybe but it seems like we should be able to control the style of the angle quote to match (quoteSize="xxl" ?)

Not blocking : there is not "open" state to this. Should there be one ? Since I guess this is extracted from an app, I guess we can live without it but I prefer to remark it. 

Those points are not blocking.

y-lohse:
> Should the API really allow to put any component in there ? (follow up of question 1) ? AMHA maybe but it seems like we should be able to control the style of the angle quote to match (quoteSize="xxl" ?)

I think the previous questions are tied to this one. We discussed it with @joel-costa and found the one-size-fits-all icon works well enough across the whole range. But yes, the alternative would be to specify a general size for the component and automatically adjust the icon. The thing I don't like much with that option is that it creates coupling between our text components and the DropdownButton component.

> there is not "open" state to this. Should there be one ?

No, contrary to accordions for example, the icon on this button won't change with the state of the menu. This might be revisited later, but it seems like most other apps keep the arrow pointing down at all times (for example, here on github  😀!)

ptbrowne:
> The thing I don't like much with that option is that it creates coupling between our text components and the DropdownButton component.

AMHA, I think there is implicit coupling here since not all text sizes will work with the arrow, so it would not be problem to show this coupling in the component.

>  But yes, the alternative would be to specify a general size for the component and automatically adjust the icon. 

There is also the quoteSize alternative.

OK for me.

react/DropdownButton/Index.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
```
Crash--, line 1: Index.jsx ? Why not `index.jsx`

PR #1424, opened 2020-04-02
fix: ActionMenu placements

y-lohse:
`ActionMenu` on desktop relies on `Popper` from MUI, itself being a small wrapper around `Popper.js`. By default, Popper will portal-render itself at the root of the DOM. This caused ActionMenu to malfunction when used within a modal (see https://github.com/cozy/cozy-ui/commit/92cccc24fb732371c4c38bb8e687771a164f7468).

The previous fix was to disable portaling. This solved the z-index problem but caused overflowing problems, so `position: fixed` was used as well. The problem is that as soon as `disablePortal` is activated, Popper won't honor the `placement` attribute anymore. So that prop has been broken since 92cccc24fb732371c4c38bb8e687771a164f7468.

Since *not* using Portal isn't an option, we need to control where the Portal is rendered. We can dothis using the [`container` prop](https://v3.material-ui.com/api/popper/#popper-api). So this PR adds a `containerElRef`, similar to the existing `anchorElRef`. It can be used to portal the ActionMenu into the Modal, which solves the initial z-index problem.

This PR is marked as breaking change because unfortunately, it does mean that there is a manual action required when the ActionMenu is used inside a Modal (for example). If no `containerElRef` is provided it will work as usual. The only affected lib by this change is cozy-scanner.

My cozy-ui preview is currently busy with another PR, i'll update this PR once the new preview is online.

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -12,13 +12,18 @@ const ActionMenuWrapper = ({
   inline,
   onClose,
   anchorElRef,
+  containerElRef,
   placement,
   preventOverflow,
   children
 }) => {
-  const getAnchorElement = useCallback(() => {
-    return anchorElRef.current
-  }, [anchorElRef])
+  const getElementFromRefCallback = ref =>
+    useCallback(() => {
+      return ref ? ref.current : undefined
+    }, [ref])
+
+  const getAnchorElement = getElementFromRefCallback(anchorElRef)
+  const getContainerElement = getElementFromRefCallback(containerElRef)
   const normalOverflowModifiers = {
     preventOverflow: { enabled: false },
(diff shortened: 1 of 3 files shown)


react/ActionMenu/index.jsx
```diff
@@ -12,13 +12,19 @@ const ActionMenuWrapper = ({
...
   placement,
   preventOverflow,
   children
 }) => {
-  const getAnchorElement = useCallback(() => {
-    return anchorElRef.current
-  }, [anchorElRef])
+  const getElementFromRefCallback = ref =>
+    useCallback(() => {
+      console.log(ref, ref.current)
```
ptbrowne, line 22:
```suggestion

```

PR #1427, opened 2020-04-03
Draft feat: Take less screenshots

ptbrowne:
Argos limits us to 500 screenshots so instead of screenshotting each
example, we screenshot each component page

One additional challenge introduced by screenshotting at component level, is that for some examples, we *need* to screenshot at example levels, for example with modals or action menus. To support this usecase, a configuration attribute `perExampleScreenshot` is introduced. When this is configured, screenshot are taken at example level.

### DIFF
--- a/scripts/screenshots.js
+++ b/scripts/screenshots.js
@@ -13,6 +13,7 @@ try {
 const path = require('path')
 const fs = require('fs')
 const sortBy = require('lodash/sortBy')
+const flattenDeep = require('lodash/flattenDeep')
 const { ArgumentParser } = require('argparse')
 
 const emptyDirectory = directory => {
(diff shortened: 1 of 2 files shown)


scripts/screenshots.js
```diff
@@ -29,6 +29,7 @@ const defaultGetScreenshotName = ({ component, viewport }) =>
 
 const formatViewport = viewport => `${viewport.width}x${viewport.height}`
 
+const pushAll = (arr, items) => arr.push.apply(arr, items)
```
y-lohse, line 32: Nit, but here's a little trick: `array.push` accepts an infinite number of args, so `arr.push(...items)` does the same :)

ptbrowne: Yes but it also makes a copy of the array and also needs a tad bit more code ;)

image

y-lohse:
Mmh I don't think this works as intended, it seems like we don't have screenshots at an example level for ActionMenu and Modal. Are you seeing the same thing?

ptbrowne:
Yes, after renaming the config attribute in code, I did not rename it in the config file, should be better now.

y-lohse:
Nice, I believe this time we're good!

(The desktop action menu screenshots are too small but it's unrelated to this PR.)

PR #1429, opened 2020-04-07
feat: Support more complex action menu items

y-lohse:
This PR adds support for having more complex content than a string in ActionMenuItems. Technically it was already possible to add any component as a child of `ActionMenuItem`, but the alignment was out of place.

The new feature is illustrated with a label + description.

Capture d'écran 2020-04-07 09 55 48

See the first example as a preview: https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu

### DIFF
--- a/react/ActionMenu/Readme.md
+++ b/react/ActionMenu/Readme.md
@@ -6,6 +6,7 @@ Use an ActionMenu to show a list of actions. ActionMenus automatically switch th
 import ActionMenu, { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu';
 import DropdownButton from 'cozy-ui/transpiled/react/DropdownButton';
 import Icon from 'cozy-ui/transpiled/react/Icon';
+import { Caption } from 'cozy-ui/transpiled/react/Text';
 
 initialState = { menuDisplayed: isTesting() };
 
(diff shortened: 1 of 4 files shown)


react/ActionMenu/Readme.md
```diff
@@ -19,7 +20,14 @@ const hideMenu = () => setState({ menuDisplayed: false });
       onClose={hideMenu}>
       <ActionMenuItem left={<Icon icon='file' />}>Item 1</ActionMenuItem>
       <ActionMenuItem left={<Icon icon='right' />}>Item 2</ActionMenuItem>
-      <ActionMenuItem left={<Icon icon='file' />}>Item 3</ActionMenuItem>
+      <ActionMenuItem left={<Icon icon='file' />}>
+        <div>
+          Item 3
+          <Caption className="u-maw-5 u-mt-half">
```
y-lohse, line 26: I don't have a problem with letting the user set a max-width, most of the time there will already be one anyway and if not it's dependent on the situation.

I'm more annoyed by the margin, but again, I figure the margin might not be the same for a description and for other components that are not specified yet… so I thought I'd leave it open as well.

ptbrowne: I don't understand why we have to do that ?

y-lohse: I thought in most cases there would be a parent with a max-width but I was wrong. This makes it worth adding a component for the description imo, I'll add it.

ptbrowne: I think the max-width should be directly on the ActionMenu.

y-lohse:
I updated the code and preview — the ActionMenu now has a max-width on desktop!

stylus/components/action-menu.styl
```diff
@@ -9,7 +10,7 @@ $actionmenu
 
 $actionmenu--inline
     width auto
-    min-width rem(220)
+    width rem(256)
```
ptbrowne, line 13: The width auto can be removed right ?

PR #1431, opened 2020-04-08
Fix upload queue item for directories

ptbrowne:
Splitting the filename had a bug when there was no extension: it would return { filename: ".", extension: "" } if the filename had no extension, resulting in upload queue rows with only "." if the item was a directory or an extension less file.

### DIFF
--- a/react/UploadQueue/index.jsx
+++ b/react/UploadQueue/index.jsx
@@ -36,9 +36,10 @@ const ERROR_STATUSES = [CONFLICT, NETWORK, QUOTA]
 
 const splitFilename = filename => {
   const parts = filename.split('.')
+  const hasExtension = parts.length > 1
   return {
-    filename: parts.slice(0, -1).join('.') + '.',
-    extension: parts.length > 1 ? parts[parts.length - 1] : ''
+    filename: hasExtension ? parts.slice(0, -1).join('.') + '.' : parts[0],
+    extension: hasExtension ? parts[parts.length - 1] : ''
   }
 }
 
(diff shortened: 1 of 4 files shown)


react/UploadQueue/index.jsx
```diff
@@ -36,9 +36,10 @@ const ERROR_STATUSES = [CONFLICT, NETWORK, QUOTA]
...
 const splitFilename = filename => {
   const parts = filename.split('.')
+  const hasExtension = parts.length > 1
   return {
-    filename: parts.slice(0, -1).join('.') + '.',
-    extension: parts.length > 1 ? parts[parts.length - 1] : ''
+    filename: hasExtension ? parts.slice(0, -1).join('.') + '.' : parts[0],
+    extension: hasExtension ? parts[parts.length - 1] : ''
   }
 }
```
Crash--, line 44: I remember @y-lohse proposing the use of the method we have in cozy-doctypes to split the filename. We can move it to cozy-client and it solves the issue of having a new lib that can't be tree-shaken, no? 

Because here, if we upload a folder starting with `.` for instance `.git`, we'll have nothing for filename but `git` as extension. And this is not what we want :/.  

We should test the type of the type of the file. We've iterated a few times on the split method, we should really use it (see the latest fix https://github.com/cozy/cozy-libs/commit/f95044581d34d4eaaa4d65cf2d79ac11e50c016f)

ptbrowne: https://github.com/cozy/cozy-client/pull/681

PR #1432, opened 2020-04-09
fix: Remove need for containerElRef

ptbrowne:
In a previous PR, containerElRef was introduced so that ActionMenus
could be rendered correctly in modals. The problem was that we don't
know a priori if a component is going to be rendered in a Modal, which
makes it impossible to pass the modal ref to the action menu.

Here a context automatically provides the ref through a PopperContainerContext.
containerElRef is kept for backward compatibility.

~~One caveat here: the ActionMenu must be rendered after the ref object has
been filled with React which means that the ActionMenu content must *not* be
rendered in the same render round as the Modal. This is why in the example
I had to separate the Modal content from the modal itself. Otherwise,
when opening the actionMenu, the Modal was re-rendered since the modalOpened
state and actionMenu state were kept in the same state object. Hopefully,
in real case scenario, the actionMenu state and modal state are not kept
together, which would prevent this bug from happening. This still feels
a bit brittle :(~~

This caveat has been fixed by putting the popperContainerRef on a sibling element of the modal content, not on the parent.

### DIFF
--- a/react/ActionMenu/Readme.md
+++ b/react/ActionMenu/Readme.md
@@ -116,32 +116,27 @@ The ActionMenu component is rendered at the root of the DOM tree using a [Portal
 In that case, we can use the `containerElRef` prop to provide a reference to an element where ActionMenu will be rendered. Here is an example for the most common case, an ActionMenu inside a Modal:
 
 ```
+import { useState, useRef } from 'react'
 import ActionMenu, { ActionMenuItem, ActionMenuHeader } from './index';
 import Icon from '../Icon';
 import Modal, { ModalDescription } from '../Modal';
 import Filename from '../Filename';
 
-initialState = { modalDisplayed: isTesting(), menuDisplayed: isTesting() };
+initialState = { modalDisplayed: isTesting() };
 
 const showModal = () => setState({ modalDisplayed: true });
 const hideModal = () => setState({ modalDisplayed: false });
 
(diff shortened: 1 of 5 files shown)


react/ActionMenu/Readme.md
```diff
@@ -116,39 +116,42 @@ The ActionMenu component is rendered at the root of the DOM tree using a [Portal
 In that case, we can use the `containerElRef` prop to provide a reference to an element where ActionMenu will be rendered. Here is an example for the most common case, an ActionMenu inside a Modal:
```
y-lohse, line 116: I think we need to update this description text as we no longer use the prop in the example. Maybe we can also update the proptype description for `containerRefEl`?

react/ActionMenu/Readme.md
```diff
@@ -109,11 +109,9 @@ const anchorRef = React.createRef();
 </div>
 ```
 
-### ActionMenu within other components and containerElRef
+### ActionMenu within modals
```
y-lohse, line 112: I think we can remove this section, `ActionMenu inside custom fixed elements` is enough imo.

ptbrowne: As discussed, I kept the section to benefit from Argos tests, and put the modal section below (for reading coherence, you learn about the technique, then learn about its usage in Modals).

PR #1434, opened 2020-04-17
feat: BarTitle

ptbrowne:
Add component suitable to display a title in the BarCenter.

https://ptbrowne.github.io/cozy-ui/react/#/BarTitle


react/BarTitle/styles.styl
```diff
@@ -0,0 +1,6 @@
+.BarTitle
+  margin 0
+  height 3rem
+  display flex
+  align-items center
+  font-size 1.125rem
```
y-lohse, line 6: Should we be reusing one of the existing title levels for the font size and line height? the component makes sense in terms of handling the spacing, but I seem to recall the actual text in cozy-bar titles is the same as one of the regular title levels?

ptbrowne: I am not sure that if someone wants to change the h3 title, the bar font-size should also change. To me, this is a coincidence that it has the same font-size.

I agree though that we need a place to configure and audit those font-size : I changed to use the typography settings to put the bar font size.

PR #1437, opened 2020-04-20
feat: Hero improvements

ptbrowne:
Following recent use of Hero in passwords-web.

- Add horizontal margin between sections
- Force CTA zone to have a max-width

### DIFF
--- a/react/Hero/Readme.md
+++ b/react/Hero/Readme.md
@@ -39,6 +39,7 @@ const BLUE = '#2d8af2';
   </Sections>
   <CTA>
     <Button
+      extension="full"
       label={t('Onboarding.connect-bank-account')}
       onClick={() => flash(t('ComingSoon.description'))} />
   </CTA>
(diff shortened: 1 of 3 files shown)


Crash--:
Can you deploy the styleguidist?

ptbrowne:
It's already deployed for the Stepper, you can see the result in argos though.

y-lohse:
In the first argos example, the button is now very wide on desktop, because of the `full` extension. Is it intended to be used like that on desktop resolutions?

ptbrowne:
> In the first argos example, the button is now very wide on desktop, because of the full extension. Is it intended to be used like that on desktop resolutions?

It is like that in the passwords app.

cc @joel-costa

joel-costa:
The width of the button was chosen to have the same width on all the buttons in the differents screen of the password manager: https://gyazo.com/2c09e680cd371f9604ff55e0976ed070
I don't know if it's the width I want for all Hero uses. I'm not even sure it's a good idea that Hero contains a title + a list of items + a button. Maybe we just need a simple list of items. Then why not make a meta-component that contains this item list component, but not sure that we would use it for the password manager.

In short, I will not have an interesting opinion until I have made a real spec for a generic component. Until then, having a larger or smaller button shouldn't be a problem.

ptbrowne:
Let's merge then.

PR #1439, opened 2020-04-20
Breadcrumbs

ptbrowne:
Breadcrumbs from Banks.

I suppose they are different in Drive, maybe we'll have a variant and then
try to reconciliate the two ?

Visible here : https://ptbrowne.github.io/cozy-ui/react/#/Breadcrumbs

### DIFF
--- a/react/Breadcrumbs/index.jsx
+++ b/react/Breadcrumbs/index.jsx
@@ -1,0 +1,103 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import cx from 'classnames'
+import styles from './styles.styl'
+import Icon from '../Icon'
+
+const BreadcrumbSeparator = () => (
+  <span className={styles.BreadcrumbSeparator}>/</span>
+)
+
+const BreadcrumbItem = ({
+  name,
+  onClick,
+  isCurrent = false,
+  tag = 'span',
+  showSeparator = false
+}) => {
+  const Tag = tag
+  return (
+    <div
+      className={cx(styles.BreadcrumbItem, {
+        [styles['BreadcrumbItem--current']]: isCurrent
+      })}
+    >
+      <Tag
+        onClick={onClick}
+        className={cx({
+          'u-clickable': onClick
+        })}
+      >
+        {name}
+      </Tag>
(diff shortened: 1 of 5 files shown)


y-lohse:
> I suppose they are different in Drive, maybe we'll have a variant and then try to reconciliate the two ?

They are definitely different. I don't mind adding this version to UI (haven't looked at the code yet), but it does mean that we might have a breaking change on this component later on.

ptbrowne:
Yes if the API changes, we'll have a breaking change.

react/Breadcrumbs/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+          isCurrent
+        />
+      </div>
+    </div>
+  )
+}
+
+Breadcrumbs.propTypes = {
+  className: PropTypes.string,
+  color: PropTypes.oneOf(['default', 'primary']),
```
y-lohse, line 92: I suggest renaming this to `theme` to be more in line with other UI components. Color is usually fo passing an actual color value.

react/Breadcrumbs/index.jsx
```diff
@@ -0,0 +1,96 @@
...
+          }
+          className={cx(styles.Breadcrumb__previousButton, styles[color])}
+          onClick={lastPreviousItem.onClick}
+        />
+      )}
+      <div className={cx(styles.Breadcrumb__items, styles[color])}>
+        <div className={styles.Breadcrumb__previousItems}>
+          {previousItems.map(({ name, onClick, tag }, index) => (
+            <BreadcrumbItem
+              key={name}
```
y-lohse, line 72:
```suggestion
              key={index}
```

I think `name` might not be unique enough, for example in a folder architecture like `Jobs/Cozy/Cozy/Whatever` we would have identical keys and i'm not sure which one react would keep if only one is removed.

ptbrowne:
Thanks, I made the fixes. I also refactored a bit so that `theme` is only used in the stylus file. Normally the Breadcrumb should not care about the theme. Since it only deals with Text and Icons, they should be automatically take their color from the context.

PR #1441, opened 2020-04-22
fix: Add missing icons

y-lohse:
This PR adds 2 icons that are used in Drive but were missing here, "check-square" and "note". Since there's nothing interactive argos should be a good enough preview of the change :)


react/Icon/Readme.md
```diff
@@ -62,8 +62,8 @@ The `icon` attribute can take a arbitrary React component. It is expected that
 this component print an `<svg>` tag and forwards any props to it.
 
 ```jsx static
-const MyIcon = (props) => <svg {...props}>…</svg>;
-<Icon icon={myIcon} width={32} height={32} color="purple" />
+const MyIcon = props => <svg {...props}>…</svg>
+;<Icon icon={myIcon} width={32} height={32} color="purple" />
```
ptbrowne, line 66: why ?

y-lohse: Woops, my editor was a bit overzealous with formatting.

PR #1445, opened 2020-04-23
Support ActionMenu having null children

y-lohse:
In Drive we have an ActionMenu but not all items are visible at all time — if one of the items should be hidden, it returns null instead, as is common in React. This was breaking `ActionMenu` because it was checking `null.type`.

I also added an example for the `right` prop of ActionMenuItem so that we have visual regression tests for it.

### DIFF
--- a/react/ActionMenu/index.spec.jsx
+++ b/react/ActionMenu/index.spec.jsx
@@ -1,0 +1,39 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import ActionMenu, { ActionMenuItem } from './'
+
+describe('ActionMenu', () => {
+  let createRangeBackup
+
+  beforeAll(() => {
+    createRangeBackup = global.document.createRange
+    global.document.createRange = jest.fn(() => ({
+      setStart: () => {},
+      setEnd: () => {},
+      commonAncestorContainer: {
+        nodeName: 'BODY',
+        ownerDocument: document
+      }
+    }))
+  })
+
+  afterAll(() => {
+    global.document.createRange = createRangeBackup
+  })
+
+  it('should support null children', () => {
+    const comp = mount(
+      <ActionMenu onClose={jest.fn()}>
+        <ActionMenuItem>Item 1</ActionMenuItem>
+        {null}
+      </ActionMenu>
+    )
+    expect(
+      comp
+        .find(ActionMenuItem)
+        .parent()
+        .getElement()
+    ).toMatchSnapshot()
+  })
+})
--- a/react/ActionMenu/index.spec.jsx
+++ b/react/ActionMenu/index.spec.jsx
@@ -2,25 +2,10 @@ import React from 'react'
 import { mount } from 'enzyme'
 
 import ActionMenu, { ActionMenuItem } from './'
+import { fixPopperTesting } from '../Popper/testing'
 
 describe('ActionMenu', () => {
-  let createRangeBackup
-
-  beforeAll(() => {
-    createRangeBackup = global.document.createRange
-    global.document.createRange = jest.fn(() => ({
-      setStart: () => {},
-      setEnd: () => {},
-      commonAncestorContainer: {
-        nodeName: 'BODY',
-        ownerDocument: document
-      }
-    }))
-  })
-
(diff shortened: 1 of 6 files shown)


ptbrowne:
Can you add a test for the null case ?

Crash--:
> In Drive we have an ActionMenu but not all items are visible at all time — if one of the items should be hidden, it returns null instead, as is common in React 

We're thinking about that and it's not impossible that instead of hiding features, we'll simply "disabled them" by displaying them with a special style and removing the onclick action.

y-lohse:
> We're thinking about that and it's not impossible that instead of hiding features, we'll simply "disabled them" by displaying them with a special style and removing the onclick action.

Some will likely stay fully hidden, for example those that we want either in the topbar or in the menu depending on the context.

> Looking good, but can you add a test for the null case ?

Yup, done !

react/ActionMenu/index.spec.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  afterAll(() => {
+    global.document.createRange = createRangeBackup
+  })
+
+  it('should support null children', () => {
+    const comp = mount(
+      <ActionMenu onClose={jest.fn()}>
+        <ActionMenuItem>Item 1</ActionMenuItem>
+        null
```
ptbrowne, line 28: This is a null string here ?

I think you want `{ null }`.

react/ActionMenu/index.spec.jsx
```diff
@@ -0,0 +1,33 @@
...
+  beforeAll(() => {
+    createRangeBackup = global.document.createRange
+    global.document.createRange = jest.fn(() => ({
+      setStart: () => {},
+      setEnd: () => {},
+      commonAncestorContainer: {
+        nodeName: 'BODY',
+        ownerDocument: document
+      }
+    }))
```
ptbrowne, line 17: Why is this needed ?

Crash--: We have issue for testing component using Tooltip from MUI. See https://github.com/mui-org/material-ui/issues/15726#issuecomment-493124813 

jsdom-16 fix the issue, but we need to use https://www.npmjs.com/package/jest-environment-jsdom-sixteen. I tried few months ago but still the issue. Maybe it's fine now.

react/ActionMenu/index.spec.jsx
y-lohse: 🤦‍♂️thanks

react/ActionMenu/index.spec.jsx
ptbrowne: AMHA this should not be in this test file and should be called with a name that reviewers can understand.

Tooltip-testing.js

```
const fixTooltipTesting = () => {
  let createRangeBackup
  beforeAll(() => {
    createRangeBackup = global.document.createRange
    global.document.createRange = jest.fn(() => ({
      setStart: () => {},
      setEnd: () => {},
      commonAncestorContainer: {
        nodeName: 'BODY',
        ownerDocument: document
      }
    }))

  afterAll(() => {
    global.document.createRange = createRangeBackup
  })
}
```

This way we do not need a comment, it's clearer for the reviewer.

Crash--: and we can import it from other libs 👍

react/ActionMenu/index.spec.jsx
ptbrowne: Was the testing passing with the "null" string ? If yes, this could be a problem.

react/Popper/testing.js
```diff
@@ -0,0 +1,23 @@
...
+      }
+    }))
+  })
+
+  afterAll(() => {
+    global.document.createRange = createRangeBackup
+  })
+}
+
+export default fixPopperTesting
```
ptbrowne, line 23: Nit: since this file could have other functions (helping to interact with a Popper from the test), I think it would be better to have a normal export instead of a default one.

As a user, I would try `import { fixPopperTesting } from './testing'` since the file is named `testing` and not `fixPopperTesting.js`

ptbrowne:
👍

Issue #1333, opened 2020-01-02
Non-MUI components are not aware of theme


drazik:
We now have two MUI theme variants: normal and inverted (see https://github.com/cozy/cozy-ui/pull/1317).

In https://github.com/cozy/cozy-keys-lib/pull/63 we had to play a lot with utility classes to change the color of some cozy-ui components to make it look good on primary color background (ie inverted theme). @ptbrowne suggested that our components could be aware of the theme and use different color according to the theme.

For now we have some kind of distinction between MUI and non-MUI components. The theme is clearly marked as "MUI". Maybe we should have a global theme that would be used for MUI and non-MUI components?

See comments on https://github.com/cozy/cozy-keys-lib/pull/63.

edas:
I see 3 options : 


1. Let, step by step, allow UI components search their color and margin inside the MUI theme, and use an MUI theme to define the context for our React components

2. Create our own theme structure, inside cozy-UI, and let cozy-UI fill itself the MUI theme when the cozy-UI theme is changed / defined

3. Create an `<CozyInverted/>` or something like that, that swaps the CSS variables for the colors 

The 3rd option is very cheap and should work in most of simplier cases. If we want a full featured solution, I prefer the 2nd, at least because we still can swap our CSS variables without rewriting every component.

y-lohse:
We agreed to go towards a mix of 2 and 3: Add a `CozyTheme` component that changes the value of the css color variables for it's children, and that also affects the Material UI theme. Feel free to take up this issue, especially if you need to introduce an `inverted` look for one of our existing components!

ptbrowne:
A CozyTheme component has now been introduced :tada:

PR #1451, opened 2020-04-28
 CozyTheme : Buttons and BarButton

ptbrowne:
Added CozyTheme wrapper component. It overrides CSS variables that are used by downstream components to style themselves.

The beginning implementation is done here for Buttons and BarButton.

Example is here : https://ptbrowne.github.io/cozy-ui/react/#!/CozyTheme

Original issue : https://github.com/cozy/cozy-ui/issues/1333

- [x] Button regular/secondary
- [x] Wrapper for MuiCozyTheme
- [x] BarButton

----

I left in this PR two implementations for the Button (see [Button use context to add a className](https://github.com/cozy/cozy-ui/pull/1451/commits/b03bb91991e9918e0ef22f66114798941a837b5f) and [CozyTheme overrides variables used in the Button](https://github.com/cozy/cozy-ui/pull/1451/commits/a3a9dbe4ab4e0d449935ffbceee383e1206cb40b)) as I found later on that overriding CSS variables as stated in the issue was better since the theme was in control of the variables and components did not have to know about the theme. I kept them as I find interesting to see the difference.

### DIFF
--- a/react/CozyTheme/Readme.md
+++ b/react/CozyTheme/Readme.md
@@ -6,8 +6,9 @@ will be automatically styled.
 
 ```
 import CozyTheme from 'cozy-ui/transpiled/react/CozyTheme';
-import { SubTitle } from 'cozy-ui/transpiled/react/Text';
+import { Title, SubTitle } from 'cozy-ui/transpiled/react/Text';
 import Button from 'cozy-ui/transpiled/react/Button';
+import BarButton from 'cozy-ui/transpiled/react/BarButton';
 
 const props = [{}, { disabled: true}, { busy: true }];
 
@@ -16,18 +17,19 @@ const themesSupportingContext = [
   'secondary'
 ];
 
-<div className='u-bg-primaryColor u-p-1'>
-  <SubTitle className='u-white'>Inverted theme</SubTitle>
-  <CozyTheme variant='inverted'>
-    {themesSupportingContext.map(theme =>
-      <p key={theme}>{
-        props.map(
-          props => <Button key={theme + JSON.stringify(props)} label={theme} theme={theme} {...props} />
-        )
-      }</p>
-    )}
-  </CozyTheme>
-</div>
+<CozyTheme variant='inverted' className='u-p-1'>
+  <Title className='u-white'>Inverted theme</Title>
+  <SubTitle className='u-white'>Buttons</SubTitle>
+  {themesSupportingContext.map(theme =>
+    <p key={theme}>{
+      props.map(
+        props => <Button key={theme + JSON.stringify(props)} label={theme} theme={theme} {...props} />
+      )
+    }</p>
+  )}
+  <SubTitle className='u-white'>BarButton</SubTitle>
+  <BarButton icon='dots' />
+</CozyTheme>
 ```
 
 `MuiCozyTheme` is used under the hood so that MaterialUI components are
--- a/react/CozyTheme/Readme.md
+++ b/react/CozyTheme/Readme.md
@@ -1,0 +1,43 @@
+An area of a page can have a different CozyTheme and components under theme
+will be automatically styled.
+
+- At the moment only some Buttons support this.
+
+
+```
+import CozyTheme from 'cozy-ui/transpiled/react/CozyTheme';
+import { SubTitle } from 'cozy-ui/transpiled/react/Text';
+import Button from 'cozy-ui/transpiled/react/Button';
+
+const props = [{}, { disabled: true}, { busy: true }];
+
+const themesSupportingContext = [
+  'regular',
+  'secondary'
+];
+
+<div className='u-bg-primaryColor u-p-1'>
+  <SubTitle className='u-white'>Inverted theme</SubTitle>
+  <CozyTheme variant='inverted'>
+    {themesSupportingContext.map(theme =>
+      <p key={theme}>{
+        props.map(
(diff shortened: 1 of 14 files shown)


joel-costa:
Primary button (hover) : You used the #silver color on background => rgba(255, 255, 255, 0.88) was expected.
Secondary button (hover) : 0.24 opacity border + 0.24 opacity background = 0.48 opacity border => Could we delete the border on hover to keep a full 0.24 opacity button ?

ptbrowne:
> Primary button (hover) : You used the #silver color on background => rgba(255, 255, 255, 0.88) was expected.

OK, fixed.

>  Secondary button (hover) : 0.24 opacity border + 0.24 opacity background = 0.48 opacity border => Could we delete the border on hover to keep a full 0.24 opacity button ?

No, this is not possible right now. This could be possible though if we introduce other variables for the button theme (borderColor and activeBorderColor; right now buttons are defined only with primaryColor, secondaryColor, activeColor and contrastColor).

I think we should check together the implementation for both of us to understand the constraints.

joel-costa:
Fine with me

react/CozyTheme/Readme.md
```diff
@@ -0,0 +1,45 @@
+An area of a page can have a different CozyTheme and inside components
```
y-lohse, line 1:
```suggestion
An area of a page can have a different CozyTheme and components inside
```

react/CozyTheme/styles.styl
```diff
@@ -0,0 +1,16 @@
+$transluentWhite=rgba(255, 255, 255, 0.88)
```
y-lohse, line 1:
```suggestion
$translucentWhite=rgba(255, 255, 255, 0.88)
```

react/CozyTheme/styles.styl
```diff
@@ -0,0 +1,16 @@
...
+
+.CozyTheme--inverted
+    background: var(--primaryColor)
+    --barIconColor white
+    --barIconColorDisabled $transluentWhite
+
+    --regularButtonPrimaryColor: var(--primaryContrastTextColor)
+    --regularButtonSecondaryColor: var(--primaryContrastTextColor)
+    --regularButtonActiveColor: $transluentWhite
```
y-lohse, line 10:
```suggestion
    --regularButtonActiveColor: $translucentWhite
```

react/CozyTheme/styles.styl
```diff
@@ -0,0 +1,16 @@
...
+    --regularButtonConstrastColor: var(--primaryColor)
+
+    --secondaryButtonPrimaryColor: var(--primaryColor)
+    --secondaryButtonSecondaryColor: var(--white)
+    --secondaryButtonActiveColor: rgba(255, 255, 255, 0.24)
```
y-lohse, line 15: Should we create another var for this?

react/CozyTheme/styles.styl
```diff
@@ -0,0 +1,16 @@
+$transluentWhite=rgba(255, 255, 255, 0.88)
+
+.CozyTheme--inverted
+    background: var(--primaryColor)
+    --barIconColor white
```
y-lohse, line 5:
```suggestion
    --barIconColor var(--white)
```

react/CozyTheme/Readme.md
```diff
@@ -0,0 +1,45 @@
...
+  'secondary'
+];
+
+<CozyTheme variant='inverted' className='u-p-1'>
+  <Title className='u-white'>Inverted theme</Title>
+  <SubTitle className='u-white'>Buttons</SubTitle>
+  {themesSupportingContext.map(theme =>
+    <p key={theme}>{
+      props.map(
+        props => <Button key={theme + JSON.stringify(props)} label={theme} theme={theme} {...props} />
```
y-lohse, line 26: i'm not sure this is a problem, but I was initially confused about Cozy theme vs Button theme by reading this example. Seeing it live makes it clearer though, but I wonder if we should make that distinction in the code?

react/CozyTheme/styles.styl
ptbrowne: I've made the change.

react/CozyTheme/styles.styl
ptbrowne: I am actually not sure this is a good idea to use CSS variables for base colors. This goes against the practice of using semantic colors in the theme. Otherwise as some point someone will override `white` to mean `black` or something to have the text on a button match what she wants, and then `white` would contain `black` and it would have no sense.

We should instead use a CSS variable referencing a contrast color. I've made the change to use `--primaryContrastTextColor`.

I had not make the change for buttons since we may use var(--white) to override the color in skins, but we sure should make the change in the future

react/CozyTheme/Readme.md
ptbrowne: Do you have a suggestion here ?

ptbrowne: I do not plan to make any change here as I want to merge this PR ASAP.

y-lohse: No change needed, it's more of a nit. I was mostly interested to know if you or someone else thinks the 2 theme concepts here is confusing or not. We can improve this as the example grows.

ptbrowne: Yes I think this is a bit confusing.

PR #1452, opened 2020-04-29
feat: Enable ActionMenu to automatically close itself

y-lohse:
At the moment, the ActionMenu stays open when an item is clicked. This is not the default behavior of most of our menus, where the menu is be closed as soon as an action is picked.

Eventually this should be the default behavior, but I couldn't find a way to make it the default without introducing a breaking change. So this PR adds an `autoclose` prop that can be set for new menus while the old ones are migrated, and after a period we can remove this prop entirely or switch its default value to `true`.

Menus that need updating, off the top of my head: in cozy-sharing, cozy-scanner, cozy-notes.

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -54,6 +54,7 @@ const ActionMenuWrapper = ({
 
 const ActionMenu = ({
   children,
+  autoclose,
   className,
   onClose,
   placement,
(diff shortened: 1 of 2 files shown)


ptbrowne:
Mmh please do not introduce breaking change right now as we do not have a lot of bandwidth.

y-lohse:
It's not so much the migration that I'm worried about (as you say with a codemod it would be quick), it's the time it takes to test and publish everything, with the risk of missing one instance. I will do a round of mlgration next week if I can but I would feel a lot safer without a breaking change at first.

Crash--:
Let's go without the BC then

PR #1461, opened 2020-05-04
Border color as a variable

ptbrowne:
Not to forget, I did a quick search and replace to have a more semantic
variable name instead of var(--silver) everywhere.

I used `var(--dividerColor)`. It follows the same logic as material UI that uses theme.palette.divider for the border color of the Paper variant='outlined'.

Wondering what @joel-costa thinks about that.

### DIFF
--- a/stylus/components/accordion.styl
+++ b/stylus/components/accordion.styl
@@ -4,7 +4,7 @@ $accordion
     padding 0
 
 $accordion-item
-    border-bottom: rem(1) solid var(--silver)
+    border-bottom: rem(1) solid var(--dividerColor)
     font-size rem(16)
 
 $accordion-title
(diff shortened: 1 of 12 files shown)


nono:
I think it would be nice to merge this

ptbrowne:
I think we can use the `dividerColor` instead of the `borderColor` variable. I already started to use it at some place, outside this PR.

stylus/settings/palette.styl
```diff
@@ -225,4 +225,5 @@ html, .CozyTheme--normal
     --navTextColor var(--slateGrey)
     --navTextActiveColor var(--slateGrey)
     --navTextHoverColor var(--charcoalGrey)
+    --primaryContrastTextColor var(--white)
```
Crash--, line 228: why this addition ?

Crash-- (approved):
Just a small question 👍

stylus/settings/palette.styl
ptbrowne: Thanks, it was an error.

PR #1467, opened 2020-05-07
Improve theme support for Breadcrumb / Figure

ptbrowne:
- Use CSS variables for text color
- Breadcrumbs and Figure use those variables
- ThemeChooser component for easy switch between themes in styleguide
  (per component for now, would love to have it globally in the sidebar)

Preview: https://ptbrowne.github.io/cozy-ui/react/#!/Breadcrumbs

Edit: added the preview

### DIFF
--- a/docs/ThemeChooser.jsx
+++ b/docs/ThemeChooser.jsx
@@ -1,0 +1,19 @@
+import React, { useState } from 'react'
+import CozyTheme from '../react/CozyTheme'
+
+const ThemeChooser = ({ children }) => {
+  const [theme, setTheme] = useState('normal')
+  const handleChange = ev => setTheme(ev.target.value)
+  return (
+    <>
+      theme:{' '}
+      <select onChange={handleChange}>
+        <option value="normal">normal</option>
+        <option value="inverted">inverted</option>
+      </select>
+      <CozyTheme variant={theme}>{children}</CozyTheme>
+    </>
+  )
+}
+
+export default ThemeChooser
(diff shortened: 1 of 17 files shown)


docs/ThemeChooser.jsx
```diff
@@ -0,0 +1,19 @@
+import React, { useState } from 'react'
+import CozyTheme from '../react/CozyTheme'
+
+const ThemeChooser = ({ children }) => {
+  const [theme, setTheme] = useState('normal')
+  const handleChange = ev => setTheme(ev.target.value)
+  return (
+    <>
+      theme:{' '}
```
y-lohse, line 9: Why do we need the `{' '}`?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cozy-ui",
-  "version": "35.14.0-beta.1",
+  "version": "35.18.0-beta.1",
```
y-lohse, line 3: Is this intended?

docs/ThemeChooser.jsx
ptbrowne: because there is a space after colons in English ? And it's the way to do this in JSX when a space precedes a component.

ptbrowne:
Here is the preview yannick : https://ptbrowne.github.io/cozy-ui/react/#!/Breadcrumbs

package.json
ptbrowne: No.

PR #1468, opened 2020-05-11
Codemod to transform withBreakpoints into useBreakpoints

ptbrowne:
Uses the [helper from @cozy/codemods](https://github.com/cozy/cozy-libs/pull/1011) to transform `withBreakpoints` into `useBreakpoints`

Advantage of the hook :

- less nesting.
- less work at app initialisation (to wrap all the components)


react/CozyTheme/Readme.md
```diff
@@ -1,7 +1,7 @@
 An area of a page can have a different CozyTheme and components inside
 will be automatically styled.
 
-- At the moment only some Buttons support this.
+- At the moment only a few components support this (Buttons / Breadcrumbs / Figure).
```
y-lohse, line 4: Nit: I think this isn't meant to be part of this PR.

codemods/use-i18n-hook/index.js
```diff
@@ -1,8 +1,4 @@
-import makeUtils from '../utils'
-
-const prepend = (arr, item) => {
-  arr.splice(0, 0, item)
-}
+import { hocReplacer } from '@cozy/codemods'
```
y-lohse, line 1: 👍

react/CozyTheme/Readme.md
ptbrowne: Yep this was because this PR was based on the theme improvements one. Thanks.

PR #1469, opened 2020-05-13
feat: ActionMenuItem improvements

ptbrowne:
- When there was a text and a description in an ActionMenuItem, the app
had the responsibility to put a margin, this is now solved by using
a stack utility class on the action menu item body, which automatically
spaces its children correctly. The app can control this spacing through
the `contentSpacing` prop that has been added to `ActionMenuItem`. This prop
is set to `xs` by default (0.5rem of margin).

- Icons were in charcoalGrey instead of slateGrey. This is now managed via a CSS variable on the iconColor. The Icon takes its color from var(--iconColor), and actionMenuItem overrides `iconColor` with `actionMenuIconColor` for its children.

https://ptbrowne.github.io/cozy-ui/react/#/ActionMenu


y-lohse:
@joel-costa Why should the icon be a different color than the text? Is that a rule so should be applied elsewhere as well, or is it specific for the ActionMenu?


Code changes LGTM though!

joel-costa:
> @joel-costa Why should the icon be a different color than the text? Is that a rule so should be applied elsewhere as well, or is it specific for the ActionMenu?

It isn't specific for ActionMenu. Generally, whenever you associate an icon with a charcoal-grey text, the icon should be colored in slate-grey.

(Visually, a filled icon is much more dense than a text. If we applied the same color to them, the icon would appear darker. Our variation is supposed to correct this impression)

PR #1470, opened 2020-05-13
Nested select improvements

ptbrowne:
- Title size
- Ellipsis for nested select items
- Caption stays with normal font weight even if selected

https://ptbrowne.github.io/cozy-ui/react/#/NestedSelect

### DIFF
--- a/react/NestedSelect/styles.styl
+++ b/react/NestedSelect/styles.styl
@@ -13,6 +13,9 @@
 .Row__selected
     font-weight bold
 
+.Row__caption
+    font-weight normal
+
 .Modal__title
     border-bottom 1px solid var(--silver)
     margin-bottom 0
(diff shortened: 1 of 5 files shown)


react/NestedSelect/styles.styl
```diff
@@ -13,9 +15,17 @@
...
+    font-weight normal
+
 .Modal__title
     border-bottom 1px solid var(--silver)
     margin-bottom 0
+    padding-top 1rem
+    font-size $h2-font-size
+
+.Modal__title h2
+    font-size $h2-font-size
```
y-lohse, line 28: Why do we need to specify h2 font size twice here?

ptbrowne: No need, the one not in the h2 was extraneous ! Thanks

PR #1471, opened 2020-05-13
feat: Rounder style for action menu in bottom drawer

ptbrowne:
ActionMenu inside a BottomDrawer in latest designs is rounder and has no margin from the
screen.


Crash--:
why do we have a regression on the avatar? It seems that `with image instead of initials` is not right anymore

ptbrowne:
I suspect a network error on the CI side, it's been several days that the network at Travis seems unstable.

ptbrowne:
I've accepted on Argos even with the Avatar bug since I fix it here 79a52270.

PR #1474, opened 2020-05-15
Illus

ptbrowne:
- Sad cozy for use in Empty contexts
- Cozy text icon

I've seen in some designs from Joel that the text from the Cozy logo
was used in the bar, and added this icon thinking it would be useful
in the future. cc @joel-costa.


ptbrowne:
> Did you consider making sad-cozy an icon instead of an illustration?

I have already taken care of having the sad-coy icon and the cozy-text icon be "color less". But I don't think we should move them to the illustration section as they do not have the proper dimensions to be rightfully called an icon.
To me, the difference between illustrations and icons was not that the color was fixe, only that the icons were smaller and square, and to be used in "icon" context. I see that in the Readme for illustrations, it is written "color fixed", but we could change the text not to have that ?

y-lohse:
>I have already taken care of having the sad-coy icon and the cozy-text icon be "color less".

I had missed that, but are you sure? cozy-sad has a `fill="#297EF2"`on the `<path>` element.

> I see that in the Readme for illustrations, it is written "color fixed", but we could change the text not to have that ?

I wouldn't say so, to me illustrations are ready to use "out of the box", which does include size and color. We also have other cloud icons that are… well, icons, so presumably they fit in a square. Which makes me think cozy-sad could too?

ptbrowne:
> I had missed that, but are you sure?

🤦 I thought I had... I had forgotten to do that on the sad-cozy ! Thanks

I am putting both icons in the icons section.

PR #1477, opened 2020-05-18
refactor: Reorganize alphabetically

ptbrowne:
Sort components alphabetically inside sections.

For the future, I think we should start to think of how to reorganize our
components. I think our sections have grown a bit since the beginning
and that some sections like "Basics" have components that could
be moved in other sections like "Content".

I think this can be done in a separate issue, what do you think ?


y-lohse:
> I think this can be done in a separate issue, what do you think ?

Completely agree with this and the need to re-organize the docs!

PR #1482, opened 2020-06-09
Example with focus trap

ptbrowne:
We use focus trap in the cozy store. In the process of trying to
understand a crash, I developed an example and thought it could
be useful to keep it here. It could serve as a base to further
integrate focus trap so that it is automatically used by the
Modal buttons or content. What do you think ?

Info: The crash in question was in cozy-store where focus-trap
could not find any focusable element since the modal was loading
and a spinner was displayed. I have not found why it would not
crash bfore the update to cozy-ui 34 but I have seen that the
close button was focused by focus-trap-react before the update
to cozy-ui 34. After the update the close button was not
focused. Maybe the change to portal the content by default
was the triggerer for the error.

Edit: add link to example https://ptbrowne.github.io/cozy-ui/react/#!/Modal/35

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "enzyme-to-json": "3.4.0",
     "eslint": "4.19.1",
     "eslint-config-cozy-app": "0.5.1",
+    "focus-trap-react": "^6.0.0",
     "git-directory-deploy": "1.5.1",
     "http-server": "0.11.1",
     "husky": "0.14.3",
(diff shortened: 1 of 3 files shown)


y-lohse:
It looks like 88a8c3287952c8ed94f28a77a0d14da2a48e3836 accidentally landed in this PR :)
And could you deploy the example so we can play around with it? 

I definitely don't mind having the extra example in there.As to integrating it automatically… what are the pros and cons? I assume it's a good thing for accessibility?

ptbrowne:
I've rebased and deployed here : https://ptbrowne.github.io/cozy-ui/react/#!/Modal/35

PR #1498, opened 2020-07-15
Use relative paths everywhere

ptbrowne:
An import was still using cozy-ui/react directly. We must use relative paths when importing other componetns. Those relative paths are expanded into `cozy-ui/transpiled/react` paths at transpile time.

### DIFF
--- a/react/AppIcon/Preloader.js
+++ b/react/AppIcon/Preloader.js
@@ -42,7 +42,7 @@ const _getAppIconURL = (app, domain, secure) => {
 }
 
 const _getInstalledIconPath = app => {
-  return app.links && app.links.icon
+  return app && app.links && app.links.icon
 }
 
 const _getRegistryIconPath = app => {
@@ -51,6 +51,7 @@ const _getRegistryIconPath = app => {
   }
 
   return (
+    app &&
     app.latest_version &&
     app.latest_version.version &&
     `/registry/${app.slug}/${app.latest_version.version}/icon`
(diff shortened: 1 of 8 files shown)


y-lohse:
I understand that when importing `SmallAppitem` in our apps, we'd like it to import the transpiled version of `AppIcon`.

But I'm not sure this is the correct solution, as the transpiled files technically don't exist in the repo, they are only created at distribution time.

What happens if all components import transpiled files? Won't we run in a problem where we need to transpile files in a certain order for it to work?
i'm also unsure of the consequences of importing files `from 'cozy-ui'` while in `cozy-ui`..? It looks like a self dependency which I haven't seen before I think.

ptbrowne:
Yes, this was the wrong solution.

In fact we had already a solution for this, which was the babel plugin that expanded relative paths to 
full `cozy-ui/transpiled/react` path at transpile time (which is important for overrides).

For ex in SmallAppItem

Before transpiling

```

import AppIcon from '../../AppIcon'
import { translate } from '../../I18n'
import { getCurrentStatusLabel } from '../status'
import { AppDoctype } from '../../proptypes'
```

After trasnpiling

```
import AppIcon from "cozy-ui/transpiled/react/AppIcon";
import { translate } from "cozy-ui/transpiled/react/I18n";
import { getCurrentStatusLabel } from "cozy-ui/transpiled/react/AppSections/status";
import { AppDoctype } from "cozy-ui/transpiled/react/proptypes";
```


I reworked the commit to use relative paths.

PR #1505, opened 2020-07-17
Add Wizard from cozy-authentication

ptbrowne:
Also fixed some bad imports that I do not understand why this had
not been detected before in the CI :/

EDIT: Ah it seems it has been detected, I don't know why I haven't seen it :/ https://travis-ci.org/github/cozy/cozy-ui/builds/709074283#L1112

### DIFF
--- a/docs/styleguide.config.js
+++ b/docs/styleguide.config.js
@@ -88,7 +88,8 @@ module.exports = {
         '../react/Text/index.jsx',
         '../react/Tooltip/index.jsx',
         '../react/UnorderedList/index.jsx',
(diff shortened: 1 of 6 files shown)


ptbrowne:
> I wonder if eventually we can get rid of DualField and Select, and use more normal cozy-ui components instead.

👍

y-lohse:
As a side-note, this is tied to https://github.com/cozy/cozy-libs/pull/1039#issue-451141749 where you mention that the wizard was producing duplicate css utility classes. Do you remember what imports caused this?

PR #1508, opened 2020-07-20
feat: Add Switch from Material UI

ptbrowne:
- Implement a Mui theme for the Switch component (also supports inverted theme)
- Re-export Switch from cozy-ui
- Deprecate Toggle

The secondary color uses the var(--validColor) instead of the "normal" secondary color that we have in cozy-ui (a kind of purple) that is not really our secondary color.

@joel-costa I'd love your input on this.

Visible here : https://ptbrowne.github.io/cozy-ui/react/#!/Switch

## From styleguide

image

## From banks

image

image


ptbrowne:
> Thanks! Could you add a mention in the docs saying this is a MUI component, maybe with a link to the original docs?

Good idea, done in 246909b3.

> The disabled style looks a bit weird to me,

Yes, me too, but since it was like for a long time in Banks, I think it can be done in a next PR.

PR #1509, opened 2020-07-21
fix: Import Intersection-observer polyfill

Crash--:
Since LoadMore use Intersection Observer (and IO doesn't work on Safari), it seems to make sense that LoadMore requires the polyfill itself.

### DIFF
--- a/react/LoadMore/index.jsx
+++ b/react/LoadMore/index.jsx
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types'
 import Button from '../Button'
 import Spinner from '../Spinner'
 
+import 'intersection-observer' // polyfill for safari (mobile and desktop)
+
 const LoadMore = ({ fetchMore, label }) => {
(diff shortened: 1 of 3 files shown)


react/LoadMore/index.jsx
```diff
@@ -3,6 +3,8 @@ import PropTypes from 'prop-types'
 import Button from '../Button'
 import Spinner from '../Spinner'
 
+require('intersection-observer') // polyfill for safari (mobile and desktop)
```
ptbrowne, line 6: Here intersection-observer would not be tree shaken since require is used. This means that even if you do not use LoadMore, you would have the intersection observer loaded.

1. We could use `import` here (need to put in the sideEffects config that intersection-observer should not be tree shaken since it has side effects)
2. We could let the app import the `intersection-observer` and warn if the relevant APIs are not present

I think I'd prefer the import solution as it is the most transparent for users.

Crash--: Nice catch!

PR #1513, opened 2020-07-28
feat: Use validColor instead of emerald for alerts

ptbrowne:
- validColor instead of emerald
- errorColor instead of pomegranate

### DIFF
--- a/react/Alerter/Readme.md
+++ b/react/Alerter/Readme.md
@@ -3,11 +3,18 @@ Include the `<Alert />` component somewhere in your app, but only once. After th
 
 ```
 import Alerter from 'cozy-ui/transpiled/react/Alerter';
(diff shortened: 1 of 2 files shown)


react/Alerter/Readme.md
```diff
@@ -3,11 +3,18 @@ Include the `<Alert />` component somewhere in your app, but only once. After th
 
 ```
 import Alerter from 'cozy-ui/transpiled/react/Alerter';
-// we have to put the Alert only once in the app
+
+if (isTesting()) {
+  setTimeout(() => {
+    Alerter.success('This is a success alert')
+  }, 1)
```
Crash--, line 10: Since we've CSS transition (opacity) I think we'll have a few false positives doing this.

PR #1514, opened 2020-07-30
Dialog changes

y-lohse:
I'll probably have more changes to ExperimentalModal coming up soon, but here's a first batch : exporting more documents and improving the docs!

### DIFF
--- a/react/MuiCozyTheme/Dialog/DialogContent.jsx
+++ b/react/MuiCozyTheme/Dialog/DialogContent.jsx
@@ -1,0 +1,3 @@
+import DialogContent from '@material-ui/core/DialogContent'
+
+export default DialogContent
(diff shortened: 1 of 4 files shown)


react/MuiCozyTheme/Dialog/DialogContent.jsx
```diff
@@ -0,0 +1,3 @@
+import DialogContent from '@material-ui/core/DialogContent'
+
```
Crash--, line 2: because we plan to override them pretty soon?

y-lohse: No, just because we use them and we'd rather import everything from cozy-ui rather than mix cozy-ui and MUI imports.

PR #1518, opened 2020-08-05
fix: Use newer version of popper in action menu

y-lohse:
Our ActionMenu on desktop uses Popper to position itself. This Popper instance is exported by material-ui (v3) and is the v1 of Popper. We've had a few problems with our more advanced use cases, eg. menus inside modals. We fought it hard, even introducing a PopperContext to help with positioning. 

But even after all that: 

Capture d'écran 2020-08-03 10 16 32

And the Popper exported by MUI is restricted, we can't use fixed positioning with it for example. So I tried to use Popper v2 instead, and:

Capture d'écran 2020-08-03 10 20 02

Everything seems to Just Work™, without even using fixed positions, without messing with containerElements etc.

So yeah, I suggest we use the newer version of Popper for the menu as it seems simply better in every regard, and as far as I can tell there is no breaking change.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.4",
+    "@popperjs/core": "^2.4.4",
     "body-scroll-lock": "^2.5.8",
     "classnames": "^2.2.5",
     "date-fns": "^1.28.5",
@@ -143,6 +144,7 @@
     "react-hot-loader": "^4.3.11",
     "react-markdown": "^4.0.8",
     "react-pdf": "^4.0.5",
+    "react-popper": "^2.2.3",
     "react-select": "2.2.0",
     "react-swipeable-views": "0.13.3"
   },
--- a/react/ActionMenu/Readme.md
+++ b/react/ActionMenu/Readme.md
@@ -91,9 +91,10 @@ const hideMenu = () => setState({ menuDisplayed: false });
 </div>
 ```
 
-### Placement
+### Placement on desktop
 
-The `placement` and `anchorElRef` prop can be used to control the placement of the menu on desktop. `anchorElRef` should be a ref to a DOM element and not a react component.
+You can pass a reference to a custom DOM element through the `anchorElRef` prop to attach the menu to that element.
+We use [popper.js](https://popper.js.org/docs/v2/) under the hood. You can use the `popperOptions` prop to pass options to the popper.js instance. This lets you control things like placement relative to the anchor, positioning strategies and more — refer to the popper doc for all the details.
 
 ```
 import ActionMenu, { ActionMenuItem } from 'cozy-ui/transpiled/react/ActionMenu';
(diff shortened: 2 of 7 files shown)


Crash--:
I thought we were finally good on the menu component... 🤦 

I'm not sure that this is what we want in terme of UI? Don't we want the menu to go above modal and down?

package.json
```diff
@@ -133,6 +133,7 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.4",
+    "@popperjs/core": "^2.4.4",
```
JF-Cozy, line 136: Certaines lib ont une version fixée et d'autre non, il y a une certaine règle/logique à respecter ?

y-lohse:
> I'm not sure that this is what we want in terme of UI? Don't we want the menu to go above modal and down?

With the new popper we can do both, by switching the positioning strategy to `fixed`. I'll add a prop so it can be changed more easily.

That being said, on the sharing modal, I'd rather change it to the screenshot above, otherwise with a lot of contacts the menu ends up going under the screen limit.

package.json
y-lohse: Oui, mais la logique est différente sur UI par rapport aux apps. Sur UI, on veux laisser les `dependencies` en range et mettre les devdeps en fixe — pour éviter de publier des nouvelles versions de UI juste parce qu'une dep a sortie une nouvelle version mineure.

JF-Cozy: Ok compris merci

react/ActionMenu/index.jsx
```diff
@@ -1,54 +1,62 @@
...
 import styles from './styles.styl'
 import { Media, Bd, Img } from '../Media'
 import BottomDrawer from '../BottomDrawer'
 import withBreakpoints from '../helpers/withBreakpoints'
-import Popper from '@material-ui/core/Popper'
 import { getCssVariableValue } from '../utils/color'
-import PopperContainerContext from '../PopperContainerContext'
 import Radio from '../Radio'
 import { spacingProp } from '../Stack'
+import { usePopper } from 'react-popper'
```
Crash--, line 12: for information: 
BUNDLE SIZE
5.2kB

2kBMINIFIED + GZIPPED

https://bundlephobia.com/result?p=react-popper@2.2.3

react/ActionMenu/index.jsx
```diff
@@ -1,54 +1,62 @@
...
+    },
+    {
+      name: 'hide',
+      enabled: false
+    }
+  ]
+  const options = popperOptions || {
+    placement,
+    modifiers: preventOverflow ? undefined : normalOverflowModifiers
   }
```
Crash--, line 41: Why not logging the depreciation notice here instead of the render? (afraid that we flood our dev console )

react/ActionMenu/index.jsx
```diff
@@ -77,9 +94,9 @@ const ActionMenu = ({
         onClose={onClose}
         inline={shouldDisplayInline}
         anchorElRef={anchorElRef || containerRef}
-        containerElRef={containerElRef}
```
Crash--, line 80: should we log something like `containerElRef is not need anymore, you can remove it from your code`?

react/ActionMenu/index.jsx
```diff
@@ -1,54 +1,62 @@
...
   placement,
   preventOverflow,
   children
 }) => {
-  const getElementFromRefCallback = ref =>
-    useCallback(() => {
-      return ref ? ref.current : undefined
-    }, [ref])
+  const [popperElement, setPopperElement] = React.useState(null)
+  const referenceElement = anchorElRef ? anchorElRef.current : undefined
```
JF-Cozy, line 26: nit: est-ce qu'on ne préfèrerait pas utiliser 'null' plutôt que 'undefined' (même question sur tout la PR)? cf https://github.com/cozy/cozy-guidelines#return-value

react/ActionMenu/index.jsx
y-lohse: The logger only sends the message once :) https://github.com/cozy/cozy-ui/blob/master/react/helpers/createDepreciationLogger.js#L6 That said I now see I slightly mis-used it.

react/ActionMenu/index.jsx
y-lohse: Pour celui ci oui, bonne remarque. 
Pour l'autre je pense qu'undefined est mieux, parce que react-popper risque d'interpreter différent un `null` d'un `undefined` pour le tableau de modifiers. On veux vraiment passer une variable pas initialisé.

y-lohse:
Here's the preview https://yannick-lohse.fr/cozy-ui/react/#!/ActionMenu !

react/ActionMenu/Readme.md
```diff
@@ -114,65 +115,13 @@ const anchorRef = React.createRef();
...
-
-<>
-  <button onClick={() => setState({ shown: !state.shown })}>
-    Toggle custom fixed element
-  </button>
-  { state.shown && <MyFixedElement /> }
-</>
-```
-
 ### ActionMenu within modals
```
Crash--, line 125: it seems we still have a paragraph about `containerElRef`. Should we remove it too?

y-lohse: Oh snap good catch ! I left it there to check that there was nothing broken in the screenshots, but yeah it should be removed now 💯

y-lohse:
i'm merging this, but for the record one of the code snapshots for the action menu seems a little flaky, with some styling props being there/not there. I couldn't figure out what was wrong, but re-transpiling things seemed to help.

PR #1519, opened 2020-08-05
chore: Update travis node version

y-lohse:
Travis is still using node 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 matrix:
   fast_finish: true
 node_js:
-  - '8'
+  - '10'
 cache:
   yarn: true
   directories:


.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 matrix:
   fast_finish: true
 node_js:
-  - '8'
+  - '10'
```
Crash--, line 5: can you add an engine in the package.json ?

y-lohse: Hu? Why? `engines` is for the supported node versions of the module, I don't think it's what we want here. I also don't recall us doing that in other packages we maintain?

Crash--: You right, we don't do that elsewhere. Maybe we should add a note in the Development's section of the Readme to say we recommend the use of node 10 ?

PR #1520, opened 2020-08-12
feat: Ability to hide developer in AppTile

ptbrowne:
In some contexts, we need to be able to hide the developer from the AppTile.
I took the opportunity to refactor a bit since the SmallAppItem had been moved from cozy-store as is.
Extracting Tile provides the opportunity for reuse. 

- [x] Renamed SmallAppItem to AppTile
- [x] Extracted the Tile component from AppTile
- [x] Added a showDeveloper prop to AppTile that is defaulted to true

This will help us override this property in contexts where
we do not want to show the developer by default.

I simplified the CSS of the Tile by using box-sizing: border-box which removes the need to remove padding and border-width from flex-basis. It results in small differences on Argos :

- in mobile, the tile width is fixed since it was previously larger than its container
- on desktop the difference is subtler: I think that its the 1px removal from the padding that was surely done to account for the border width, that was actually wrong since there are two borders, it should have been 2px)


react/AppSections/components/__snapshots__/AppsSection.spec.jsx.snap
```diff
@@ -1,285 +1,34 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
```
JF-Cozy, line 1: nit: peut-être qu'on pourrait utiliser un expected = [CONTENT_TO_TEST] dans le code de test, ça permettrait d'enlever le système de snaphot ?

ptbrowne: Qu'est ce que tu y verrais comme avantage ? Je trouve ca justement pratique de ne pas avoir à modifier le contenu du fichier de test lors de changements, j'ai juste à mettre à jour le snapshot.

JF-Cozy: J'ai dû mal comprendre, il me semblait qu'on essayait de sortir des snapshots si possible. Je pense qu'il y a 2 approches, c'est subjectif mais au contraire je trouve plus pratique et plus lisible de maintenir un fichier de test qu'un snapshot.

ptbrowne: OK je note, je suis d'accord que c'est subjectif :D A mon avis le "sortir des snapshots", c'était plutôt ne pas forcément snapshotter tout un composant pour que les snapshots contiennent le contenu minimum.

JF-Cozy: Oui c'est déjà pas la même approche en effet ! 👍

react/AppSections/components/AppTile.styl
```diff
@@ -0,0 +1,76 @@
...
+  white-space nowrap
+  overflow hidden
+
+.AppTile-title
+  height 1.25rem
+  color var(--black)
+
+.AppTile-developer
+  height 1rem
+  font-size 12px
```
Crash--, line 37: `px`? Shouldn't we use `rem` instead? Also, we have a few stylus variables for font-size, we should use them (https://github.com/cozy/cozy-ui/blob/master/stylus/settings/typography.styl#L14)

react/Tile/styles.styl
```diff
@@ -0,0 +1,45 @@
...
+.Tile
+  box-sizing border-box
+  position relative
+  display flex
+  flex-direction column
+  flex $tileSize 0 0
+  height $tileSize
+  align-items center
+  background var(--white)
+  border-radius 4px
```
Crash--, line 13: rem?

react/Tile/styles.styl
```diff
@@ -0,0 +1,45 @@
...
+  padding .375rem
+  margin-bottom 1rem
+  margin-right .75rem
+  overflow hidden
+  transition .15s ease all
+
+.Tile:hover,
+.Tile:active,
+.Tile:focus
+  box-shadow 0 4px 12px 0 #00000020
```
Crash--, line 24: Should we use a css var?

react/Tile/styles.styl
```diff
@@ -0,0 +1,45 @@
...
+  cursor pointer
+  transform scale(1.1)
+
+@media (min-width 1041px)
```
Crash--, line 28: Is it a new media query? I knew 1023 ( or 1200 but not 1041). Also still px

react/Tile/styles.styl
```diff
@@ -0,0 +1,45 @@
...
+  .Tile
+    margin-right 1rem
+
+@media (max-width 48rem)
```
Crash--, line 32: small-screen

react/Tile/styles.styl
ptbrowne: Not sure that this is relevant for border-radiuses.

react/Tile/styles.styl
ptbrowne: Done

react/Tile/styles.styl
ptbrowne: I removed this small tweak since this breakpoint is not standard.

react/Tile/styles.styl
ptbrowne: Done

react/AppSections/components/AppTile.styl
ptbrowne: Ah yes true. It's not code that has been written in this PR, just code that has been moved, this is why I did not refactor much more that the split between Tile and AppTile.

Anyway, thanks for the suggestions : https://github.com/cozy/cozy-ui/pull/1521

PR #1522, opened 2020-08-14
feat: Use cozy-interapp to provide intents

ptbrowne:
The IntentIframe now uses cozy-interapp and does not need
cozy-client-js to work.

### DIFF
--- a/react/IntentIframe/IntentIframe.jsx
+++ b/react/IntentIframe/IntentIframe.jsx
@@ -1,6 +1,8 @@
-/* global cozy */
 import React from 'react'
 import PropTypes from 'prop-types'
+import { Intents } from 'cozy-interapp'
+import { withClient } from 'cozy-client'
+
 import Spinner from '../Spinner'
 import styles from './styles.styl'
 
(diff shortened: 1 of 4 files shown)


react/IntentIframe/IntentIframe.jsx
```diff
@@ -17,7 +19,13 @@ class IntentIframe extends React.Component {
   componentDidMount() {
     const { action, data, type, onCancel, onError, onTerminate } = this.props
```
JF-Cozy, line 20: Maybe we should add `client` in those props

react/IntentIframe/IntentIframe.jsx
```diff
@@ -17,7 +19,13 @@ class IntentIframe extends React.Component {
...
 
-    const create = this.props.create || cozy.client.intents.create
+    let create
+    if (this.props.create) {
+      create = this.props.create
+    } else {
+      const intents = new Intents({ client: this.props.client })
```
JF-Cozy, line 26: If `client` is in DidMount props, maybe we should write `new Intents({ client })` ?

react/IntentIframe/IntentIframe.jsx
ptbrowne: Yes, the purpose of withClient is to add the client in the props of the underlying component. You're right that it's better to destructure the client at the top of the function 👍

PR #1523, opened 2020-08-20
Tweak ListSubHeader styles

ptbrowne:
Tweak ListSubHeader theme to match recent designs.
Use ListSubHeader in ContactList.


react/CompositeRow/index.jsx
```diff
@@ -77,7 +81,11 @@ CompositeRow.propTypes = {
   /* Element(s) to the show to the right of the CompositeRow */
   right: PropTypes.node,
   /** Row height will be fixed to 48px */
-  dense: PropTypes.bool
+  dense: PropTypes.bool,
+  /** Two adjacent CompositeRows will have a border between them */
+  border: PropTypes.bool,
```
Crash--, line 86: I'm wondering if we should not use a `list` to display several `compositeRows`. It'll be the `list` goal to add border or not to its children and also deal with the hover state. 

Also for our `List` we already have a `sub header` component (https://docs.cozy.io/cozy-ui/react/#/Material-UI?id=list-1). Can't we reuse the same thing?

ptbrowne: Good, thanks, I did not know this List component. I'll remove the border handling from the CompositeRow and also update the sub header to match the recent designs.

ptbrowne: I removed the NavRow and will try to use directly the ListItems in the app. So this PR consists only of tweaking the ListSubHeader styles and using it in the contact list.

Crash-- (approved):
👍

ptbrowne:
The red build is due to the snapshot with data-popper. Not related to this PR.

PR #1525, opened 2020-08-24
feat: Add font-weight to styleguide

JF-Cozy:
- Styleguide now have font-weight utilities (normal and bold)

### DIFF
--- a/stylus/utilities/text.styl
+++ b/stylus/utilities/text.styl
@@ -174,11 +174,25 @@ fontstyle = {
     'fs-normal': 'fs-normal',
     'fs-italic': 'fs-italic'
 }
-
 if cssmodules == true
     cssModulesUtils(fontstyle, breakpoints)
 else
     nativeUtils(fontstyle, breakpoints)
+
+fw-normal()
+    font-weight normal
+
+fw-bold()
+    font-weight bold
+
+fontweight = {
+    'fw-normal': 'fw-normal',
+    'fw-bold': 'fw-bold'
+}
+if cssmodules == true
+    cssModulesUtils(fontweight, breakpoints)
+else
+    nativeUtils(fontweight, breakpoints)
 // @stylint on
 
 // Global classes
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -890,6 +890,23 @@ Display an chip that represents complex identity
 */
 
 /*
+ Font-weight
+
+ Sets a font-weight to a text<br>
+[Can be used with breakpoints modifier suffixes (`-t`,`-s`,`-m`)](section-settings.html#kssref-settings-breakpoints)
+
+ Markup:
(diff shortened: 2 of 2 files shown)


react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -6,7 +6,7 @@ exports[`ActionMenu should render examples: ActionMenu 1`] = `
         <use xlink:href=\\"#bottom\\"></use>
       </svg></button>
     <div>
-      <div style=\\"position: absolute; left: 0px; top: 0px; z-index: #fff; transform: translate(0px, 0px);\\" data-popper-placement=\\"bottom-start\\">
+      <div style=\\"position: absolute; left: 0px; top: 0px; z-index: #fff;\\">
```
Crash--, line 9: there is a pending issue with the popper component. Don't try to fix it in your PR, even if the CI is red

stylus/utilities/text.styl
```diff
@@ -174,11 +174,29 @@ fontstyle = {
...
 if cssmodules == true
     cssModulesUtils(fontstyle, breakpoints)
 else
     nativeUtils(fontstyle, breakpoints)
+
+fw-normal()
+    font-weight normal
+
+fw-bold()
+    font-weight bold
```
Crash--, line 186: We have a `Bold` Component (https://github.com/cozy/cozy-ui/blob/master/react/Text/index.jsx#L59). Shouldn't we be coherent with it? ping @y-lohse

y-lohse: Why do you think it's incoherent? They both use a font-weight of `bold` 🤔

stylus/utilities/text.styl
```diff
@@ -174,11 +174,29 @@ fontstyle = {
...
+
+fw-light()
+    font-weight lighter
```
y-lohse, line 189: What is the use case fo this? As far as I know we don't use that font-weight. We also don't load that font weight from lato iirc, so it would rely on native interpolation  which is a bit meh.

stylus/utilities/text.styl
Crash--: Because `<Bold>` uses `title-h4`. This class : 
- Sets font-size 
- deals with small-screen 
- sets text-color 

So I'm wondering if `u-bold` should not do the same thing?

y-lohse: Ah, I see. I think it's a mistake for bold to use `title-h4` in the first place though. It shouldn't be coupled to a title class, but more importantly, utility classes are not components. It's ok for a component to set multiple css rules, but the idea of utility-classes is that they only do one thing and that you combine them as you need. So I think it's right that `Bold` component and `u-fw-bold` utility classes don't do the same thing.

Crash--: Thanks!

JF-Cozy:
I added a commit for Bold component @y-lohse . Now Bold don't change font-size anymore, do you think it's an issue ?

Crash--:
It's at least a beaking change.

y-lohse (approved):
Thank you!

PR #1526, opened 2020-08-24
fix: Show border even if item is not first and has 2ndary action

ptbrowne:
If the last child of a ListItem is a SecondaryAction, it is wrapped in a
`container`. This is why we have to have the border-top on the `container`
otherwise, when wrapped, the root is the first child, so the border is not
displayed.


JF-Cozy (approved):
LGTM 👍

ptbrowne:
Failing because of the snapshot of action menu : https://travis-ci.org/github/cozy/cozy-ui/builds/720602878#L1050

Merging

PR #1529, opened 2020-08-24
fix: ActionMenu proptypes

### LINKED ISSUE #1524: Warning for ActionMenu using placement props
There is a deprecation warning when using the placement
props for the ActionMenu component.

```
<ActionMenu placement /> is deprecated, use <ActionMenu popperOptions={{ placement }} /> instead
```

The problem is that there is defaultProps.placement which
means the warning is there even if the user has not used
the placement props.

Can you take a look @y-lhose ?

y-lohse:
Updates the proptypes of action menu and fixes #1524
.

### DIFF
--- a/test/testFromStyleguidist.js
+++ b/test/testFromStyleguidist.js
@@ -6,7 +6,14 @@ import pretty from 'pretty'
 
 import * as content from '../docs/fixtures/content'
 
-const testFromStyleguidist = (name, markdown, require) => {
+const sleep = ms => new Promise(resolve => setTimeout(resolve, ms))
+
+const testFromStyleguidist = (
+  name,
+  markdown,
+  require,
+  { delay } = { delay: 0 }
+) => {
   const evalInContext = a =>
     // eslint-disable-next-line no-new-func
     new Function(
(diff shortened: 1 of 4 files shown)


JF-Cozy (approved):
LGTM 👍

Crash-- (approved):

test/testFromStyleguidist.js
```diff
@@ -49,11 +49,12 @@ const testFromStyleguidist = (name, markdown, require) => {
         })
         done()
       }
-      codes.forEach(code => {
+      codes.forEach(async code => {
         const root = mount(
           <Preview code={code.content} evalInContext={evalInContext} />,
           options
         )
+        await sleep(100)
```
ptbrowne, line 57: Can you add a comment telling us why is this necessary ? I am a bit unconfortable to add this delay to all tests when it is necessary only for one. Is it not possible to specify the delay only for ActionMenu ?

y-lohse: Sorry, I wasn't done yet! Since the test seems more flaky on the CI than on my machine, I first wanted to confirm that the sleep *really* works, including on the CI.

But yes, I wouldn't like it to stay that way either. I haven't found a way to *not* use a timer at all, but running it only for components that need it as you suggest is a good idea, thanks 👍

ptbrowne: Have you understood why the test is flaky, checking the code in material ui ?

y-lohse: > Have you understood why the test is flaky, checking the code in material ui ?

The problem comes from popper-core, but no I haven't dug into the source to see what is going on. All I can say for sure is that the return value of `usePopper` changes multiples times, and it seems to come from popper-core and not react-popper itself.

y-lohse:
Ok, the tests seem stable now, by adding a small delay on the ActionMenu test!

ptbrowne:
Thanks 👍

ptbrowne (approved):

PR #1530, opened 2020-08-28
feat: Add u-lowercase class

doubleface:
This is needed in harvest for
https://github.com/cozy/cozy-libs/pull/1066

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -1002,6 +1002,8 @@ Display an chip that represents complex identity
  <h4 class="u-title-h4">Heading 4</h4>
  <p class="u-text">Basic text</p>
  <p class="u-caption">Caption text</p>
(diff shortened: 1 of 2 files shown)


JF-Cozy:
I think the doc's missing ?

JF-Cozy:
Maybe we can refactor a bit all text-decoration to use something like `u-td-uppercase` like with `u-ta-` or `u-fw-` ?

Crash-- (approved):

doubleface:
> I think the doc is missing ?

Right, here it is

doubleface:
> Maybe we can refactor a bit all text-decoration to use something like `u-td-uppercase` like with `u-ta-` or `u-fw-` ?

Since u-uppercase already existed and since I do not know if it is widely used or not, I prefer not to change this for my first cozy-ui PR (I did some in 2015 but it does not count :-) )

PR #1532, opened 2020-08-31
feat: Ability to add classNames to primaryText and secondaryText

ptbrowne:
It is useful to be able to add u-error on list item components.

In the usecase section, I tweaked the example to have a usecase
that is nearer to the real usecase of the navigation menu.


ptbrowne:
Le build a failé suite à un glitch lors de l'upload des screenshots chez Argos. Les screenshots ont quand même était uploadés.

Crash-- (approved):

PR #1537, opened 2020-09-02
feat: Styleguide example for navigation list + tabs

ptbrowne:
Add combination of nav list + tabs.

Visible on https://ptbrowne.github.io/cozy-ui/react/#!/Tab


react/Tabs/Readme.md
```diff
@@ -1,67 +1,82 @@
+Content needs to be separated into sections and accessed via a single content area using a flat navigation structure that does not refresh the page when selected.
+
+
 ```
+import MuiCozyTheme from 'cozy-ui/transpiled/react/MuiCozyTheme';
 import { Tabs, TabList, Tab, TabPanels, TabPanel } from 'cozy-ui/transpiled/react/Tabs';
 import Icon from 'cozy-ui/transpiled/react/Icon';
+import ListSubHeader from '../MuiCozyTheme/ListSubheader'
```
Crash--, line 8: no need for this one

Crash-- (commented):
I really like the idea, but I don't like to not be able to copy / paste the exemple and run it directly. 

I don't know if we can do something about it? For instance, if we import a component from '../docs/organisms' in a Readme, maybe we can display its source code when we 'view code' on the docs.

y-lohse:
> I don't know if we can do something about it? For instance, if we import a component from '../docs/organisms' in a Readme, maybe we can display its source code when we 'view code' on the docs.

I think typically in this example, it would add quite a lot of code to the example that would make the component harder to understand. Is there really a use case for straight up pasting an example somewhere else without making any change to it?

Crash--:
> I think typically in this example, it would add quite a lot of code to the example that would make the component harder to understand. Is there really a use case for straight up pasting an example somewhere else without making any change to it?


This is what I do when I work with M-UI. They have a kind of resume
Capture d’écran 2020-09-03 à 10 23 35

but if you take a look at the code, you have the full code you can copy / paste and run.

Capture d’écran 2020-09-03 à 10 24 44

Maybe we should just publish the organism component and import it from `cozy-ui/docs/...` instead of relative path and it'll solve my issue.

y-lohse:
> Maybe we should just publish the organism component and import it from cozy-ui/docs/... instead of relative path and it'll solve my issue.

Nah, we're [not publishing the docs folder on npm](https://unpkg.com/browse/cozy-ui@35.43.0/), so it wouldn't work 😐

Crash--:
This is my point, why not publishing it?

ptbrowne:
I've reworked the component and examples to have directly usable NavigationList and NavigationListSection components.

I agree that copy/pasting from the example would be great but I do not see how it's possible easily here, since I want the NavigationList example to be shown in several places.

ptbrowne:
1. To fix the problem of having to reuse the NavigationList example in List and Tabs

I've moved the NavigationListExample out of Tab and List and I've added a link from Tab and List to the NavigationList example, this way a user can find out that the NavigationList uses components from List, and can be used with Tabs.

2. I also needed the example to be both copy pastable, and not duplicated, inside the NavigationList example (which presents usecases with or without tabs, both with the same navigation list). Since variables cannot be shared between examples, I think the best is to have 1 code example presenting the two usecases.

Crash-- (approved):

ptbrowne:
I am merging since I need it. @y-lohse if you have any qualms, do not hesitate :)

PR #1539, opened 2020-09-03
feat: {primary,secondary}Text can be React nodes

ptbrowne:
PropTypes.string is often too restrictive. Here for example, we
wanted to show an icon next to the text, hence we used a fragment
containing <Icon /> and the text string as secondaryText prop.


y-lohse (approved):
Thanks!

PR #1544, opened 2020-09-07
Semantic color for infos

### LINKED ISSUE #1543: Theme the Infos component
Right now, the Infos component, especially its danger variant
uses hardcoded colors like pomegranate. It makes it
impossible to theme.

We should use u-error, u-valid for those usecases. We may need
an u-info to support the info variant.

The chablis color should be replaced by something
more semantic like : u-bg-error-light

ptbrowne:
It replaces hardcoded colors with semantic ones.

Fixes #1543

### DIFF
--- a/react/Infos/styles.styl
+++ b/react/Infos/styles.styl
@@ -12,13 +12,13 @@
         padding spacing_values.s spacing_values.m
 
     &.Infos--primary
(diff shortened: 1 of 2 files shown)


y-lohse (approved):
All good, thank you!

PR #1545, opened 2020-09-08
feat: Add utility classes for semantic backgrounds

ptbrowne:
Using semantic utility classes that use semantic
CSS variables enables us to build themes.

### DIFF
--- a/stylus/utilities/bgcolor.styl
+++ b/stylus/utilities/bgcolor.styl
@@ -7,16 +7,3 @@ for color in keys(colors)
         background-color: colors[color] !important // @stylint ignore
 
     global('.u-bg-' + color, '$bg-color-' + color)
-
-$bg-primary-light
-    background-color: var(--primaryColorLightBackground) !important // @stylint ignore
-
-$bg-error-light
-    background-color: var(--errorLightBackground) !important // @stylint ignore
-
-$bg-neutral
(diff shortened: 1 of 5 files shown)


stylus/utilities/bgcolor.styl
```diff
@@ -7,3 +7,16 @@ for color in keys(colors)
         background-color: colors[color] !important // @stylint ignore
 
     global('.u-bg-' + color, '$bg-color-' + color)
+
+$bg-primary-light
+    background-color: var(--primaryColorLightBackground) !important // @stylint ignore
```
JF-Cozy, line 12: I'm wondering about those `!important` everywhere. Is this the right approach, or should we just clean it all up and make it work without the `!important` (if it's even possible, I don't know if it's ever been debated)?

JF-Cozy:
Isn't the documentation missing?

stylus/utilities/bgcolor.styl
ptbrowne: Utility classes have !important and it is not a problem since they should override everything.

I think we could debate this if it is a problem for you but I do not think it is the role of this PR.

ptbrowne:
Thanks, the documentation was indeed missing. I also saw that I could leverage the palette.json to automatically generate the utility classes.

JF-Cozy (approved):

PR #1548, opened 2020-09-09
Removal of warnings in tests.

ptbrowne:
Removal of warnings in tests.

The trigger for this PR was the PropTypes.oneOf warnings that broke tests in Banks (the tests in banks are configured to break on console.warn).


y-lohse (approved):
Thanks, LGTM!

PR #1550, opened 2020-09-09
fix: Prevent badge warnings from MUI

y-lohse:
When using `InfosBadge` or `Ghostbadge`, we got a warning `Material-UI: the key `badge` provided to the classes property is not implemented in Component.`. This is because even though `badge` is a valid class name for [MUI Badge](https://v3.material-ui.com/api/badge/), it wasn't declared in the cozy-ui export.

I also fixed a problem where custom classes where not spread in the right place.

### DIFF
--- a/react/Badge/index.jsx
+++ b/react/Badge/index.jsx
@@ -13,6 +13,8 @@ const MEDIUM_DOT = '.75rem'
 const SMALL_DOT = '.625rem'
 
 const customStyles = theme => ({
+  badge: {},
+  root: {},
   top: {
     top: '16%'
   },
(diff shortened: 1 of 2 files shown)


ptbrowne:
How come we did not see this on Argos ? There were only warnings without any incidence ?

y-lohse:
>  There were only warnings without any incidence ? 

Yes that's correct.

> Should we have an example with customClasses ?

If it's for learning how to do it, there are some examples in the MUI docs.If it's for the screenshots… we could, but we don't have a use case at the moment, it's just something I noticed while addressing the `badge` warning.

ptbrowne:
OK fine by me, thanks :)

ptbrowne (approved):

PR #1560, opened 2020-09-17
Extract Tile components from AppTile

ptbrowne:
For the Tile pattern to be usable elsewhere, it needs to have
ready to use components. This is why I extracted TileIcon, TileTitle,
TileDescription and TileFooter from the AppTile.

I also fixed an ellipsis bug (app title would be wider than container).


Crash-- (approved):

react/Tile/styles.styl
```diff
@@ -43,3 +43,70 @@ $tileMobileHeight = 3.75rem
...
+    margin-left .5rem
+    justify-content center
+    text-align left
+    /* 4rem for icon and margins */
+    width calc(100% - 4rem)
+
+  .Tile-title
+    height 1rem
+    font-weight normal
+    font-size 12px
```
JF-Cozy, line 103: shouldn't we use 'rem' instead of 'px' ?

ptbrowne: This file was only moved, I did not touch its content, but you're right that it's better to use rem as a general rule. Even better, we have typography variables, I made the change in my last commit, thanks.

https://github.com/cozy/cozy-ui/pull/1560/commits/2c638b18512761a82e5fbfdc9d9c608983165fd5

PR #1564, opened 2020-09-21
Button full width does not reset margins

### LINKED ISSUE #1561: Button props extension full should not set margin to 0
Prop `extension="full"` of `<Button />` component (https://docs.cozy.io/cozy-ui/react/#/Basics?id=button) set `margin: 0` in CSS, which replaces the behavior, for example, of the `<Stack />` component which defines vertical margins.

It's probably not a desired behavior.

ptbrowne:
Make button full width work with Stack component. Fixes #1561.

### DIFF
--- a/react/Button/Readme.md
+++ b/react/Button/Readme.md
@@ -49,12 +49,13 @@ import Button from 'cozy-ui/transpiled/react/Button';
 
 ```
 import Button from 'cozy-ui/transpiled/react/Button';
+import Stack from 'cozy-ui/transpiled/react/Stack';
 
-<div>
-  <p><Button label='Normal'/></p>
-  <p><Button extension="narrow" label='N…'/></p>
-  <p><Button extension="full" label='Full width'/></p>
-</div>
+<Stack>
+  <Button className='u-db' label='Normal'/>
+  <Button className='u-db' extension="narrow" label='N…'/>
+  <Button className='u-db' extension="full" label='Full width'/>
+</Stack>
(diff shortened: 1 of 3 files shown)


Crash--:
Wondering if it's not a BC since it can have an hight impact if the consuming App is based on the fact that margin is set to 0, no?

ptbrowne:
I am not sure it would be high impact, but I can put a breaking change mention here.

JF-Cozy:
I kind of see it the other way around. In case we're not in `<Stack>`, we won't have a null vertical alignment. That's the breaking change, isn't it?

ptbrowne:
I do not understand your question. Before, inside a Stack, the full width button does not respect what is told by its parent, since it resets the `margin-{top,bottom}: 0` whereas it only needs to reset the horizontal margins.

See https://github.com/cozy/cozy-ui/blob/master/stylus/components/button.styl#L72 where the button already sets the vertical margins to 0.

The breaking change is that now, if in a Stack, the full-width button will work "as expected", it will follow the margins given by its parent.

JF-Cozy:
Okay, I didn't have the button case in mind.

JF-Cozy (approved):

Crash--:
> The breaking change is that now, if in a Stack, the full-width button will work "as expected", it will follow the margins given by its parent.

Not just `Stack` Component. 

It'll close https://github.com/cozy/cozy-ui/issues/797

PR #1565, opened 2020-09-21
Add yarn watch

JF-Cozy:
- Add yarn watch

```
yarn build:doc:react
```

- [x] Did you think of ARIA attributes if you are coding new components ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,8 @@
     "watch:css": "yarn build:css --watch",
     "prewatch:doc:react": "yarn sprite",
     "watch:doc:react": "env BUILD_ENV=watch-styleguidist styleguidist server --config docs/styleguide.config.js",
-    "watch:doc:kss": "nodemon --ext styl,md --watch stylus --exec 'yarn build:doc:kss && http-server build/styleguide -p 4242'"
+    "watch:doc:kss": "nodemon --ext styl,md --watch stylus --exec 'yarn build:doc:kss && http-server build/styleguide -p 4242'",
+    "watch": "yarn transpile:react --watch"
   },
   "sideEffects": [
     "*.css",


react/ActionMenu/index.jsx
```diff
@@ -174,11 +176,12 @@ const ActionMenuItem = ({
   className,
   contentSpacing
 }) => {
+  const hasCaption = Array.isArray(children) && children.some(child => get(child, 'type.displayName', '').includes('Caption'))
```
ptbrowne, line 179: This seems too magical for me, and it links the two components (ActionMenuItem and Caption) too much I think. I ask myself what is the general rule here, and also what happens when we migrate to new text components. I think it'd be preferrable to let this responsibility to the app, by providing an align prop on the ActionMenuItem.

react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -8,7 +8,7 @@ exports[`ActionMenu should render examples: ActionMenu 1`] = `
     <div>
       <div style=\\"position: absolute; left: 0px; top: 0px; z-index: #fff; transform: translate(0px, 0px);\\" data-popper-reference-hidden=\\"true\\" data-popper-escaped=\\"true\\" data-popper-placement=\\"bottom-start\\">
         <div class=\\"styles__c-actionmenu___22Fp1 styles__c-actionmenu--inline___1SXZa\\">
-          <div class=\\"styles__media___cSJMp styles__media--top___K9w0I styles__c-actionmenu-item___gODqd\\">
+          <div class=\\"styles__media___cSJMp undefined styles__c-actionmenu-item___gODqd\\">
```
ptbrowne, line 11: Careful, I do not think that setting align to "middle" is allowed, this is why there are "undefined" classes now.

react/ActionMenu/index.jsx
JF-Cozy: The idea is to have a default behavior that vertically aligns the icon, except for cases with a caption, we then position the icon at the top.

react/__snapshots__/examples.spec.jsx.snap
JF-Cozy: Wow, I read it the other way around when I reread it...

For the 'middle' case, according to the doc https://docs.cozy.io/cozy-ui/react/#/Content?id=media it's supposed to work, right ? (see `align=middle (default)`)

react/ActionMenu/index.jsx
ptbrowne: I understand the intention but I think that it may not be desirable to add this kind of coupling between Caption and ActionMenuItem.


* What happens when a new component other than Caption needs to have capability ? Should we keep adding new components in the list ?
* What happens if the Caption is wrapped into `<span></span>` to add for example another icon ? The check won't work 
* Is it really necessary to introduce a hard coded rule like that when it can be done by the app on a case by case basis ?

JF-Cozy: Hmmm... you're right it's getting complicated, maybe we need to take a little more time to list all the cases and make a more complex rule? Or we can simply say that for the moment it's not relevant because there are no concrete cases, we're here in anticipation.

There are also questions with texts that are too long, and texts of different lengths in different languages... in this case, do you think we can manage on the application side?

react/__snapshots__/examples.spec.jsx.snap
JF-Cozy: fixed here https://github.com/cozy/cozy-ui/pull/1571

react/ActionMenu/index.jsx
JF-Cozy: No more magic, this is not needed anymore

ptbrowne (approved):

PR #1572, opened 2020-09-22
AppTile improvements

### LINKED ISSUE #1569: AppTile should be exported with withLocales
AppTile uses locales that should not be part of the app
locales, but should be "embedded" in the component.
Otherwise the app has to do that itself.

ptbrowne:
- Exported with withLocales (fix #1569)
- Shown in styleguide
- Tile can be shown in a non flex container

### DIFF
--- a/react/AppTile/Readme.md
+++ b/react/AppTile/Readme.md
@@ -1,0 +1,16 @@
+Component used to show an app status, can be used in
+a list of apps for example in the Store.
+
+```
+import AppTile from '.'
+import mockApps from '../AppSections/_mockApps'
+import { I18n } from '../I18n'
+
+const locale = {}
+
+const app = mockApps[0];
+
+<I18n dictRequire={lang => locale} lang="en">
+  <AppTile app={app} />
+</I18n>
+```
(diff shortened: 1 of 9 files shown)


react/AppSections/shared-locales.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * This file is present since babel inlines the content of JSON
+ * files where they are imported. Since we need to share
+ * those locales between Sections and AppTile, this file is
```
Crash--, line 4:
```suggestion
 * those locales between AppSections and AppTile, this file is
```

react/AppTile/Readme.md
```diff
@@ -0,0 +1,16 @@
+Component used to show an app status, can be used in
```
Crash--, line 1:
```suggestion
Component used to show an app statu. Can be used in
```

Crash-- (approved):
Great! 

Wondering if we should not create an `App` folder with AppSections / AppTile and so on since we rely on the same langage files.

JF-Cozy (approved):

ptbrowne:
> Wondering if we should not create an App folder with AppSections / AppTile and so on since we rely on the same langage files.

I wonder if it would not better instead that they rely on different locale files 🤔 I think it's better if they do not rely on the same file in fact... I am going to split them. This way, no weird shared-locales.js file and through tree shaking, we embark only the locales that we need.

PR #1573, opened 2020-09-24
fix: ActionMenu display on small screen with safe area

Crash--:
We set the container to bottom 0, even if there is a
safearea in order to display the background, but then
we add the padding-bottom to the actionmenu in order
to follow the safe area.

before: 
Capture d’écran 2020-09-24 à 18 42 56

after:

Capture d’écran 2020-09-24 à 18 40 18

https://crash--.github.io/cozy-ui/react/#/Navigation?id=actionmenu


ptbrowne:
Cool :+1:

ptbrowne (approved):

PR #1575, opened 2020-09-25
Better behavior for CollectionField

ptbrowne:
Following @joel-costa's remarks during validation phase, the
CollectionField now supports the use of an imperative callback called when
a component is added. This is useful for example for the contact picker that should be
shown directly just after clicking "Add".

See the CollectionField/Readme.md for more information (see the onAddField prop).

[new behavior](https://ptbrowne.github.io/cozy-ui/react/#!/CollectionField) to compare with [current behavior](https://docs.cozy.io/cozy-ui/react/#!/CollectionField)

### DIFF
--- a/react/ContactPicker/index.jsx
+++ b/react/ContactPicker/index.jsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react'
+import React from 'react'
+import PropTypes from 'prop-types'
 import ContactsListModal from '../ContactsListModal'
 import styles from './styles.styl'
 import cx from 'classnames'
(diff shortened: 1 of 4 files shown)


joel-costa:
C'est top :) 
Par contre, est-ce qu'on pourrait enlever le bouton de suppression (la petite croix à droite du champs) quand on est en "nullData" ?

ptbrowne:
We talked about that in Zoom with @joel-costa and decided it could be done another time. I've logged https://github.com/cozy/cozy-ui/issues/1581.

JF-Cozy (commented):
LGTM :)

PR #1579, opened 2020-09-28
Screenshot stack page

ptbrowne:
Here we add an example of the kind of page served by the stack. An express server is
used to be able to switch the "theme" of the app via a `theme` query string.
This way, we have screenshots taken : this should help us detect regressions with the theming ability for CSS only usage of cozy-ui.


examples/stack/connection.html
```diff
@@ -0,0 +1,57 @@
...
+    {{ STYLE }}
+  </style>
+</head>
+<body>
+<div role="application">
+  <main class="wizard">
+    <form method="POST" action="/sharings/32176e6db46e8106bacf29f53b19abb4/discovery" class="wizard-wrapper">
+      <div role="region" class="wizard-main">
+        <input type="hidden" name="state" value="" />
+        <input type="hidden" name="sharecode" value="eyJhbGciOiJIUzUxMiIsInR5cCI6IkpXVCJ9.eyJhdWQiOiJzaGFyZSIsImlhdCI6MTYwMTI4MDI5MiwiaXNzIjoicGJyb3duZS5teWNvenkuY2xvdWQiLCJzdWIiOiJwdC5icm93bmUrdGVzdDFAZ21haWwuY29tIn0.-wZ6vEAY0p-LthjXlXtDWSUfjxjbzt8zyQbvgrrjRmtE2MAO_N85GX14DSpiU1eiusi54mv7TVXIEXl5ppju1Q" />
```
nono, line 15: I think we can remove the value here, as it's an hidden input

examples/stack/connection.html
```diff
@@ -0,0 +1,57 @@
...
+        <div class="c-avatar u-mh-auto u-mv-0-m u-mv-1 c-avatar--xlarge">
+          <img class="c-avatar-image" src="/public/avatar" alt="" />
```
nono, line 17: We don't an image that we can use here? I think it would be better to have any image that a 404

nono (commented):
LGTM

examples/stack/connection.html
ptbrowne: I added an image, thanks.

Crash-- (approved):
TOP!

nono (approved):

PR #1582, opened 2020-09-29
Fix/alerter mobile

Crash--:
Two issues resolved in this PR : 

- alerter's zIndex was too low on mobile. I don't know why we had a different value on mobile or desktop
- alerter now handles correctly the safe area. Same as we have done for action menu (#1573) 

I fixed a case issue on a CSS var. It's a breaking change... 

let's see what argos tells me

### DIFF
--- a/stylus/components/alerts.styl
+++ b/stylus/components/alerts.styl
@@ -12,7 +12,7 @@ alert-width = rem(640)
 
 $alert
     position fixed
-    z-index $alert-index-mobile
+    z-index $alert-index
     right 0
     bottom calc(3rem + env(safe-area-inset-bottom))
     left 0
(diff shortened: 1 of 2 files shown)


stylus/components/alerts.styl
```diff
@@ -46,7 +46,8 @@ $alert-wrapper
     background-color var(--validColor)
     padding rem(13 16)
     pointer-events auto
-
+    padding-bottom 3rem
+    padding-bottom env(safe-area-inset-bottom)
```
JF-Cozy, line 50: l'un ne remplace pas l'autre dans ce cas ? Quelle est l'idée ici, d'avoir 3rem ou env(safe-area-inset-bottom) c'est ça ?

Crash--:
Need to work on it, closing it for now

PR #1591, opened 2020-10-07
fix: Use correct proptype for CollectionField::component

ptbrowne:
element is for the instance of a component, elementType is
for the class/functional component.

### DIFF
--- a/react/Labs/CollectionField/index.jsx
+++ b/react/Labs/CollectionField/index.jsx
@@ -116,8 +116,8 @@ const CollectionField = props => {
 CollectionField.propTypes = {
   /** @type {Array} Individual values will be passed through `value` to the underlying Component */
   values: PropTypes.array.isRequired,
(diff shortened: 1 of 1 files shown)


react/Labs/CollectionField/index.jsx
```diff
@@ -117,7 +117,7 @@ CollectionField.propTypes = {
   /** @type {Array} Individual values will be passed through `value` to the underlying Component */
   values: PropTypes.array.isRequired,
   /** @type {Element} Component used to render a field */
```
Crash--, line 119:
```suggestion
  /** @type {ElementType} Component used to render a field */
```

Crash-- (approved):

react/Labs/CollectionField/index.jsx
ptbrowne: Thanks

Issue #1593, opened 2020-10-08
SelectBox Control with isMenuOpen on mobile


JF-Cozy:
In case of managing isMenuOpen with a custom function on a custom Control, there is a bug on mobile, see https://github.com/JedWatson/react-select/issues/3806.

Cozy-ui should therefore be able to manage this case directly, especially when using a non-customized Control. We should have a default Control that takes into account this little hack.

PR #1595, opened 2020-10-09
fix: Default Control works now on mobile

JF-Cozy:
We have now a default react-select control that works on mobile in the case we want to use an onClick function on the control, specially to manage the isMenuOpen prop.

Fix https://github.com/cozy/cozy-ui/issues/1593

```
yarn build:doc:react
```

- [x] Did you think of ARIA attributes if you are coding new components ?

### DIFF
--- a/react/SelectBox/ControlDefault.jsx
+++ b/react/SelectBox/ControlDefault.jsx
@@ -1,0 +1,26 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import { components } from 'react-select'
+
+const ControlDefault = ({
+  selectProps: { onControlClicked },
+  innerProps,
+  ...props
+}) => {
+  // onTouchStart is necessary on mobile
+  // see https://github.com/JedWatson/react-select/issues/3806#issuecomment-541325710
+
+  const enhancedInnerProps = {
+    ...innerProps,
+    onClick: onControlClicked,
+    onTouchStart: onControlClicked
+  }
+  return <components.Control {...props} innerProps={enhancedInnerProps} />
+}
+
+ControlDefault.propTypes = {
+  selectProps: PropTypes.object.isRequired
+}
+
+export default ControlDefault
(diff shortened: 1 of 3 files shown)


react/SelectBox/ControlDefault.jsx
```diff
@@ -0,0 +1,24 @@
...
+const ControlDefault = ({ selectProps: { onControlClicked }, ...props }) => {
+  // onTouchStart is necessary on mobile
+  // see https://github.com/JedWatson/react-select/issues/3806#issuecomment-541325710
+  return (
+    <div
+      onClick={onControlClicked}
+      onTouchStart={onControlClicked}
+      data-testid="selectBox-controlDefault"
+    >
+      <components.Control {...props} />
```
Crash--, line 15: Why not setting the onClick / onTouchStart on the `components.Control`? We're adding a new `div` which creates a new node. Can't we avoid that?

JF-Cozy: No we can't, passing props (onClick, onTouchStart, data-testid) directly on <components.Control {...props} /> won't work

Crash--: And something like: 

```
const ControlDefault = ({ selectProps: { onControlClicked }, innerProps, ...props }) => {

const enhancedInnerProps = {...innerProps, onClick: onControlClicked, onTouchStart: onControlClicked} 

<components.Control {...props} innerProps={enhancedInnerProps} />
```

is not working? 🤔

JF-Cozy: yes it does, thanks 👍

Crash--: Next time, take a look at the `react-select` implementation to understand why passing `props` was not working (https://github.com/JedWatson/react-select/blob/v2.x/src/components/Control.js#L56)

Crash-- (approved):

PR #1597, opened 2020-10-12
Fix/scroll lock

Crash--:
- If the content of the Dialog was > on the view port size, then the scroll was not working anymore after opening a dialog from an ActionMenu (because of a conflict between MUI and body-scroll-lock) 

- Disable scroll when BottomDrawer is displayed 

- Disable scroll when Dialog is opened

https://crash--.github.io/cozy-ui/react

Discussion on MUI side: https://github.com/mui-org/material-ui/issues/5750#issuecomment-471806500

### DIFF
--- a/react/ActionMenu/index.jsx
+++ b/react/ActionMenu/index.jsx
@@ -21,6 +21,28 @@ const ActionMenuWrapper = ({
   preventOverflow,
   children
 }) => {
+  if (!inline) return <BottomDrawer onClose={onClose}>{children}</BottomDrawer>
+  return (
+    <NotInlineWrapper
+      anchorElRef={anchorElRef}
+      popperOptions={popperOptions}
+      placement={placement}
+      preventOverflow={preventOverflow}
+      onClose={onClose}
+    >
+      {children}
+    </NotInlineWrapper>
+  )
+}
+
+const NotInlineWrapper = ({
+  anchorElRef,
+  popperOptions,
+  placement,
+  preventOverflow,
+  onClose,
+  children
+}) => {
   const [popperElement, setPopperElement] = React.useState(null)
   const referenceElement = anchorElRef ? anchorElRef.current : null
 
@@ -44,8 +66,7 @@ const ActionMenuWrapper = ({
     popperElement,
     options
   )
-
-  return inline ? (
+  return (
     <div
       ref={setPopperElement}
       style={{
@@ -56,11 +77,8 @@ const ActionMenuWrapper = ({
     >
       <ClickAwayListener onClickAway={onClose}>{children}</ClickAwayListener>
     </div>
-  ) : (
-    <BottomDrawer onClose={onClose}>{children}</BottomDrawer>
   )
 }
-
 const logDepecratedPlacement = createDepreciationLogger()
 const logDepecratedOverflow = createDepreciationLogger()
 const logDepecratedContainer = createDepreciationLogger()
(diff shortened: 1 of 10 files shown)


react/ActionMenu/index.jsx
```diff
@@ -56,8 +57,6 @@ const ActionMenuWrapper = ({
     >
       <ClickAwayListener onClickAway={onClose}>{children}</ClickAwayListener>
     </div>
-  ) : (
-    <BottomDrawer onClose={onClose}>{children}</BottomDrawer>
```
ptbrowne, line 60: Why this change ? All the hooks call after the early return are conditional now :/

react/Overlay/index.jsx
```diff
@@ -54,13 +47,14 @@ class Overlay extends Component {
   render() {
     const { children, className } = this.props
     const domProps = omit(this.props, nonDOMProps)
+    const Wrapper = bodyTallerThanWindow() ? React.Fragment : RemoveScroll
```
ptbrowne, line 50: Why do we have to check this ? Shouldn't we render RemoveScroll in all cases ? If it's necessary, could you add a comment why ?

ptbrowne:
Do you know how react-remove-scroll avoids the issue of body-scroll-lock ?

react/ActionMenu/index.jsx
Crash--: Oh you right, `inline` can change during life cycle. I'll go deeper and extract the react hooks in a new component to fix the point. 

The idea was : 
We don't change viewport size pretty often, and in mobile at least, we don't need to call `usePopper` and all.

react/Overlay/index.jsx
Crash--: Done and issue created too 👍 https://github.com/cozy/cozy-ui/issues/1599

Crash--:
> Looking good, but I'd like to know more what makes react-remove-scroll work and not body-scroll-lock. Do you know how react-remove-scroll avoids the issue of body-scroll-lock ?

Yep: To be concise, react-remove-scroll is not adding inline style to `body` but instead is create a new inline `<style>` in our HTML document (setting body { overflow: hidden...}). 

By itself, react-remove-scroll is not resolving exactly the issue, but since it uses a different paradigm we don't have the conflict anymore (between MUI and scroll locker) 

I say `not exactly resolving the issue`, but I can be wrong, it seems that `react-remove-scroll` generates an `id` for each `animations` and by doing so it seems to be able to compute the right style at the right moment

JF-Cozy (approved):

ptbrowne:
Ah ok, it's clever to use a new style element. This way the component only has to add/remove the style element, and it does not have to keep track of the previous value, it's handled by the browser.

ptbrowne (approved):

Crash--:
argos is broken because I added an exemple with a long text in `Dialog`. I think it's ok to merge ;)

JF-Cozy:
> argos is broken because I added an exemple with a long text in `Dialog`. I think it's ok to merge ;)

Tu devrais pouvoir le mettre à jour pour le faire passer au vert normalement, non ?

Crash--:
Oui, c'est ce que je vais faire, mais je laisse le temps @ptbrowne de remettre son 👍

ptbrowne:
In Argos 

* on the "Dialog-800x600.png" example, the overlay is much darker, is it expected ?
* on the "Dialog-300x600.png" example, the text seems to go above the buttons, is it expected ?

ptbrowne (approved):

PR #1606, opened 2020-10-14
feat: Better formatting for numbers

ptbrowne:
- We rely on a custom function instead of the localization function
    from the browser since some browsers like safari can lack this
    library
   - Doing the formatting ourselves opens up opportunities to format the
    decimal part and normal part differently (for example the decimal part in a smaller font, as seen in some designs from joel)

See https://ptbrowne.github.io/cozy-ui/react/#!/Figure

### DIFF
--- a/react/Figure/Figure.jsx
+++ b/react/Figure/Figure.jsx
@@ -1,18 +1,43 @@
-import React from 'react'
+import React, { useMemo } from 'react'
 import Types from 'prop-types'
 import cx from 'classnames'
 import styles from './Figure.styl'
+import { useI18n } from '../I18n'
 
-/**
- * Shows a number, typically a balance or an important financial
- * number in a bold way.
- */
 const stylePositive = styles['Figure-content--positive']
 const styleNegative = styles['Figure-content--negative']
 const styleWarning = styles['Figure-content--warning']
 const styleBig = styles['Figure--big']
 const styleClickable = styles['Figure--clickable']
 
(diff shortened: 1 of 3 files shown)


Crash--:
Do we have a radar or a bug track for this Safari bug that we can link to this issue?

ptbrowne:
I can't find a bug report :/ 

reproducing bug : https://output.jsbin.com/taxicuk/1

safari

image

firefox

image

See the lack of space on Safari.

code:

```
const nf = new Intl.NumberFormat(['fr-FR', 'en'])
document.body.innerHTML = nf.format(10000.50)
```

ptbrowne:
Thanks for the question, I ended up testing more thorougly and it seems that the formatting went correctly but then it was the space that was not correctly rendered in safari. By replacing spaces in the formatted string into non breaking spaces, it's working correctly. The implementation is thus
much simpler: https://github.com/cozy/cozy-ui/pull/1606/commits/3f6ff0e7f1e538d1df7a1e8eeaf98c578a0228af

nono:
Ping @ptbrowne

ptbrowne:
We have not heard again from this problem, I wonder if the complexity added here is worth it, we should check if Safari still has a problem formatting number. I think it was fab that had the problem in banks (big numbers did not have a thousand separating space).

JF-Cozy:
Ne semble plus être une priorité

PR #1607, opened 2020-10-14
Fix/selection bar

Crash--:
There is a lot to be done for the SelectionBar component, but let's do it gradually.

First: SelectionBar action can now have an icon attribute to use instead of relying only on the Action name
Second: Added an example on the Styleguidist 

https://crash--.github.io/cozy-ui/react/#/Layout%20components?id=selectionbar

@ptbrowne it seems that the style for the SelectionBar is not exported / imported in the styleguidist. Do you know how I can add it?

### DIFF
--- a/react/SelectionBar/index.jsx
+++ b/react/SelectionBar/index.jsx
@@ -12,7 +12,8 @@ If you want use SelectionBar component, you must have `actions` parameter, like 
 
 actions = {
   trash: {
-    action: selections => dispatch(trashFiles(selections)))
+    action: selections => dispatch(trashFiles(selections))),
+    icon: 'trash'
   },
   rename: {
     action: selections => dispatch(startRenamingAsync(selected[0])),
@@ -52,7 +53,7 @@ const SelectionBar = ({
           key={index}
           disabled={selectedCount < 1}
           onClick={() => actions[actionName].action(selected)}
-          icon={actionName.toLowerCase()}
+          icon={actions[actionName].icon || actionName.toLowerCase()}
           label={t('SelectionBar.' + actionName)}
           iconOnly={isMobile || isTablet ? true : false}
           subtle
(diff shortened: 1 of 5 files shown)


ptbrowne:
The styles were not applied to the component because the CSS rules were scoped under `role=application`. I put the rules under a mixin in the CSS implementation that I import under a class in the React style component. I also could not stop myself and did a tiny bit of refactoring, see the subsequent commits.

Other improvements that could be done>

- The docstring could be improved into a real docstring
- PropTypes could be improved with the shape of the actions
- Locales should be shipped with the component
- A deprecation warning could be logged when an action does not have actions[actionName].icon (could surely be done via the PropTypes)
- Semantic color variables could be used in the CSS
- I wonder why content-box is used, I find border-box to be easier to use and understand

react/SelectionBar/index.jsx
```diff
@@ -23,13 +23,9 @@ actions = {
...
-const SelectionBar = ({
-  t,
-  actions,
-  selected,
-  hideSelectionBar,
-  breakpoints: { isMobile, isTablet }
-}) => {
+const SelectionBar = ({ actions, selected, hideSelectionBar }) => {
+  const { t } = useI18n()
+  const { isMobile, isTablet } = useBreakpoints()
```
Crash--, line 28: Using useBreakpoins() is a BC if the APP doesn't use a `BreakpointsProvider`

ptbrowne:
> Have you an idea of why we were doing that?

No, I guess it was a mistake.

react/SelectionBar/index.jsx
ptbrowne: Thanks, I've edited https://github.com/cozy/cozy-ui/pull/1607/commits/b97fd3cefabd616ed289eb25f3346fe09957b1dd to add a BREAKING CHANGE mention.

stylus/components/selectionbar.styl
```diff
@@ -29,9 +29,6 @@ $selectionbar
         padding 0 rem(16)
         color   var(--white)
 
-    .SelectionBar-action-extra
-        display none
-
     .SelectionBar-action-close
```
JF-Cozy, line 32: ce ne serait pas plutôt `SelectionBar-actionClose` suite aux changement du commit précédent ?

ptbrowne (approved):

PR #1613, opened 2020-10-19
Change tabs font size to match subtitle1

ptbrowne:
We had a small discrepancy between our tabs font-size and the one that
we should use (the fontsize from subtitle1).

https://ptbrowne.github.io/cozy-ui/react/#!/Tab

### DIFF
--- a/react/MuiCozyTheme/theme.js
+++ b/react/MuiCozyTheme/theme.js
@@ -83,16 +83,6 @@ export const normalTheme = createMuiTheme({
     borderRadius: defaultValues.borderRadius
   },
   breakpoints: {
-    // Define custom breakpoint values.
-    // These will apply to Material-UI components that use responsive
-    // breakpoints, such as `Grid` and `Hidden`. You can also use the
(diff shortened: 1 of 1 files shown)


react/MuiCozyTheme/theme.js
```diff
@@ -83,16 +83,6 @@ export const normalTheme = createMuiTheme({
...
-    // Define custom breakpoint values.
-    // These will apply to Material-UI components that use responsive
-    // breakpoints, such as `Grid` and `Hidden`. You can also use the
-    // theme breakpoint functions `up`, `down`, and `between` to create
-    // media queries for these breakpoints
-    // xs = all
-    // sm = tiny
-    // md = small
-    // lg = medium
-    // xl = large
```
Crash--, line 95: This is not a "theme exemple doc". It's in fact the correspondence table between UI's breakpoints and MUI's ones.

Crash-- (approved):

Crash--:
MAIS NON ARGOS ETAIT ROUGE auto-bot ! (bon le rouge était logique mais quand même)

PR #1615, opened 2020-10-21
chore: Simplify watch a bit

JF-Cozy:
- Since React is default, no need to specify watch:doc:react
- move helpers inside /react to have only one watch


ptbrowne:
> yarn watch should watch everything

What do you mean in the context of this PR ?

JF-Cozy:
I mean react and helpers

package.json
```diff
@@ -55,10 +55,10 @@
     "posttranspile": "postcss transpiled/react/stylesheet.css --replace",
     "travis-deploy-once": "travis-deploy-once",
     "watch:css": "yarn build:css --watch",
-    "prewatch:doc:react": "yarn sprite",
-    "watch:doc:react": "env BUILD_ENV=watch-styleguidist styleguidist server --config docs/styleguide.config.js",
+    "prewatch:doc": "yarn sprite",
+    "watch:doc": "env BUILD_ENV=watch-styleguidist styleguidist server --config docs/styleguide.config.js",
     "watch:doc:kss": "nodemon --ext styl,md --watch stylus --exec 'yarn build:doc:kss && http-server build/styleguide -p 4242'",
-    "watch": "yarn transpile:react --watch"
+    "watch": "yarn transpile --watch"
```
ptbrowne, line 61: Here the flag "--watch" will not be passed to `yarn transpile:react` since the final command will be "yarn transpile:react && yarn transpile:helpers --watch". Is it working on your machine ?

ptbrowne: I think [npm-run-all](https://github.com/mysticatea/npm-run-all/blob/master/docs/npm-run-all.md) could work here as it seems possible to forward arguments to each sub command.

JF-Cozy: No it does not, you're right. So the doc was not relevant. I change it again.

ptbrowne (dismissed):

ptbrowne (approved):
👍

PR #1617, opened 2020-10-22
fix: Lorempixel is down lets use another service

Crash--:
Resulting by having wrong screenshots on argos. Let's fix that :)


ptbrowne (approved):

ptbrowne:
Nit: A base64 image could also do the job, this way we would not rely on an external provider. If we problems in the future, we could try that.

PR #1620, opened 2020-10-26
Muiv4

Crash--:
Come on Argos, let's see what we broke !

[See below for the post describing the migration](https://github.com/cozy/cozy-ui/pull/1620#issuecomment-773872546).


ptbrowne:
From Patrick:

- Update material-ui to V4
- Lots of small updates to theme, docs, and components for screenshots diff to be as small as possible

## Screenshot server

To easily check screenshots locally, I included in this PR a server that does a bit like Argos but locally. I could extract from this PR if deemed necessary.

## BREAKING CHANGES

- Your app must use material ui V4

```
yarn remove @material-ui/core
yarn add @material-ui/core@4
```

- Rename ExpansionPanel to Accordion

We already had the Accordion component in cozy-ui. We removed it and renamed the ExpansionPanel that we re-export from material-ui into Accordion.

```patch
- import ExpansionPanel from 'cozy-ui/transpiled/react/ExpansionPanel'
+ import Accordion from 'cozy-ui/transpiled/react/Accordion'
+ import AccordionSummary from 'cozy-ui/transpiled/react/AccordionSummary'
+ import AccordionDetails from 'cozy-ui/transpiled/react/AccordionSummary'
```

- Must use a StyleProvider to disable the global styles, since we have multi tree apps (since the bar is in its own React tree)

```
+ // Disable global styles because of multi react tree app
+ // https://material-ui.com/styles/api/#stylesprovider
+ const generateClassName = createGenerateClassName({
+   disableGlobal: true
+ })

const AppContainer = ({ children }) => {
  return (
+    <StylesProvider generateClassName={generateClassName}>
      { children }
+    </StylesProvider>
  )
}
```


## Screenshots

I have checked with @joel-costa the screenshots, the difference should be minimal. A lot of screenshots have differences because of a change of default color of the font. This change is only visible because
the Text in examples is not always wrapped into Typography components.

## CI

The CI is red but it seems it is only because of a TouchRipple issue in the snapshots. This seems like a schrodinger bug (it is sometimes here but sometimes is not). I am not sure how to fix the problem and I do not think it is due to the update.

Example of [diff](https://travis-ci.com/github/cozy/cozy-ui/builds/216162181#L2444):

```
-     <div class="styles__img___3SHpG"><span class="MuiSwitch-root"><span class="MuiButtonBase-root MuiIconButton-root PrivateSwitchBase-root-16 MuiSwitch-switchBase MuiSwitch-colorPrimary PrivateSwitchBase-checked-17 Mui-checked" aria-disabled="false"><span class="MuiIconButton-label"><input class="PrivateSwitchBase-input-19 MuiSwitch-input" type="checkbox" value="" checked=""><span class="MuiSwitch-thumb"></span></span><span class="MuiTouchRipple-root"></span></span><span class="MuiSwitch-track"></span></span></div>

+     <div class="styles__img___3SHpG"><span class="MuiSwitch-root"><span class="MuiButtonBase-root MuiIconButton-root PrivateSwitchBase-root-16 MuiSwitch-switchBase MuiSwitch-colorPrimary PrivateSwitchBase-checked-17 Mui-checked" aria-disabled="false"><span class="MuiIconButton-label"><input class="PrivateSwitchBase-input-19 MuiSwitch-input" type="checkbox" value="" checked=""><span class="MuiSwitch-thumb"></span></span></span><span class="MuiSwitch-track"></span></span></div>
```

## Next steps

The beta of cozy-ui already has been tested with Banks. Some [tweaks were needed](https://github.com/cozy/cozy-banks/pull/1997) (mostly on the Accordion whose theme was moved from banks to ui) but the migration was smooth.

- [x] Test with banks
- [ ] Upgrade in other apps.
   - [ ] Home
   - [ ] Settings
   - [ ] Notes
   - [ ] Contacts
   - [ ] Drive
   - [ ] Photos 

I think to reduce risk that we should merge this branch into a next version, publish betas from this branch, use the betas in our apps, and wait a little bit before publishing a stable version of the lib.
Subsequent PRs should be made against the next branch, but in case of an urgent fix, we can use the master branch to publish an hotfix.

Crash--:
👍 for the next steps.

PR #1638, opened 2020-11-04
feat: Use open prop like MUI Dialog

ptbrowne:
This way we have no discrepancy between MUI Dialog and our Dialog, we
have less props to learn. And it follows the principle of least astonishment.

### DIFF
--- a/react/CozyDialogs/dialogPropTypes.js
+++ b/react/CozyDialogs/dialogPropTypes.js
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 
 export default {
-  opened: PropTypes.bool.isRequired,
+  open: PropTypes.bool.isRequired,
   onClose: PropTypes.func.isRequired,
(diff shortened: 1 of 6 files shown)


react/CozyDialogs/dialogPropTypes.js
```diff
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types'
 
 export default {
-  opened: PropTypes.bool.isRequired,
+  open: PropTypes.bool.isRequired,
```
JF-Cozy, line 4: Pourquoi enlever opened ici ? Pour trigger un message d'erreur dans la console ?

JF-Cozy (approved):

react/CozyDialogs/dialogPropTypes.js
ptbrowne: Pour ne pas que ce soit documenté.

PR #1639, opened 2020-11-04
feat: Allow Mui.Modal & Mui.Dialog props

Crash--:
Allow the use of https://v3.material-ui.com/api/dialog/#props & https://v3.material-ui.com/api/modal/#props

PR #1640, opened 2020-11-04
More powerful useCozyDialog

ptbrowne:
useCozyDialog now returns full props for children components of 
Dialog. This allows us 

* to less repeat ourselves in the CozyDialogs
* to make useCozyDialog public and useful when creating a custom dialog since you can spread
the props given by useCozyDialog to the respective components and have the same behavior as the CozyDialogs (fullscreen for modals with size > m on mobile, transition when opening/closing the modal, aria-label).

Additionally, I have shown an example to show how to deal with list items inside modals, and now all CozyDialogs spread their props to the underlying Dialog (with the help of useCozyDialog).

Supersedes https://github.com/cozy/cozy-ui/pull/1639/files.

I have taken CozyDialogs screenshots from master and from this branch : here is the results, I can't see any difference [screenshots.zip](https://github.com/cozy/cozy-ui/files/5488306/screenshots.zip).


Crash-- (approved):
I really love this implementation! 👍

PR #1644, opened 2020-11-09
Fix/dialog z index

Crash--:
MUi's dialog comes with a 1300 zIndex. We need to fix it back to 70 as our previous modal. Like that, Alerter are displayed on top of Dialog...

We already done that on https://github.com/cozy/cozy-ui/pull/1408 but it was removed in https://github.com/cozy/cozy-ui/commit/391704c5e7ee74c75ef1b78b597199a8b9ad1367#diff-68934137b7c3b9a38678727aa3dbd31f8cb5fe366b4e3f31a30a675856010841

https://crash--.github.io/cozy-ui/react/#!/Dialog/1

### DIFF
--- a/react/Dialog/Readme.md
+++ b/react/Dialog/Readme.md
@@ -22,6 +22,7 @@ import {
 import useBreakpoints, {
   BreakpointsProvider
 } from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
+import Alerter from 'cozy-ui/transpiled/react/Alerter';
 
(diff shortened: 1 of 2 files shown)


JF-Cozy (approved):

ptbrowne:
OK I wonder if should have done the other way around, increasing the z-index of the modal ? To follow better MUI. I do not have strong opinion on this one.

LGTM.

Crash--:
Merging since argos faillure is not because of this PR (see https://github.com/cozy/cozy-ui/pull/1645)

PR #1645, opened 2020-11-09
fix: Don't rely on external services for img

Crash--:
Sometimes we have issues on argos with IntentIframe & AppIcon because they used external services to have access to an image and network can be slow or down.

Let's use base64 img instead in order to not rely on external services


ptbrowne (commented):
Nit: Having the images in their own files placeholder-img-1.js could be a little better as we would be able to reuse the images, and it would not pollute the example.

LGTM though on the solution of using base64 images.

Crash--:
@ptbrowne Thanks for the idea! Done

JF-Cozy (approved):

ptbrowne (approved):
Thanks

PR #1646, opened 2020-11-09
Improve Circle and Avatar to use specific size

JF-Cozy:
We would like to add a 24px size for the Avatars => we can now use specific size for avatars.
There was also a need to correct some overlapping behavior.

Circle is improved as well

Avatar : https://jf-cozy.github.io/cozy-ui/react/#!/Avatar
Circle : https://jf-cozy.github.io/cozy-ui/react/#!/Circle

### DIFF
--- a/react/Avatar/index.jsx
+++ b/react/Avatar/index.jsx
@@ -1,11 +1,12 @@
 import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
+import assign from 'lodash/assign'
+
 import palette from '../palette'
 import styles from './styles.styl'
 import Icon from '../Icon'
-
-import PeopleIcon from 'cozy-ui/transpiled/react/Icons/People'
+import { createSizeStyle } from '../Circle'
 
 const nameToColor = (name = '') => {
   const colors = [
@@ -33,6 +34,9 @@ const nameToColor = (name = '') => {
   return colors[key]
 }
 
+const createBgColorStyle = ({ text, textId }) =>
+  text ? { backgroundColor: `${nameToColor(textId || text)}` } : {}
+
 export const Avatar = ({
   text,
   textId,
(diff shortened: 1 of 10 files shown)


ptbrowne:
I think we should have a component for the Avatar "chain" with overlap. Also, in the example, the avatars that overlap seem to lack a white border and/or box-shadow something to add a depth feelink ? @joel-costa

joel-costa:
> I think we should have a component for the Avatar "chain" with overlap. Also, in the example, the avatars that overlap seem to lack a white border and/or box-shadow something to add a depth feelink ? @joel-costa

Yep 👍
I don't know if it is something to provide in the component but I think it is interesting to note that ideally, white borders would not stay white when we switch the background-color of his container : 
image

JF-Cozy:
I will add the Avatar "chain" in the overlap exemple.

Overlap is a "forced" example just to show that it's not buggy if you do it (which you do in cozy-sharing). It's not something that can be used turnkey.

react/Avatar/index.jsx
```diff
@@ -60,23 +61,34 @@ export const Avatar = ({
...
-  size: PropTypes.oneOf(['xsmall', 'small', 'medium', 'large', 'xlarge']),
-  className: PropTypes.string
+  size: PropTypes.oneOf([
+    'xsmall',
+    'smaller',
+    'small',
+    'medium',
+    'large',
+    'xlarge'
+  ]),
```
Crash--, line 79: Hard to follow all this size. Why not starting to add "size=24" instead of "smaller" ?

stylus/components/avatar.styl
```diff
@@ -3,6 +3,8 @@
 $avatar
     background-color var(--paleGrey)
     color var(--silver)
+    position relative
+    z-index 1
```
Crash--, line 7: we do have a z-index.styl with css var. We should use for instance `$low-index` instead

Crash-- (commented):
Wondering if we should not provide in UI a component to display an AvatarsList 🤔

react/Avatar/index.jsx
ptbrowne: Yes. @joel-costa  do we really need 6 sizes of Avatar ? I think we should have less sizes and the ability for the user to give a custom size.

joel-costa: @ptbrowne : Size doesn't matter for you ? :p
You would have only 3 sizes like on MUI ? No worries for me if we can give a custom size. Even on the home, we have a specific size : 
image

But I don't know if it would be relevant to set the same 3 sizes as MUI (24 - 40 - 56). 
If we exclude the home, we only use 16 - 24 - 32 - 40 right ?

JF-Cozy:
For now, I will focus on this point: to be able to give a specific size in pixels

react/Circle/index.jsx
```diff
@@ -1,28 +1,34 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
+
 import styles from './styles.styl'
 
+export const createSizeStyle = size => {
+  const sizeToPixel = Number.isInteger(size)
+    ? size
+    : { xsmall: 16, small: 32, medium: 40, large: 48, xlarge: 64 }[size]
```
ptbrowne, line 10: Nit: The mapping can be put outside the function

ptbrowne (approved):

PR #1647, opened 2020-11-09
fix: Clear Timeout when reseting

Crash--:
We need to reset the timeout when reseting. Without, the call in the didUpdate is not done since timeout is setted.

It's a quick fix, I'm working on a refacto on this HOC to convert it to hooks. But I need to rewrite the tests, so let's go for this fix first, and later the refacto 👍

### DIFF
--- a/react/Viewer/withFileUrl.jsx
+++ b/react/Viewer/withFileUrl.jsx
@@ -62,6 +62,7 @@ const withFileUrl = BaseComponent =>
     }
 
     reset = () => {
+      this.clearTimeout()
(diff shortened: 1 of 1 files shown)


ptbrowne:
No problem with the quick fix. For further refactors, I think we should have the ability to control the timeout at the network layer, on the getDownloadLinkById method, or by decorating the promise. Here the TTL is by component whereas we'd want it to be on the request.

```
const prom = downloadLinkById(id)
try {
  this.prom = await timeout(prom, 5000)
} finally {
  this.prom = null
}
```

PR #1648, opened 2020-11-09
feat: Use SVGr icons, this way, an application does not have to include the whole sprite

ptbrowne:
With this change, our components no longer rely on the Icon sprite.

Automatically done via the codemod svgr-icons.


Crash-- (approved):

Crash--:
il semblerait qu'il manque quelques icones 🤔

PR #1650, opened 2020-11-10
Push changelog after semantic release

ptbrowne:
- Changelog was generated but not pushed.
- It seems we have to use the @semantic-release/git plugin to push release assets
to the repo.

I've upgraded @semantic-release, revamped the config to use a standard config ([taken from](https://github.com/semantic-release/git)), (the config format may have changed between 15 and 17) and I removed the custom parser preset since we no longer use emojis in our commits.

Should fix https://github.com/cozy/cozy-ui/issues/1531


JF-Cozy (approved):

Crash-- (approved):
<3

PR #1666, opened 2020-11-18
WIP fix: Scroll not locked when Modal is opened for screen width < 1024px

ptbrowne:
Add a simple scroll locker in useCozyDialogs. This can be removed when
we upgrade material ui.

I've added an example with an ActionMenu (inside CozyDialogs::Dialog) and everything seems to work
OK. I need to test on mobile browsers now.

https://ptbrowne.github.io/cozy-ui/react/#!/Dialog

On safari iOS I still have the problem : the issue from material ui is here : https://github.com/mui-org/material-ui/issues/5750.

### DIFF
--- a/react/CozyDialogs/Readme.md
+++ b/react/CozyDialogs/Readme.md
@@ -33,6 +33,8 @@ import {
   FixedDialog,
   FixedActionsDialog
 } from  'cozy-ui/transpiled/react/CozyDialogs'
+import ActionMenu, { ActionMenuItem, ActionMenuRadio } from 'cozy-ui/transpiled/react/ActionMenu';
+import { Text, Caption } from 'cozy-ui/transpiled/react/Text';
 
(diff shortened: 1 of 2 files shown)


Crash--:
@ptbrowne this is why you need to use react-scrolllock, it should work as expected (since I've it working on Safari Mobile for few Overlay)

Crash--:
@ptbrowne J'ai ce genre de fix : https://github.com/cozy/cozy-ui/compare/feat/BlockDialogScrollIos?expand=1 moins élégante que ton fix, mais semble faire le taff sur mobile & ios. Tu en penses quoi ?

ptbrowne:
Ca me va, pense bien à mettre un gros commentaire pour expliquer l'historique concernant ce bug.

ptbrowne:
https://github.com/cozy/cozy-ui/pull/1676 solves the problem.

PR #1674, opened 2020-11-26
feat: Better support for secondary color

ptbrowne:
- We do not use the secondary color much but it can be useful for
custom themes.
- Tweak MUI button styles so that they follow our chart. We now
can use MUIButtons directly

I checked with @joel-costa and we're aligned.

Styleguide : https://ptbrowne.github.io/cozy-ui/react/#!/Buttons

### DIFF
--- a/react/MuiCozyTheme/theme.js
+++ b/react/MuiCozyTheme/theme.js
@@ -103,10 +103,10 @@ export const normalTheme = createMuiTheme({
       main: getCssVariableValue('pomegranate')
     },
     secondary: {
-      light: getCssVariableValue('monza'),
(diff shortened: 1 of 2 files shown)


Crash-- (approved):
"Mui" Bouttons are so much bigger now, but looks like our own Buttons. Is it time to deprecated "ours"?

ptbrowne:
Yep, I have tweaked their styles so that they follow our UI guideliens. I am aiming at some point to remove our Buttons yes. This won't happen overnight though, I think first we need to use them in our own apps to see if everything is OK. I am planning to talk about this on Tuesday.

Issue #1678, opened 2020-12-07
TableCell doesn't forwardRef


JF-Cozy:
We can't add `ref={ref}` on `<TableCell />` component, so if we need that in app, we need to wrap it in a <span ref={ref} /> or use this can of trick in the TableCell children. See https://github.com/cozy/cozy-drive/pull/2187#discussion_r536729598 for context.

PR #1679, opened 2020-12-07
feat: Table components now forward ref

Crash--:
Let's forward ref for our Table's components. 

Fix https://github.com/cozy/cozy-ui/issues/1678 

http://crash--.github.io/cozy-ui/react/#/Table

### DIFF
--- a/react/Table/index.jsx
+++ b/react/Table/index.jsx
@@ -1,39 +1,69 @@
 import React from 'react'
-import styles from './styles.styl'
 import cx from 'classnames'
+import PropTypes from 'prop-types'
 
-export const Table = props => {
+import styles from './styles.styl'
+
+export const Table = React.forwardRef((props, ref) => {
   const { className, ...rest } = props
 
-  return <div className={cx(styles.Table, className)} {...rest} />
+  return <div className={cx(styles.Table, className)} {...rest} ref={ref} />
+})
+
+export const TableHead = React.forwardRef((props, ref) => {
+  const { className, ...rest } = props
+
+  return <div className={cx(styles.TableHead, className)} {...rest} ref={ref} />
+})
+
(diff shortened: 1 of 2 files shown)


react/Table/index.jsx
```diff
@@ -1,39 +1,69 @@
...
 
-  return <div className={cx(styles.TableHeader, className)} {...rest} />
+  return (
+    <div className={cx(styles.TableHeader, className)} {...rest} ref={ref} />
+  )
+})
+
+Table.propTypes = {
+  className: PropTypes.string,
+  ref: PropTypes.oneOfType([PropTypes.func, PropTypes.object])
```
ptbrowne, line 47: I do not think that we should put `ref` in the props, nor className in the propType. It should be a given that className / style and ref are supported by our components, no ? It oculd be a mention at the top of the styleguide, I can't find the page on material UI that said the same thing though.

Crash--: If all our components support className, style and ref I will be agreed with you. Since they don't, I find it useful to show the fact they support it. 

Maybe the real fix is to pass on all our components and fix this issue, but it's a bit of work I don't want to do right now.

ptbrowne: Why not putting, "all our components should support className / style and ref", please open an issue if you find a component that does not conform to do this ?

Crash--: I will. We did it for this one, and we'll do it again for others. But instead of relying on "external reports", I want to take the time to pass on all our components to check if they support those props or not before writing somewhere that our components handle "className, style & ref" without any issues.

ptbrowne: OK.

nono:

JF-Cozy (approved):

PR #1699, opened 2021-01-06
Deprecations and theme improvements

ptbrowne:
Removal of components that we do not use
Better support for themes


react/index.js
```diff
@@ -61,15 +51,6 @@ export { default as IntentHeader } from './IntentHeader'
...
 export { default as Chip } from './Chip'
-export {
-  default as Text,
-  MainTitle,
-  Title,
-  SubTitle,
-  Bold,
-  Caption,
-  Uppercase
-} from './Text'
```
Crash--, line 73: this commit is also a BC since we don't expose anymore Text or Modal from the root of UI, right?

stylus/settings/palette.styl
```diff
@@ -201,6 +201,7 @@ html, .CozyTheme--normal
     --secondaryButtonContrastColor: var(--black)
 
     --paperBackgroundColor: #FFF
+    --defaultBackgroundColor: var(--paleGrey)
     --dividerColor: var(--silver)
 
     --iconColor: currentColor
```
Crash--, line 207: can you rename this commit ? ;)

Crash-- (commented):
👍

react/index.js
ptbrowne: Yep, I've added the BC in the commit.

stylus/settings/palette.styl
ptbrowne: Done

JF-Cozy (approved):

PR #1705, opened 2021-01-12
Alert info color override

ptbrowne:
Ability to override AlertInfo background colors.
We add a level of indirection so that we can override only the background color
of a type of Alert.

### DIFF
--- a/stylus/settings/palette.styl
+++ b/stylus/settings/palette.styl
@@ -225,4 +225,13 @@ html, .CozyTheme--normal
     --navTextColor var(--slateGrey)
     --navTextActiveColor var(--slateGrey)
     --navTextHoverColor var(--charcoalGrey)
+
+    --alertErrorColor #fff
+    --alertErrorBackgroundColor var(--errorColor)
+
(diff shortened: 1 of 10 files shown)


stylus/settings/palette.styl
```diff
@@ -225,4 +225,13 @@ html, .CozyTheme--normal
...
     --navTextHoverColor var(--charcoalGrey)
+
+    --alertErrorColor #fff
+    --alertErrorBackgroundColor var(--errorColor)
+
+    --alertSuccessColor #fff
+    --alertSuccessBackgroundColor var(--validColor)
+
+    --alertInfoColor #fff
+    --alertInfoBackgroundColor var(--slateGrey)
```
flohhhh, line 236: We don't use semantic color here ? (for example --infoColor)

ptbrowne: In the theme, we are already "high" enough that we can use absolute colors IMHO.

flohhhh (approved):

JF-Cozy (approved):

PR #1708, opened 2021-01-12
Fix Dialogs and CozyDialogs theme

JF-Cozy:
PR necessary in order to correct the fact that with the addition of the theme via CozyTheme, it adds a tag in the children of Dialog, thus modifying the internal structure expected by Mui, thus changing the behavior of Mui Dialog.

This was particularly the case with FixedDialogs, which were no longer fixed.

https://jf-cozy.github.io/cozy-ui/react/#!/CozyDialogs

### DIFF
--- a/react/CozyDialogs/DialogBackButton.jsx
+++ b/react/CozyDialogs/DialogBackButton.jsx
@@ -1,17 +1,23 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { withStyles } from '@material-ui/core/styles'
 
+import IconButton from 'cozy-ui/transpiled/react/IconButton'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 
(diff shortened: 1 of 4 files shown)


react/CozyDialogs/DialogBackButton.jsx
```diff
@@ -1,17 +1,16 @@
...
 const DialogBackButton = ({ onClick }) => {
   return (
-    <Icon
-      onClick={onClick}
-      icon={PreviousIcon}
-      className="u-c-pointer u-coolGrey u-mr-1"
-    />
+    <IconButton className="u-mr-half" onClick={onClick}>
+      <Icon icon={PreviousIcon} />
+    </IconButton>
```
ptbrowne, line 13: 👍

ptbrowne (approved):

PR #1714, opened 2021-01-14
docs: Change Stack doc to show default spacing

JF-Cozy:
and use radio to switch examples

fix https://github.com/cozy/cozy-ui/issues/1712

https://jf-cozy.github.io/cozy-ui/react/#!/Stack

### DIFF
--- a/react/Stack/Readme.md
+++ b/react/Stack/Readme.md
@@ -5,42 +5,63 @@ See [this article](https://every-layout.dev/layouts/stack/) for more information
 
 The padding is responsive, it is smaller on mobile/tablet.
 
-```
+You can use `xs`, `s`, `m` (default), `l`, `xl`, and `xxl` values for "spacing" to have less/more padding.
+
+```jsx
 import Stack from 'cozy-ui/transpiled/react/Stack';
 import Card from 'cozy-ui/transpiled/react/Card';
 import Typography from 'cozy-ui/transpiled/react/Typography';
 
-<Stack>
-  <Card><Typography variant='body1'>Homer Simpson</Typography></Card>
-  <Card><Typography variant='body1'>Marge Simpson</Typography></Card>
-  <Card><Typography variant='body1'>Lisa Simpson</Typography></Card>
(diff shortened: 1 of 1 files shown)


react/Stack/Readme.md
```diff
@@ -5,42 +5,43 @@ See [this article](https://every-layout.dev/layouts/stack/) for more information
...
-</Stack>
+initialState = {
+  spacing: 'm'
+};
+
+const spacings = ['xs', 's', 'm', 'l', 'xl', 'xxl'];
+
+const StateRadio = ({ name, value, ...props }) => {
+  return (
+    <span className="u-mr-1">
```
ptbrowne, line 23: Please use a Typography variant='body1' here otherwise the example does not react well to theme change.

ptbrowne: Variants actually already does that, maybe you can use that ?

JF-Cozy: je l'ai utilisé plus bas à la place de `<p>` pour tout englober

ptbrowne: Actually, Variants does only work for boolean variables and is not directly suited for this.

The only problem I see here is that stack different spacings are no longer screenshotted. It is not a big problem since spacing values are not changed often and will be tested through other component examples.

Anyway, in the future it could be interesting to have a way to have in the styleguide an interactive version with checkboxes while in the screenshots have all the variants.

ptbrowne (approved):
After the change to Typography, LGTM

react/Stack/Readme.md
JF-Cozy: Pour ça on ne peut pas simplement ajouter un block avec tous les exemples dont l'affichage est conditionné avec un `isTesting()` ?

ptbrowne: Yep c'est bien ca :) Mais je me demandais si ca vaudrait pas le coup d'abstraire ça, c'était un peu le but de Variant au départ.

Ton exemple me va aussi.

PR #1718, opened 2021-01-18
feat: Ability to use lib locales without withLocales

ptbrowne:
Use a function to create a useI18n that is bound to lib locales.
Contrary to withLocales, context is not used to pass the polyglot
instance. Instead, the useI18n is created with the locales.

Advantage: 

- I18n context is not lost for children of components using the library locales.
- No nesting


JF-Cozy (approved):

react/I18n/index.jsx
```diff
@@ -94,7 +95,22 @@ export const translate = () => WrappedComponent => {
...
+  return useMemo(() => {
+    const polyglot = new Polyglot({
+      lang: lang,
+      phrases: locales[lang]
+    })
+    const f = initFormat(lang)
+    const t = polyglot.t.bind(polyglot)
+    return { t, f, lang }
+  }, [lang])
+}
```
Crash--, line 113: Awesome! Is it possible to add a test for this method to test it and also to have an example of how to use it?

Crash-- (approved):
I already enjoy this new method!

PR #1728, opened 2021-02-02
feat: Icon is no longer required in Empty [ready to review]

### LINKED ISSUE #1701: Better Empty Component
Have the possibility to hide Icon in Empty Component

Change Also Empty PropType :
 - icon: iconPropType.isRequire => icon: iconPropType

ptbrowne:
Fix #1701

### DIFF
--- a/react/Empty/index.jsx
+++ b/react/Empty/index.jsx
@@ -23,16 +23,7 @@ export const Empty = ({
   }
 
   return (
-    <div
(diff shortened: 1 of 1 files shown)


ptbrowne:
@Crash-- I see that you forced push, I do not understand the change though, and I think this PR is no longer necessary since we can remove the icon prop without any warning.

PR #1729, opened 2021-02-03
SVGr icons

ptbrowne:
Replace "string" icons by svgr icons. Apps that use svgr icons no
longer need to include the full icon sprite.


Crash-- (approved):
🎉

JF-Cozy (approved):

PR #1743, opened 2021-02-19
feat: Remove a lot of colors from the utility classes

ptbrowne:
I tried to remove CSS variables for colors that we do not use in our apps.
Absolute-color-named variables do not promote building component that work well with theming.
Instead of using absolute-color-named CSS variables, we prefer to use semantic colors.

Tried to make this a non breaking change by looking which classes were
used in our apps
Anyway, we should not use those utility classes as we should strive to
use intent based colors u-success, u-error, u-warning can be used instead.

Command used to check usage

```fish
for app in libs contacts drive store banks settings home; pushd $app; echo $app; rg '\bu-.[a-zA-Z\-0-9]*\b' -o | cat; popd; end | sort | uniq | grep -E 'black|white|overlay|zircon|frenchPass|dodgerBlue|scienceBlue|puertoRico|grannyApple|emerald|malachite|seafoamGreen|brightSun|texasRose|chablis|yourPink|fuchsia|pomegranate|monza|portage|azure|melon|blazeOrange|mango|pumpkinOrange|lavender|darkPeriwinkle|purpley|lightishPurple|barney|weirdGreen' | cut -d':' -f 2 | sort | uniq
```

BREAKING CHANGE: Some utility classes for text/background color have
been removed, please use semantic utility classes instead.

- u-primaryColor
- u-secondaryColor
- u-bg-primaryColor
- u-bg-secondaryColor
- u-warning
- u-error
- u-success

Some utility classes have been renamed to stick to [material-ui's naming](https://material-ui.com/customization/palette/#palette-colors).

`u-valid` has been renamed `u-success`
`u-warn` has been renamed `u-warning`
`u-danger` has been renamed `u-error`

---

@nono After our discussion I tried to make the documentation clearer on the subject of CSS variables.

Check the utilities section : https://ptbrowne.github.io/cozy-ui/styleguide/section-utilities.html
Check the settings theme section: https://ptbrowne.github.io/cozy-ui/styleguide/section-settings.html#kssref-settings-theme
I also added a point on the settings color section that links to the theme section : https://ptbrowne.github.io/cozy-ui/styleguide/section-settings.html#kssref-settings-colors

I hope this is clearer


nono:
The new parts of the documentation are nice :+1:

By the way, is it expected that `ligher` variants of primary/secondary colors are darker than the `light` variants?

ptbrowne:
> By the way, is it expected that ligher variants of primary/secondary colors are darker than the light variants?

This is "expected" but this is not the goal. It was already stated in the stylus file that primary color lighter should not be used. I removed it from the documentation, along with secondaryColorLighter. I haven't found `secondaryColorLighter` used in our apps, primaryColorLighter is only used in banks.

stylus/settings/palette.styl
```diff
@@ -122,64 +107,110 @@ html
...
+
+    /*
+    Text colors
 
-    --primaryTextColor var(--charcoalGrey)
-    --secondaryTextColor var(--coolGrey)
+    Stylus: primaryTextColor -  #32363F, CSS: var(--primaryTextColor)
+    Stylus: secondaryTextColor - #95999D, CSS: var(--secondaryTextColor)
+    Stylus: primaryContrastTextColor -  #FFF, CSS: var(--primaryContrastTextColor)
+    Stylus: primaryContrastTextColor -  #FFF, CSS(--secondaryContrastTextColor)
```
nono, line 177: s/primary/secondary/

stylus/utilities/colors.styl
```diff
@@ -0,0 +1,23 @@
+utility_colors = 'black',
+   'charcoalGrey',
+   'coolGrey',
+   'silver',
+   'slateGrey',
+   'danger',
+   'lightishPurple',
+   'dodgerBlue',
+   'error',
```
nono, line 9: Is it normal to have both `danger` and `error`?

nono:
If we are making a breaking change and promote some utility classes, I think we should use the material-UI naming: `u-success` instead of `u-valid`, and `u-warning` instead of `u-warn`. It will avoid another change later.

stylus/settings/palette.styl
ptbrowne: Fixed, thanks !

stylus/utilities/colors.styl
ptbrowne: It was "known" that we had the two colors, I removed "danger", thanks.

nono (commented):
LGTM but Argos wasn't happy for the avatars on last build

stylus/settings/palette.styl
```diff
@@ -135,7 +135,6 @@ html, .CozyTheme--normal
     --primaryColor #297EF2
     --primaryColorDark #0B61D6
     --primaryColorLight #5C9DF5
-    --primaryColorLighter #4B93F7 // should not be used
```
Crash--, line 168: You just use it here https://github.com/cozy/cozy-ui/pull/1743/commits/a67a446da44a756b03fe8827679004d07ea24e2f#diff-17e9585c61eae7bb5de203c99c5467520ba0d6f1a65cd45e3557287c70cce344R56

react/PasswordExample/styles.styl
```diff
@@ -1,5 +1,5 @@
 .Token--number
-    color var(--dodgerBlue)
+    color #297EF2 // dodgerBlue
```
Crash--, line 2: Should we import the palette and use it as `palette.dodgerBlue` instead of having an  other absolute colors? We already previously defined the absolute color for dodgerBlue, no? (hard to remember the start of this PR ^^')

stylus/settings/palette.styl
```diff
@@ -201,7 +201,7 @@ html, .CozyTheme--normal
     Styleguide Settings.theme.intention
     */
     --successColor: $successColor
-    --warningColor: $warnColor
+    --warningColor: $warningColor
```
Crash--, line 211: commit should be squashed or renamed

Crash-- (commented):
Lot of changes here! I need to read again this PR with a global picture to really understand the scope of the changes.

stylus/settings/palette.styl
```diff
@@ -146,15 +145,14 @@ html, .CozyTheme--normal
...
-    Stylus: secondaryColorLighter   - #fd7461, CSS: var(--secondaryColorLighter)
     Stylus: secondaryColorLightest   - #FECDC6, CSS: var(--secondaryColorLightest)
 
     Styleguide Settings.theme.secondary
     */
     --secondaryColor #fd7461
     --secondaryColorDark #E3503B
     --secondaryColorLight #ffdeda // lighten(dodgerBlue, 24)
-    --secondaryColorLighter #fd7461 //
+    --secondaryColorLighter #fd7461 // should not be used
```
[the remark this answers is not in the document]
ptbrowne: Removed later.

stylus/settings/palette.json
```diff
@@ -44,5 +16,33 @@
...
+  "blazeOrange": "#FC6D00",
+  "mango": "#FF962F",
+  "pumpkinOrange": "#FF7F1B",
+  "lavender": "#C2ADF4",
+  "darkPeriwinkle": "#6984CE",
+  "purpley": "#7F6BEE",
+  "lightishPurple": "#B449E7",
+  "barney": "#922BC2",
+  "weirdGreen": "#40DE8E"
 }
```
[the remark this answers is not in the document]
ptbrowne: Because Avatars use them, but we do not want to have the CSS vars as part of our CSS properties because then people start to use them too easily.

stylus/settings/palette.styl
```diff
@@ -122,25 +100,6 @@ html
...
-
-    Styleguide Settings.colors.purple
-    */
-    --lavender        #C2ADF4
-    --darkPeriwinkle  #6984CE
-    --purpley         #7F6BEE
-    --portage         #9169F2
-    --lightishPurple  #B449E7
-    --barney          #922BC2
-
```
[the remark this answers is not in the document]
ptbrowne: Done

react/PasswordExample/styles.styl
ptbrowne: Yep, done, don't know what I was thinking here ;)

stylus/settings/palette.json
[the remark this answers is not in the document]
Crash--: > Also, the CSS custom properties are there whether you use the colors or not

For my personal culture, is it bad? Can we have perf issue with lot of vars? 

> we do not want to have the CSS vars as part of our CSS properties because then people start to use them too easily

 We want to make it "harder" for the dev to use a custom palette colors to force him/her to think about semantic colors first. Clear enough, thanks!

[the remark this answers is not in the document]
ptbrowne: :+1: Sorry if it was unclear 😅 !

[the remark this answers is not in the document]
ptbrowne: Also I think it could be confusing to have CSS custom properties that are meant to be overrided and others that should not be.

nono:
On Argos, we can see that the border for chips and buttons has changed on some cases. Is it expected?

stylus/settings/palette.styl
```diff
@@ -122,64 +113,112 @@ html
...
+    --paperBackgroundColor: #FFF
+    --defaultBackgroundColor: #F5F6F7 // paleGrey
+
+    /*
+    Text colors
+
+    Stylus: primaryTextColor -  #32363F, CSS: var(--primaryTextColor)
+    Stylus: secondaryTextColor - #95999D, CSS: var(--secondaryTextColor)
+    Stylus: primaryContrastTextColor -  #FFF, CSS: var(--primaryContrastTextColor)
+    Stylus: secondaryContrastTextColor -  #FFF, CSS(--secondaryContrastTextColor)
```
nono, line 181: s/CSS/CSS: var/

stylus/settings/palette.styl
```diff
@@ -122,64 +113,112 @@ html
...
 
-    --primaryTextColor var(--charcoalGrey)
-    --secondaryTextColor var(--coolGrey)
+    Styleguide Settings.theme.text
+    */
+    --primaryTextColor #32363F
+    --secondaryTextColor #95999D
```
nono, line 186: For for those two CSS variables, I think I do not have the same definition in cozy-bootstrap: https://github.com/cozy/cozy-bootstrap/blob/main/src/cozy/_colors.scss#L26-L28

I have a value in RGBA with alpha < 1. Is it MUI that combines the CSS var and an alpha value from somewhere else? If yes, what do you think we should do for cozy-bootstrap? Use another name for the CSS variable?

stylus/settings/palette.styl
ptbrowne: Thanks

stylus/components/chip.styl
```diff
@@ -53,14 +53,14 @@ $normal-theme-chip
...
-    color var(--dodgerBlue)
+    border-color var(--primaryColorLighter)
+    background-color var(--primaryBackgroundLight)
+    color var(--primaryColor)
 
 $error-theme-chip
-    border-color var(--yourPink)
-    background-color var(--chablis)
-    color var(--pomegranate)
+    border-color var(--errorColorLighter)
```
nono, line 61: `--errorColorLighter` is not defined

stylus/settings/palette.styl
ptbrowne: > Is it MUI that combines the CSS var and an alpha value from somewhere else?

Not that I am aware of, but I think it can depend on the component, for example in Alerts where the `color` [is generated](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Alert/Alert.js#L70). I guess some components could create a text color with lowered opacity.

> If yes, what do you think we should do for cozy-bootstrap? Use another name for the CSS variable?

I am not sure why you have a different value in cozy-bootstrap. Is it a requirement from @joel-costa ? I know that text with a reduced opacity will work better on colored background, for example in alerts. That said, I do not think this warrants a change between cozy-bootstrap and cozy-ui, I would be in favor to stick with a non-alpha color value ?

nono: > Is it a requirement from @joel-costa ?

Yes, but it is also what MUI does. When I inspect a text with default settings in MUI documentation, the color is `rgba(0, 0, 0, 0.87)`.

> I know that text with a reduced opacity will work better on colored background, for example in alerts.

It is also the case for `--primaryTextColor` on inverted theme (and dark theme in the future).

> That said, I do not think this warrants a change between cozy-bootstrap and cozy-ui,

I already have some CSS variables for cozy-bootstrap that are not defined by Cozy-UI, like the text color for alerts. MUI adds the alpha channel in JS, but I can't do that with just CSS variables.

> I would be in favor to stick with a non-alpha color value?

Why should we take the non-alpha and not the with-alpha color value?

stylus/components/chip.styl
ptbrowne: Fixed, thanks.

stylus/settings/palette.styl
ptbrowne: Because I don't want to take any risk at this point. If someone wants to take this task, I have no problem.

stylus/components/button.styl
```diff
@@ -20,19 +21,19 @@ secondaryTheme = {
...
-    primaryColor: var(--zircon), secondaryColor: var(--frenchPass), activeColor: var(--frenchPass), contrastColor: var(--primaryColor)
+    primaryColor: var(--zircon), secondaryColor: palette['frenchPass'], activeColor: palette['frenchPass'], contrastColor: var(--primaryColor)
 }
 highlightTheme = {
-    primaryColor: var(--emerald), secondaryColor: var(--emerald), activeColor: var(--malachite), contrastColor: var(--white)
+    primaryColor: palette['emerald'], secondaryColor: palette['emerald'], activeColor: palette['malachite'], contrastColor: var(--white)
 }
 dangerTheme = {
-    primaryColor: var(--pomegranate), secondaryColor: var(--pomegranate), activeColor: var(--monza), contrastColor: var(--white)
+    primaryColor: palette['pomegranate'], secondaryColor: palette['pomegranate'], activeColor: var(--monza), contrastColor: var(--white)
```
nono, line 30: I don't understand why we have a mix of CSS variables and palette here. I think we should use CSS variables only, to allow custom themes. But if it is not possible currently, we should use only palette colors to avoid having surprising things on custom themes.

ptbrowne: The palette colors are used for CSS variables that are not themable, this is why I use them.

react/PasswordExample/styles.styl
```diff
@@ -1,5 +1,7 @@
+palette=json('../../stylus/settings/palette.json', { hash: true })
+
 .Token--number
-    color var(--dodgerBlue)
+    color palette['dodgerBlue']
```
nono, line 4: [nit] No `:` here? In another commit, the message says that it was mandatory to put `:` after the property name when an hash was used.

stylus/settings/palette.styl
```diff
@@ -213,10 +249,6 @@ html, .CozyTheme--normal
     --errorColor: $errorColor
```
nono, line 249: `--errorColor` is declared twice

nono (approved):

react/PasswordExample/styles.styl
ptbrowne: Ah yes it does not seem necessary here, it seems like a stylus peculiarity :?

stylus/settings/palette.styl
ptbrowne: Décidément...

stylus/settings/palette.styl
Crash--: Is it something we need to track somewhere?

Crash-- (approved):

PR #1745, opened 2021-02-23
feat: Have the option of not having a layout

flohhhh:
When screen is in "medium-screen" there is a margin top and a translate z.

The component must not have a defined layout,
we let the parent manage the display of the component.

We have the possibility of disable the layout.


react/Empty/index.jsx
```diff
@@ -4,17 +4,36 @@ import PropTypes from 'prop-types'
...
   title,
   text,
   children,
   className,
+  layout,
   ...restProps
 }) => {
+  if (layout) {
+    depreciationLogger(
+      'The use of c-empty style is deprecated, Empty component must not have a defined layout'
```
ptbrowne, line 22: I think it'd be more explicit with "Please add layout=false in the props of the Empty component, and implement the vertical centering with a custom layout component".

react/Empty/index.jsx
```diff
@@ -36,7 +55,12 @@ Empty.propTypes = {
   icon: iconPropType.isRequired,
   title: PropTypes.node.isRequired,
   text: PropTypes.node,
-  className: PropTypes.string
+  className: PropTypes.string,
+  layout: PropTypes.boolean
```
JF-Cozy, line 59: C'est normalement `PropTypes.bool`. Ca fonctionne avec `PropTypes.boolean` ?

flohhhh: Oops :)

react/Empty/index.jsx
flohhhh: You're right

ptbrowne (approved):

PR #1771, opened 2021-03-17
feat: Improvements

ptbrowne:
#### Dependencies

As cozy-ui is used as a library, it should have no problems with deps
that are minor versions above the current version it necessitates. It
should solve some of the duplicate deps that we have inside our apps

#### BarTitle

Used the wrong typography variant. Also added align-items-center to have vertical alignment.

https://ptbrowne.github.io/cozy-ui/react/#!/BarTitle
compare with https://cozy.github.io/cozy-ui/react/#!/BarTitle

#### NestedSelect

*  did not show back button

After migrating to new Dialog, we lost the back button when inside a category. This was due
to a bad usage of the showBack prop.

* when changing slide, the modal would jump due to centered alignment

Allow to add align: 'top' to useCozyDialogs. This way the modal can change its height but the top is always the same. This is also helpful for Dialog containing tabs like Harvest where the content of the modal can change.

https://ptbrowne.github.io/cozy-ui/react/#!/NestedSelect
compare with https://cozy.github.io/cozy-ui/react/#!/NestedSelect


react/NestedSelect/Modal.jsx
```diff
@@ -1,21 +1,46 @@
...
+  const { dialogTitleProps } = useCozyDialog({
+    open: true
+  })
+  return (
+    <DialogTitle {...dialogTitleProps}>
+      <Media>
+        {showBack && (
+          <Img className={styles.Modal__back}>
+            <IconButton onClick={onClickBack}>
+              <Icon icon={LeftIcon} color={palette['coolGrey']} />
```
JF-Cozy, line 28: Est-ce que coolGrey peut-être remplacé par quelque chose de plus sémantique ?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cozy-ui",
-  "version": "45.0.0-beta.3",
+  "version": "46.2.0-beta.1",
```
JF-Cozy, line 3: nit: Pour ma culture, dans quel cas doit-on / ne doit-on pas incrémenter ça ? Ce n'est pas fait automatiquement après une publication ?

react/NestedSelect/styles.styl
```diff
@@ -4,3 +4,10 @@
     margin-left -0.25rem
     font-size 1.5rem
     line-height 0
+
+    // Need to cancel the height of the icon button
+    // that expands the height of the modal header
+    // too much
+    margin-bottom -1rem
+    position relative
+    top -0.5rem
```
JF-Cozy, line 13: est-ce qu'on ne peut pas plutôt réduire la taille du bouton ?

ptbrowne: Non c'est l'effet de hover de l'icône qui pose problème. On peut pas réduire l'icône bouton sans changer la taille de l'effet de hover. Donc on est obligé de passer par là.

package.json
ptbrowne: C'est parce que j'ai publié sur @next, je l'enlèverais de ma PR ensuite.

react/NestedSelect/Modal.jsx
ptbrowne: Good point, I was actually able to completely remove the color props since IconButton already are colored with `theme.palette.text.secondary`.

JF-Cozy (approved):

PR #1778, opened 2021-03-24
Inter project coherent names for script

ptbrowne:
To have the same nomenclature across projects, I made some changes:

- transpile -> build
- watch -> start


Crash-- (commented):
I'm not confident with the upgrade of react-select. It should be tested and having a migration path if needed. 

We should split this PR to make it easier to merge

ptbrowne:
What do you mean it must be tested ? It's not like I did check both changelog pages (links in the commit) and all the examples from the styleguide page before pushing. Do you have an example where you are worried ?

Crash--:
> What do you mean it must be tested ? It's not like I did check both changelog pages (links in the commit) and all the examples from the styleguide page before pushing. Do you have an example where you are worried ?

I also checked the changelogs, they have changed the way they work internally with emotion, and I'm a bit worried about the visual regression. Seems that argos is agreed.

`Contacts` has an advanced usage of `SelectBox` (with CustomMenu, customOption, customSelect)

ptbrowne:
OK but we should not know about contacts when doing the change, to me the contract is what is shown in the styleguide, if an app does not follow the styleguide, it's difficult to foresee problems. I had not seen yet the Argos screenshots and yes it seems that it breaks stuff (I think locally I need to transpile to see the problems), so evidently we cannot merge as is.

Crash-- (approved):

Crash--:
Just checked the styleguide, we have added exemple with customOption and an other one pretty close to what we do on Contacts so we're clean and should not be too worried about the upgrade if argos is passing.

ptbrowne:
Great 💯 Thanks to have checked that !

PR #1783, opened 2021-03-24
feat: Upgrade react-select

ptbrowne:
Hopefully, this will remove duplicate dependencies (like an old version of react-transition) from our builds.

BREAKING CHANGE: react-select is upgraded from v2 to v4. It should not
impact you but you should be extra careful around the areas that use a
select.

Values are now normalized:
https://github.com/JedWatson/react-select/pull/4339 The onChange callback
receives values that are more coherent now (array if multi select, value is
always an array (can be empty), if a single select, value can be
object/null.

V4 changelog:
https://github.com/JedWatson/react-select/blob/master/packages/react-select/     CHANGELOG.md#400
V3 changelog:
https://github.com/JedWatson/react-select/blob/master/packages/react-select/     CHANGELOG.md#300

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "react-pdf": "^4.0.5",
     "react-popper": "^2.2.3",
     "react-remove-scroll": "^2.4.0",
-    "react-select": "^2.2.0",
+    "react-select": "^4.3.0",
     "react-swipeable-views": "^0.13.3"
   },
   "peerDependencies": {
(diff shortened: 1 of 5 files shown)


ptbrowne:
From crash:

> Just checked the styleguide, we have added exemple with customOption and an other one pretty close to what we do on Contacts so we're clean and should not be too worried about the upgrade if argos is passing.

Issue #1758, opened 2021-03-08
Buttons icons should react to button size


ptbrowne:
In the example `startIcon` should be used instead of putting an Icon
child. In the end, we should have bigger icons for bigger buttons
and smaller icons for smaller buttons.

See working example here: https://material-ui.com/components/buttons/#buttons-with-icons-and-label

JF-Cozy:
issue relative aux Button déprécié. Les nouveaux Buttons utilisent l'api MUI

PR #1788, opened 2021-03-30
Improvements to icons

ptbrowne:
The original goal was to have the icons inside buttons change following the
size of the button.

So the first commit deal with that issue (https://github.com/cozy/cozy-ui/issues/1758), and then I saw that Material-UI uses width='1em' and height='1em'
directly in the icons that they generate, so that they are usable without
the need for <Icon />. I regenerated the icons svgr with the --icon flag
from svgr that does just that. This can be a breaking change if you are using
an SVGr icon directly.

When contained in an Icon, there will be no breaking change normally.


ptbrowne:
I think this could be a nice improvement. We have already talked about this with @joel-costa .

JF-Cozy:
On garde sous le coude si on se penche à nouveau sur le sujet

PR #1794, opened 2021-04-06
Fix/viewer

Crash--:
- Upgrade testing-lib to get waitFor() which is more better than wait()
- To get latest testing-lib version, I had to upgrade jest (in order to get jsdom-env-sixteen by default)
- Latest jest... upgrades JSDOM (and JSDOM upgrades lot of stuff on css). So this is why we now have the css variables in our snapshots. It seems also that we have changes on font-family 🤷‍♂️ 

- Fix ImageLoader : Before, when we had an error, the error was not correctly throws so we didn't follow the right path now it's clearer: 
  - First: we try to load the image from file.links if it exists 
  - Second: if we have an error during the mount of the image (links expired), we make a request to fetch the links 
  - Third: if we still have an error, let's make a call to the download link to get the binary


react/__snapshots__/examples.spec.jsx.snap
```diff
@@ -101,16 +101,16 @@ exports[`AppTitle should render examples: AppTitle 1`] = `
 exports[`Avatar should render examples: Avatar 1`] = `
 "<div>
   <div class=\\"u-flex\\">
-    <div class=\\"styles__c-avatar___PpDI-\\"><svg viewBox=\\"0 0 16 16\\" class=\\"styles__icon___23x3R\\" width=\\"16\\" height=\\"16\\">
+    <div class=\\"styles__c-avatar___PpDI-\\" style=\\"--circleSize: 40px;\\"><svg viewBox=\\"0 0 16 16\\" class=\\"styles__icon___23x3R\\" width=\\"16\\" height=\\"16\\">
```
JF-Cozy, line 104: C'est normal qu'on rajoute `--circleSize: 40px;` partout ?

JF-Cozy (approved):

flohhhh:
> nit: img = null must be sufficient.
> clear img.onload = img.onerror = null and img.src = '' ?

have you looked at this point ?

Crash--:
not yet @flohhhh will do later :)

PR #1795, opened 2021-04-07
fix: UploadQueue header style

JF-Cozy:
Harmonisation du style entre la phase d'`upload` et la phase `done`. Ajout de l'état `done` dans les screenshots.

Besoin de ce correctif pour Drive.

Exemple UI : https://jf-cozy.github.io/cozy-ui/react/#!/UploadQueue

Screenshot Drive : 

**Avant :**
image

image

**Après :**
image

image

### DIFF
--- a/react/UploadQueue/Readme.md
+++ b/react/UploadQueue/Readme.md
@@ -9,7 +9,7 @@ the upload queue:
 * in a fixed bottom-right position on desktop
 * underneath the top-bar on mobile
 
-```
+```jsx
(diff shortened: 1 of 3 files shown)


Crash--:
The example of the component should really reflect its use. Having an use case with a table and a table header, like in Drive and in the screenshots you posted, can be really useful to prevent this kind of regression.

Crash-- (approved):

ptbrowne (approved):

PR #1796, opened 2021-04-08
feat: Show alerts on top of mui dialogs

ptbrowne:
Previously, the z-index was too low and alerts would appear beneath
the MUI Dialog overlay.

See https://ptbrowne.github.io/cozy-ui/react/#/Content?id=cozydialogs

### DIFF
--- a/react/CozyDialogs/Readme.md
+++ b/react/CozyDialogs/Readme.md
@@ -35,6 +35,7 @@ import {
 import { BreakpointsProvider } from 'cozy-ui/transpiled/react/hooks/useBreakpoints'
 
 import Button from 'cozy-ui/transpiled/react/Button'
+import Alerter from 'cozy-ui/transpiled/react/Alerter'
 import Icon from 'cozy-ui/transpiled/react/Icon'
 import Typography from 'cozy-ui/transpiled/react/Typography'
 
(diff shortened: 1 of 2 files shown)


JF-Cozy (approved):

stylus/settings/z-index.styl
```diff
@@ -49,7 +49,7 @@ $drawer-index = 60
 $modal-index = 70
 $modal-footer-index = $modal-index + 1
 $modal-toolbar-index = $modal-footer-index + 1
-$alert-index = 80
+$alert-index = 1400 // z-index for MUI dialogs is 1300, alerts need to be on top
```
Crash--, line 52: It seems that we force the z-Index for MUI dialogs here https://github.com/cozy/cozy-ui/blob/master/react/MuiCozyTheme/theme.jsx#L421

It doesn't work anymore?

ptbrowne: Mmh good point, it seems like a better solution. And now, right now it does not work, I am looking into it, thanks

image

ptbrowne: Fixed :) https://ptbrowne.github.io/cozy-ui/react/#/Content?id=cozydialogs

https://github.com/cozy/cozy-ui/pull/1796/commits/92994264382f66cea16ffe49dc51b892f0837aec

Crash-- (approved):
Thanks!

PR #1803, opened 2021-04-19
feat: Add file-type-note icon

JF-Cozy:
So we can remove it from apps assets (like Drive)

the svg file is coming from Drive


Crash--:
I checked on Cozy Notes because it seems weird that this icon was not in UI. 

But you're right, we've a manual importation ... 🤦 https://github.com/cozy/cozy-notes/blob/master/src/components/notes/List/NoteRow.jsx#L16

Crash-- (approved):

PR #1813, opened 2021-05-06
feat: Ability to remove a notification programmatically

ptbrowne:
### API changes

Alerter has a new method removeNotification that can be used to
programmatically remove a notification that has been shown.

### Internal state changes

Previously, there would be two states, one in the store, and one in
the Alerter. Now there is only the state of the store, and the Alerter
keeps a reference on it.

Ids would be kept as the current length of the list but this could be
problematic since you could have a situation where two notifications
have the same id. Now we have an incremental id.


ptbrowne:
MMh this is weird I cannot reproduce the too strong shadow problem.

JF-Cozy:
@ptbrowne je ne reproduis pas non plus

react/Alerter/Alert.jsx
```diff
@@ -18,14 +18,16 @@ export class Alert extends Component {
...
   componentDidMount() {
-    this.closeTimer = setTimeout(
-      () => {
-        this.beginClosing()
-      },
-      this.props.duration === 'auto'
-        ? this.computeDuration()
-        : this.props.duration
-    )
+    if (this.props.duration !== Infinity) {
```
JF-Cozy, line 21: nit: `{ duration } = this.props `

JF-Cozy (approved):

flohhhh (approved):

PR #1820, opened 2021-05-10
fix: CozyDialogs Title width and close button

JF-Cozy:
- fix Title width on desktop and mobile
- adjust Title width according to close button
- remove back button if no close fn

### DIFF
--- a/react/CozyDialogs/ConfirmDialog.jsx
+++ b/react/CozyDialogs/ConfirmDialog.jsx
@@ -32,7 +32,7 @@ const ConfirmDialog = props => {
       <DialogContent>
         <div className="dialogContentInner withFluidActions">
           <DialogTitle {...dialogTitleProps} className="dialogTitleFluid">
-            {fullScreen ? <DialogBackButton onClick={onClose} /> : null}
+            {fullScreen && onClose && <DialogBackButton onClick={onClose} />}
             {title}
           </DialogTitle>
           {content}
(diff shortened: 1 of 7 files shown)


react/CozyDialogs/useCozyDialog.js
```diff
@@ -1,7 +1,21 @@
...
+  return {
+    dialogTitle: {
+      width: ({ onClose }) =>
+        onClose ? 'calc(100% - 88px)' : 'calc(100% - 64px)',
+      [theme.breakpoints.down('sm')]: {
+        width: 'calc(100% - 32px)'
+      }
+    }
+  }
+})
```
Crash--, line 17: The style related to our Dialog is already defined in theme.jsx. We should follow this path. Now, with this PR we have several files to look at. We should avoid that. 

Also, can we know where these values (88 / 64 / 32) come from? Should they be related / linked to other margins / variables?

JF-Cozy: > The style related to our Dialog is already defined in theme.jsx. We should follow this path. Now, with this PR we have several files to look at. We should avoid that.

Oui mais la taille dépend d'une prop, on ne peut pas faire autrement si ?

Crash--: If we take a look at how we handle that for now: 

- We add a classname : https://github.com/cozy/cozy-ui/blob/master/react/CozyDialogs/IllustrationDialog.jsx#L43
- We apply a style depending of the classname https://github.com/cozy/cozy-ui/blob/master/react/MuiCozyTheme/theme.jsx#L505 

no?

Crash-- (approved):

PR #1825, opened 2021-05-27
feat: Put styles as part of theme

ptbrowne:
This way ListItemIcon is only re-exported and we have access to
its type definition in apps

### DIFF
--- a/react/MuiCozyTheme/ListItemIcon/index.js
+++ b/react/MuiCozyTheme/ListItemIcon/index.js
@@ -1,13 +1,7 @@
-import { withStyles } from '@material-ui/core/styles'
 import ListItemIcon from '@material-ui/core/ListItemIcon'
 
 export const smallSize = 16
 export const mediumSize = 24
 export const largeSize = 32
 
-export default withStyles({
-  root: {
-    width: 32,
-    justifyContent: 'center'
-  }
-})(ListItemIcon)
+export default ListItemIcon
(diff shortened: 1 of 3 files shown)


Ldoppea:
I don't think this is critical but this PR also changes the behaviour of the hot reload feature.

Using the old `withStyles` syntax seems to hot reload only the CSS, it's a bit faster and the user doesn't loose its view state. Here is an exemple : after the edit the style is applied without reloading all the page.

https://user-images.githubusercontent.com/1884255/120214547-e9b56800-c234-11eb-83d1-833fee6349f6.mp4


But with the new way, all the page is reloaded and the user has to manually scroll back to where he was.

https://user-images.githubusercontent.com/1884255/120214587-f639c080-c234-11eb-9ad4-359bb8bf2f0f.mp4

However on smaller screen sizes it succeed to automatically scroll back to where the user was (I did not investigate why) but it still reload all the page.

https://user-images.githubusercontent.com/1884255/120215110-a5769780-c235-11eb-82bb-c4aa5eb65720.mp4

If you consider this side effect is not an issue then I agree to merge this PR.

ptbrowne:
This is expected since the theme file is not exporting only a React component, hot reloading cannot work and the whole page has to be reloaded. The problem that you saw can be circumvented by focusing only on the component by clicking on the small "fullscreen" icons to the top right of each section/component.

Ldoppea (approved):

Ldoppea:
Ok then let's approve this PR :)

PR #1829, opened 2021-06-03
Add maintenance display app sections

flohhhh:
Add maintenance display in props

This display is similar to what we have already in cozy home, when a konnector is in maintenance.

By default: maintenanceDisplay is `false` (allow to keep current behavior)

Capture d’écran 2021-06-03 à 17 25 55

[https://flohhhh.github.io/cozy-ui/react/#/Special?id=appsections](https://flohhhh.github.io/cozy-ui/react/#/Special?id=appsections)

### DIFF
--- a/react/AppSections/components/AppsSection.jsx
+++ b/react/AppSections/components/AppsSection.jsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import { useI18n } from '../../I18n'
 import useBreakpoints from '../../hooks/useBreakpoints'
 import { getTranslatedManifestProperty } from '../helpers'
@@ -12,7 +13,8 @@ export const AppsSection = ({
   appsList,
   subtitle,
   onAppClick,
-  IconComponent
+  IconComponent,
+  displaySpecificMaintenanceStyle
 }) => {
   const { isMobile } = useBreakpoints()
   const { t } = useI18n()
@@ -29,6 +31,7 @@ export const AppsSection = ({
               onClick={() => onAppClick(app.slug)}
               key={app.slug}
               showDeveloper={!isMobile}
+              displaySpecificMaintenanceStyle={displaySpecificMaintenanceStyle}
               IconComponent={IconComponent}
             />
           ))}
@@ -36,6 +39,14 @@ export const AppsSection = ({
       )}
     </div>
   )
+}
+
+AppsSection.propTypes = {
+  appsList: PropTypes.array,
+  subtitle: PropTypes.element,
+  onAppClick: PropTypes.func,
+  IconComponent: PropTypes.element,
+  displaySpecificMaintenanceStyle: PropTypes.bool
 }
 
 export default AppsSection
--- a/react/AppTile/index.jsx
+++ b/react/AppTile/index.jsx
@@ -78,6 +78,7 @@ export const AppTile = ({
         />
         {isInMaintenanceWithSpecificDisplay && (
           <Icon
+            data-testid="icon-maintenance"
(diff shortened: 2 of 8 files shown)


Crash--:
If it similar to what we've in the Home App, can we refactor the Home App to use what you propose? 

After a quick search, I found out https://github.com/cozy/cozy-ui/pull/1749 Shouldn't we use the same api?

flohhhh:
For the Home App, it's a specific component (KonnectorTile) like ours (AppTile).

The PR #1749 offers the possibility of displaying the status. In our case it allows to display a style.

The question is do we want to generalize it on all apps that use AppSections?
In this case, we can directly display the style when the connector is in maintenance without having a prop that allows to display or not the style.

What do you think ?

react/AppSections/components/AppsSection.jsx
```diff
@@ -38,4 +41,8 @@ export const AppsSection = ({
   )
 }
 
+AppsSection.propTypes = {
+  maintenanceDisplay: PropTypes.bool
```
Crash--, line 45: displaySpecificMaintenanceStyle ?

react/AppTile/index.jsx
```diff
@@ -52,14 +55,30 @@ export const AppTile = ({
     ? showStatus.indexOf(statusLabel) > -1 && statusLabel
     : showStatus && statusLabel
   IconComponent = IconComponent || AppIcon
+  const isInMaintenance = maintenanceDisplay && statusLabel === 'maintenance'
```
Crash--, line 58: It is not enough. Take a look at `isUnderMaintenance`

react/AppTile/index.jsx
```diff
@@ -52,14 +55,30 @@ export const AppTile = ({
...
+        isInMaintenance ? styles['AppTile-container-maintenance'] : undefined
+      }
+    >
       <TileIcon>
         <IconComponent
           app={app}
           className={styles['AppTile-icon']}
           {...getAppIconProps()}
         />
+        {isInMaintenance && (
```
Crash--, line 75: Also not a big fan of having severals const for the same thing. We should avoid to rename things. 

This is not `isInMaintenance` then we display... This is `isInMaintenance && displaySpecificMaintenanceStyle` . We should not mix the two concept

react/AppTile/index.jsx
```diff
@@ -52,14 +55,30 @@ export const AppTile = ({
...
+          <Icon
+            data-testid="icon-maintenance"
+            icon={WrenchCircleIcon}
+            className={`${styles['AppTile-icon-maintenance']} u-coolGrey`}
```
Crash--, line 79: Why this `u-coolGrey`? By default `<Icon` will use `icon` classname https://github.com/cozy/cozy-ui/blob/master/react/Icon/index.jsx#L58 which is https://github.com/cozy/cozy-ui/blob/master/react/Icon/styles.styl#L4 which is 

`currentColor` https://github.com/cozy/cozy-ui/blob/master/stylus/settings/palette.styl#L246 

So: 
- Maybe Icon shouldn't be colored by default by currentColor but by secondaryTextColor ? (cc @joel-costa) (Like we do for invertedThem) 
- let's use the color props instead of an utility class to be more declarative ;)

Crash--:
Also, instead of using custom "variant", can you use `Variants` provided by Cozy UI?

react/AppTile/index.jsx
flohhhh: Hmm above there is getCurrentStatusLabel (statusLabel)  which calls `isUnderMaintenance` (which returns 'maintenance'). 

So it's ok.

PR #1862, opened 2021-08-06
fix: rename createMuiTheme

acezard:
BREAKING CHANGE: `@material-ui/core` has been updated to v4.12

Because `@material-ui/core` is a peerDependency, it was possible for consumers of this library to use a v4.12 mui version while cozy-ui was still using 4.11. This created an error because v4.11 `createMuiTheme` function was renamed to `createTheme` in mui v4.12. While this does not break the app, because mui semver is correctly used and allows apps to work with the depreciated function, it still throws an error in the console and does not look good. Moreover the function will most likely be deleted in a future major upgrade.

It will be necessary here to upgrade your version to match at least v4.12.0

update: will be handled more elegantly there: https://github.com/cozy/cozy-ui/issues/1865

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "@babel/cli": "7.11.5",
     "@babel/core": "7.11.5",
     "@cozy/codemods": "^1.9.0",
-    "@material-ui/core": "4.11.4",
+    "@material-ui/core": "4.12.3",
     "@semantic-release/changelog": "5.0.1",
     "@semantic-release/git": "7.0.18",
     "@semantic-release/npm": "5.3.4",
@@ -157,7 +157,7 @@
     "react-swipeable-views": "^0.13.3"
   },
   "peerDependencies": {
-    "@material-ui/core": "4",
+    "@material-ui/core": ">=4.12.0",
     "cozy-client": ">=18.1.2",
     "cozy-device-helper": "^1.10.0",
     "cozy-doctypes": "^1.69.0",
--- a/react/MuiCozyTheme/theme.jsx
+++ b/react/MuiCozyTheme/theme.jsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { createMuiTheme } from '@material-ui/core/styles'
+import { createTheme } from '@material-ui/core/styles'
 import merge from 'lodash/merge'
 
 import { getCssVariableValue } from '../utils/color'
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -157,7 +157,7 @@
     "react-swipeable-views": "^0.13.3"
   },
   "peerDependencies": {
-    "@material-ui/core": "4",
+    "@material-ui/core": "^4.12.0",
```
Crash--, line 160: >= 4.12.0 (cc @Ldoppea 😘 )

package.json
```diff
@@ -68,7 +68,7 @@
     "@babel/cli": "7.11.5",
     "@babel/core": "7.11.5",
     "@cozy/codemods": "^1.9.0",
-    "@material-ui/core": "4.11.4",
+    "@material-ui/core": "4.12.3",
```
Crash--, line 71: Since this is a breaking change, the commit should be tagged as being a breaking change. https://github.com/cozy/cozy-guidelines#commit-messages

acezard: Updated though I'm not sure how to provide a migration path. The updated peerDependency should take care of it. Worst case scenario, everything still works but the thrown error still shows in the console.

Crash--: Worst scenario = crash since createTheme doesn't exist in previous version. 

This is not because you update the peerDep that the app consuming will automatically updates its dep. It'll throw a warning in the terminal saying "unmeet peerdep" that's all. And since there is a lot of message like this currently, app's dev will not see it.

acezard: I think the simplest way to handle that would be to have material-ui as a real dependency. Otherwise I guess the best move is to upgrade cozy-ui to a new major version since the breaking change doesn't "fix itself" (the consumer will have to update his own material ui dep).

Crash--: Since you put a BREAKING CHANGE in the commit message, it'll publish a new major version ;) 

https://github.com/cozy/cozy-ui/issues/1865

acezard: Ah that's nice 😅 I updated the commit message to reflect what action is necessary after updating

JF-Cozy: Pas de souci avec la 4.12.3 ? Sur la doc https://material-ui.com/getting-started/installation/ on est encore en 4.12.2...

acezard: J'avoue que sur les patchs en général je pars du principe que c'est sans risque, même sur les minor en fait, mais comme vous préférez j'ai juste get @latest

JF-Cozy: Ok. Et pas d'autre BC entre la 4.11 et la 4.12 que ce que tu as modifié ?

PR #1875, opened 2021-08-18
fix: Change text orientation in MidEllipsis

acezard:
Related to https://trello.com/c/AcrjAGzm/1498-vue-recent-chemin-mal-d%C3%A9coup%C3%A9

There were multiple visual "bugs" in the MidEllipsis component, related to how the ellipsis is handled, particularly the second span containing the second half of the truncated text. In order to have the text-overflow hidden from the left instead of from the right, that span was in the RTL direction. The problem is that multiple characters will then be switched to the start of the span, like punctuation and emojis. In order to fix this, we used left to right mark HTML entity to treat the content of the span as LTR, but the span itself remains RTL so the overflow management still works as intended.

In any case, there is an enormous amount of possibilities of folder names, it is impossible to be 100% sure that this fix will work for all scenarios. A true infallible solution would require a JS algorithm, but Drive for instance can not allow this kind of computation on such repeated elements.

image

image

https://acezard.github.io/cozy-ui/react/#!/MidEllipsis


JF-Cozy (commented):
Pas mal l'astuce ! Est-ce qu'on peut imaginer avoir des tests avec react testing lib pour vérifier quelques cas typiques ?

Crash-- (changes requested):
can you add sample / test inside the readme in order to have screenshot for these use cases:
- /Administratif/Netflix
- /Administratif/Ameli//1 88 88 88 888 888
- /Partages reçus/Cozy 🗄 - Team/Customers & Partners 🛒/Xxxxxx (Xxxxxxxx)/4_Suivi opérationnel/Point de synchro

acezard:
> can you add sample / test inside the readme in order to have screenshot for at least these use cases:
> 
> * /Administratif/Netflix
> * /Administratif/Ameli//1 88 88 88 888 888
> * /Partages reçus/Cozy 🗄 - Team/Customers & Partners 🛒/Xxxxxx (Xxxxxxxx)/4_Suivi opérationnel/Point de synchro

sure thing

acezard:
> Pas mal l'astuce ! Est-ce qu'on peut imaginer avoir des tests avec `react testing lib` (ou jest simplement) pour vérifier quelques cas typiques ?

J'add des screenshots. La nature quasi 100% css de ce component (raisons de perf) le rend très difficile à tester programmatiquement selon moi. Enfin en tout cas c'est une tâche qui me parait complexe

Crash-- (approved):

PR #1881, opened 2021-08-26
fix: Remove `height:auto` from `app-2panes-sticky` class

Ldoppea:
This class is used when the UI is split in two panes, usually when the app has a sidebar and a main content

When using `height:auto`, this prevents the main content to be scrollable if it is smaller than the navbar. Also the navbar is not scrollable at all as its height won't adapt to the browser size

Another effect is that the main content is not scrollable to the end of document when the navbar overflows the document as it virtually push the document's limits outside of the browser view

This behavior is visible only in `cozy-store` as it is the only app that have a navbar tall enough to overflow the browser's body

Other apps are not impacted as their navbar is very small

By removing `height:auto`, then the navbar height adapt to its container size (vs its own size). This allow the main content to be scrollable when smaller

This fix should be completed by some CSS edit on apps side with an `overflow:auto` on the sidebar. But here again only `cozy-store` is impacted for now

If your changes have graphic impacts, it is useful to deploy a version of the styleguidist to your repository.

⚠️ this edit is not visible in docs, so I didn't upload them for previsualization

### Before (main content smaller than navbar):
Neither navbar nor main content are scrollable
image

### After (main content smaller than navbar):
Both navbar and main content are scrollable
image

### Before (main content taller than navbar):
Navbar not scrollable and main content's scroll bar is overflowing the document view and we cannot scroll more.
image

### After (main content taller than navbar):
Navbar is scrollable and main content's can now be fully scrolled
image

### DIFF
--- a/stylus/objects/layouts.styl
+++ b/stylus/objects/layouts.styl
@@ -113,7 +113,6 @@ $app
 $app-2panes-sticky
     @extend $app
     flex 0 0 100%
-    height auto
     align-items stretch
 
     & > aside


JF-Cozy:
> This behavior is visible only in cozy-store as it is the only app that have a navbar tall enough to overflow the browser's body

Tu as testé sur Drive j'ai vu, avec une petite et une grande fenêtre de navigateur ?

Si tu n'es pas serein, ne peut-on pas plutôt faire la modif dans l'app `store` ? (si tu es serein on peut le passer dans Ui ;) )

acezard (approved):

Ldoppea:
> > This behavior is visible only in cozy-store as it is the only app that have a navbar tall enough to overflow the browser's body
> 
> Tu as testé sur Drive j'ai vu, avec une petite et une grande fenêtre de navigateur ?
> 
> Si tu n'es pas serein, ne peut-on pas plutôt faire la modif dans l'app `store` ? (si tu es serein on peut le passer dans Ui ;) )

Oui testé avec plusieurs tailles de fenêtres et aussi en mode mobile pour checker les éventuels effets de bords.

Avec mes connaissances actuelles je suis assez serein si d'autant que l'app store est la seule application avec un sidebar assez grand pour devenir scrollable sans aller dans des ratio d'écrans extravagants. Mais est-ce qu'il peut y avoir des apps que je n'ai jamais vu (genre app spécifiques à certains clients)?

JF-Cozy (approved):

JF-Cozy:
> Mais est-ce qu'il peut y avoir des apps que je n'ai jamais vu (genre app spécifiques à certains clients)?

Pas à ma connaissance

PR #1888, opened 2021-09-01
Fix/data-cozy-*

Ldoppea:
This PR is a draft as some questions are still open and those may not be processed in the next few days.

This PR fixe the usage of `data-cozy` to reflect recent changes on how it is injected by the cozy-bar:
- https://github.com/cozy/cozy-store/pull/740
- https://github.com/cozy/cozy-banks/pull/2176
- https://github.com/cozy/cozy-drive/pull/2385

Points to be checked first:
- I'm not sure about the edits on tracker.jsx as I didn't check yet all app that use it and also how userId should be handled
- The fixed code in `AppTile` seems not to be used anymore. It may be used if the caller injects a custom `IconComponent`. As this `IconComponent` may be injected from an App, I should check all Cozy apps first and maybe the solution will be to remove this code instead of fixing it.
  - Cozy Store doest not use this mechanism
  - Other apps should be checked
  
____
  
Todo:
- [x] make code backward compatible
- [ ] create an issue to remove potential dead code

### DIFF
--- a/react/AppTile/index.jsx
+++ b/react/AppTile/index.jsx
@@ -28,12 +28,14 @@ let dataset
 const getDataset = () => {
   if (dataset) return dataset
   const root = document.querySelector('[role=application]')
-  dataset = root && root.dataset
+  if (root) {
+    dataset = root.dataset.cozy ? JSON.parse(root.dataset.cozy) : root.dataset
+  }
   return dataset
 }
 
 const getAppIconProps = () => ({
-  domain: getDataset() && getDataset().cozyDomain,
+  domain: getDataset() && (getDataset().cozyDomain || getDataset().domain),
   secure: window.location.protocol === 'https:'
 })
 
--- a/react/helpers/tracker.jsx
+++ b/react/helpers/tracker.jsx
@@ -87,8 +87,8 @@ export const configureTracker = (options = {}) => {
   const root = document.querySelector('[role=application]')
 
   if (root && root.dataset) {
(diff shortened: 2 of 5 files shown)


react/helpers/tracker.jsx
```diff
@@ -87,8 +87,8 @@ export const configureTracker = (options = {}) => {
   const root = document.querySelector('[role=application]')
 
   if (root && root.dataset) {
-    appName = readCozyDataFromDOM('appName')
-    cozyDomain = readCozyDataFromDOM('cozyDomain')
+    appName = readCozyDataFromDOM('app').name
+    cozyDomain = readCozyDataFromDOM('domain')
```
Crash--, line 91: We should do something more backward compatible : 

readCozyDataFromDOM('appName') || readCozyDataFromDOM('app').name etc.

react/AppTile/index.jsx
```diff
@@ -28,12 +28,12 @@ let dataset
 const getDataset = () => {
   if (dataset) return dataset
   const root = document.querySelector('[role=application]')
-  dataset = root && root.dataset
+  dataset = root && JSON.parse(root.dataset.cozy)
```
Crash--, line 31: Same here, we should be backward compatible

Ldoppea:
New attempt with backward compatible code. Can you double check unit tests to be sure that I covered the expected rules?

rezk2ll:
@Crash-- @JF-Cozy do we still need this?

JF-Cozy:
No

Issue #1880, opened 2021-08-25
Screenshots are sometime taken during a component 'entrance' animation


Ldoppea:
Note: I did not find a build that reproduce this issue on cozy/cozy-ui, but it often happens on my fork Ldoppea/cozy-ui.

Some component have an entrance animation, for example NestedSelect documentation on mobile show a select drawer that enter the screen from the bottom and then go fullscreen.

In this case the screenshot may be taken during the entrance animation and so a diff is detected as the component is randomly placed somewhere between the bottom state and the fullscreen state :

![image](https://user-images.githubusercontent.com/1884255/130776282-34c7a100-eb87-4616-9886-84b9a11858d5.png)

Reproduction case can be seen in this build from my fork : https://www.argos-ci.com/Ldoppea/cozy-ui/builds/25 (see `HideNestedSelect-inverted-300x600.png`, `ContactsListModal-inverted-300x600.png` and `NestedSelect-inverted-800x600.png`)

Ldoppea:
Certainly related to https://github.com/cozy/cozy-ui/issues/1839

Crash--:
Animation should be disabled when screenshoting to avoid such issue. See https://github.com/cozy/cozy-ui/pull/1394 for a basic implem. We should have missed some components

PR #1894, opened 2021-09-08
fix: Desactivate MUI transition for inverted theme if is testing

Ldoppea:
In testing environment, MUI transitions were disabled for normal theme
but not for inverted one

Related commit: 5dc817a50e4ebb4d612ebc3d0d002de3664a9ce6

This may fix https://github.com/cozy/cozy-ui/issues/1880


Crash-- (approved):
🎉

PR #1959, opened 2021-11-05
docs: Use variants for ActionMenu

JF-Cozy:
to simplify documentation
see https://jf-cozy.github.io/cozy-ui/react/#!/ActionMenu

### DIFF
--- a/react/ActionMenu/Readme.md
+++ b/react/ActionMenu/Readme.md
@@ -1,225 +1,106 @@
 Use an ActionMenu to show a list of actions. ActionMenus automatically switch their display between desktop and mobile.
 
-### Classic
+### Example
 
-You can pass a reference to a custom DOM element through the `anchorElRef` prop to attach the menu to that element. This is useful is you want to be able to have autoclose true and be able to close the menu by clicking on the same component that opens it. 
+You can pass a reference to a custom DOM element through the `anchorElRef` prop to attach the menu to that element. This is useful is you want to be able to have autoclose true and be able to close the menu by clicking on the same component that opens it.
+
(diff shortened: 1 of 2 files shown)


react/ActionMenu/Readme.md
```diff
@@ -2,224 +2,105 @@ Use an ActionMenu to show a list of actions. ActionMenus automatically switch th
 
 ### Classic
```
Ldoppea, line 3: Now that there is only one category, then having `Classic` as section name seems not to be relevant

If I read `Classic` my first reaction is "hey are there other modes available?"

react/ActionMenu/Readme.md
```diff
@@ -2,224 +2,105 @@ Use an ActionMenu to show a list of actions. ActionMenus automatically switch th
...
-      </ActionMenuHeader>
-      <ActionMenuItem onClick={() => alert('click')}left={<Icon icon={FileIcon} />}>Item 1</ActionMenuItem>
-  </ActionMenu>}
-</BreakpointsProvider>
-```
-
-### Placement on desktop
+A header can be used to provide context on the menu actions. Since on desktop, we display a popper and not a BottomDrawer, context for the user is not lost, so the ActionMenuHeader would be redundant. This is why it is not rendered unless we are on mobile.
 
 We use [popper.js](https://popper.js.org/docs/v2/) under the hood. You can use the `popperOptions` prop to pass options to the popper.js instance. This lets you control things like placement relative to the anchor, positioning strategies and more — refer to the popper doc for all the details.
```
Ldoppea, line 9: By removing the title we loose the info about this option being useful only on desktop mode.

Maybe we should mention it at the end of paragraph as for the header displayed only on mobile?
Or maybe we can change the the variant name for `withDesktopPlacement` (in opposition to `withMobileHeader`)?
Or both 😄

Ldoppea (approved):

PR #1969, opened 2021-11-12
fix: Add viewBox to LogoutLarge icon

doubleface:
to get a better display in different sizes

### DIFF
--- a/react/Icons/LogoutLarge.jsx
+++ b/react/Icons/LogoutLarge.jsx
@@ -3,7 +3,7 @@ import React from 'react'
 
 function SvgLogoutLarge(props) {
   return (
-    <svg {...props}>
+    <svg viewBox="0 0 44 44" {...props}>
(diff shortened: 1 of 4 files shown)


react/Icons/LogoutLarge.jsx
```diff
@@ -3,7 +3,7 @@ import React from 'react'
 
 function SvgLogoutLarge(props) {
   return (
-    <svg {...props}>
+    <svg viewBox="0 0 44 44" {...props}>
```
Crash--, line 6: We should not manually edit those files! The header says: 
```
// Automatically created, please run `scripts/generate-svg-icon.sh assets/icons/illus/logout-large.svg` to regenerate;
```

So you need to change the source file to get this viewBox!

Crash--: @doubleface

PR #1971, opened 2021-11-12
feat: Add type prop to SquareAppIcon

doubleface:
This property is based on AppIcon's type property

It is needed to force the type of konnector icons, or else AppIcon will
try to get their icon from the stack first with the wrong app type
This leads to lots of 404 errors in the home application.

Only after that, AppIcon tries to get the icon from the registry and
succeeds.

### DIFF
--- a/react/SquareAppIcon/index.jsx
+++ b/react/SquareAppIcon/index.jsx
@@ -61,7 +61,7 @@ const useStyles = makeStyles(theme => ({
   }
 }))
 
-export const SquareAppIcon = ({ app, name, variant, IconContent }) => {
+export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
   const classes = useStyles()
   const appName = name || (app && app.name) || app || ''
   const letter = appName[0] || ''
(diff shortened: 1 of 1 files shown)


react/SquareAppIcon/index.jsx
```diff
@@ -119,7 +119,7 @@ export const SquareAppIcon = ({ app, name, variant, IconContent }) => {
               ) : IconContent ? (
                 IconContent
               ) : (
-                <AppIcon app={app} />
+                <AppIcon app={app} type={type} />
```
Crash--, line 122: I'm not a big fan of adding new prop like this. We can now set the `type` but what about other AppIcon's props? https://github.com/cozy/cozy-ui/blob/master/react/AppIcon/index.jsx#L136-L149 . I'm confident that you open a PR today to pass the `priority` props to SquareAppIcon. And we should not do that. 

I'm wondering if we should not pass an `appIconProps` to SquareAppIcon to be able to pass the props we want to 

cc @doubleface @JF-Cozy

doubleface: looks like appIconProps is a good idea

PR #1968, opened 2021-11-12
feat: Add IconContent prop to SquareAppIcon

doubleface:
To allow a custom icon or even a grid of icon to be displayed

This prop will be used to display category candidates (with a grid of icons) and the exit button with a custom icon on the home application.

image

- [ X] [Deployed the styleguidist](https://refined-github-html-preview.kidonng.workers.dev/doubleface/cozy-ui/raw/gh-pages/react/index.html#!/SquareAppIcon)
- [ ] Did you think of ARIA attributes if you are coding new components ?

PR #1972, opened 2021-11-16
fix: Update LogoutLarge icon properly from the source

doubleface:
This is a fix of https://github.com/cozy/cozy-ui/pull/1968#discussion_r750010959


Crash-- (approved):
Can you write an explanation please? We should always have the context in the commit message

PR #1973, opened 2021-11-16
feat: Add appIconProps to SquareAppIcon

doubleface:
To configure to behavior of AppIcon, in SquareAppIcon, we need access to
all AppIcon properties. SquareAppIcon does not have access to the
context of the app enough to make choices regarding :

Ex :
- priority : stack or registry, to avoid too many 404 requests
in the app

Now SquareAppIcon takes it's own properties and forwards the remaining
ones directly to AppIcon

This way, when the AppIcon changes, we do not need to change
SquareAppIcon.

### DIFF
--- a/react/SquareAppIcon/index.jsx
+++ b/react/SquareAppIcon/index.jsx
@@ -2,13 +2,13 @@ import React from 'react'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import { makeStyles } from '@material-ui/styles'
+import get from 'lodash/get'
 
 import AppIcon from '../AppIcon'
 import Badge from '../Badge'
 import InfosBadge from '../InfosBadge'
 import { nameToColor } from '../Avatar'
 import Typography from '../Typography'
-import { AppDoctype } from '../proptypes'
 import Icon from '../Icon'
 import iconPlus from '../Icons/Plus'
 import iconWarning from '../Icons/WarningCircle'
@@ -61,9 +61,15 @@ const useStyles = makeStyles(theme => ({
   }
 }))
 
-export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
+export const SquareAppIcon = ({
+  name,
+  variant,
+  IconContent,
+  ...appIconProps
+}) => {
   const classes = useStyles()
-  const appName = name || (app && app.name) || app || ''
+  const appName =
+    name || get(appIconProps, 'app.name') || get(appIconProps, 'app') || ''
   const letter = appName[0] || ''
 
   return (
@@ -119,7 +125,7 @@ export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
               ) : IconContent ? (
                 IconContent
               ) : (
-                <AppIcon app={app} type={type} />
+                <AppIcon {...appIconProps} />
               )}
             </div>
           )}
@@ -137,7 +143,6 @@ export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
 }
 
 SquareAppIcon.propTypes = {
-  app: PropTypes.oneOfType([AppDoctype, PropTypes.string]),
   name: PropTypes.string,
   variant: PropTypes.oneOf([
     'ghost',
@@ -146,8 +151,7 @@ SquareAppIcon.propTypes = {
     'add',
     'shortcut'
   ]),
-  IconContent: PropTypes.node,
-  type: PropTypes.oneOf(['app', 'konnector'])
+  IconContent: PropTypes.node
 }
 
 export default SquareAppIcon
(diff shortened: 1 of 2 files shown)


react/SquareAppIcon/index.jsx
```diff
@@ -61,9 +62,10 @@ const useStyles = makeStyles(theme => ({
   }
 }))
 
-export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
+export const SquareAppIcon = ({ app, type, name, variant, IconContent, appIconProps = {} }) => {
+  // "app" and "type" properties are deprecated in favor of appIconProps
   const classes = useStyles()
-  const appName = name || (app && app.name) || app || ''
+  const appName = name || (get(appIconProps, 'app.name')) || (app && app.name) || app || ''
```
Crash--, line 68: why don't we do the same for appSlug? 

If I read the code correctly, we style need to use `app` props to specify the slug right? 

So with that change https://github.com/cozy/cozy-ui/pull/1973/files#diff-7bbce49ffc2d1b4a041913caac7d27c2d1812a321158075dc1094df8978459b6R21 we pass an `undefined` to <AppIcon app={undefined}` right?

doubleface: The app slug is only used by AppIcon. I don't understand why we would need to specify the slug.

> we pass an undefined to <AppIcon app={undefined} right?

    app = { name: "Test App", slug: "testapp", type: "app" }

and not `undefined`. I don't understand

JF-Cozy: Si "app" and "type" sont dépréciés, pourquoi ne pas les supprimer et faire un BC ? A mon avis on peut se le permettre, le composant étant assez jeune et (chez nous) uniquement utilisé dans la home. Il est déjà assez compliqué comme ça...

Ca nous donnerait simplement `const appName = name || (get(appIconProps, 'app.name')) || ''`

Crash--: @doubleface 

here: https://github.com/cozy/cozy-ui/pull/1973/files#diff-fce2b54630424a9a413c69c08723d6e83ffc43e4ce8f8983c30325f9b97a4c7eR124 we pass to `<AppIcon ` `app={app}`

`app` comes from the props here:
https://github.com/cozy/cozy-ui/pull/1973/files#diff-fce2b54630424a9a413c69c08723d6e83ffc43e4ce8f8983c30325f9b97a4c7eR65

if you follow the sample you wrote here: https://github.com/cozy/cozy-ui/pull/1973/files#diff-7bbce49ffc2d1b4a041913caac7d27c2d1812a321158075dc1094df8978459b6R21

we don't pass any `app` props. So we're calling `<AppIcon` with `app={undefined}` no? And `AppIcon` need the slug in `app` props, right? 

At the end you're calling `<App app={undefined} type={type} slug={'myslug'} type={type} />` no? 

I'm wondering if we should not pass the AppIcon props directly to the SquareAppIcon without adding this `appIconProps` since there only `variant, IconContent` that is specific to SquareAppIcon. The result will be: 

```
const app = { name: "Test App", slug: "testapp", type: "app" }
<SquareAppIcon variant={XXX} IconComponent={CCC} {...app} />
```

I find it easier to understand. For instance this is how MUI works: Dialog can also get the props of Modal https://v4.mui.com/api/dialog/#inheritance

Crash--:
ping @doubleface

Crash--:
ping @doubleface

react/SquareAppIcon/index.jsx
doubleface: > we don't pass any app props. So we're calling <AppIcon with app={undefined} no? And AppIcon need the slug in app props, right?

In this case app is undefined but appIconProps has an app props. Then AppIcon finally gets an app prop.

doubleface: > I'm wondering if we should not pass the AppIcon props directly to the SquareAppIcon without adding this appIconProps since there only variant, IconContent that is specific to SquareAppIcon. The result will be

That is what we got in the first place. But you made this [comment](https://github.com/cozy/cozy-ui/pull/1971/files#r749066601). Which was right I think, We will need the priority prop for the home application.

doubleface: Ok, non I understand. SquareAppIcon will get it's own props (variant, name, IconComponent) and pass the remaining props to AppIcon.

react/SquareAppIcon/index.jsx
```diff
@@ -61,9 +61,16 @@ const useStyles = makeStyles(theme => ({
...
 }))
 
-export const SquareAppIcon = ({ app, type, name, variant, IconContent }) => {
+export const SquareAppIcon = ({
+  name,
+  variant,
+  IconContent,
+  ...appIconProps
+}) => {
+  // "app" and "type" properties are deprecated in favor of appIconProps
```
Crash--, line 70: If you want to deprecate props, you should / can use https://github.com/cozy/cozy-ui/blob/67485ca3a6db28dd4ba64d2c91188c21b00034ce/react/helpers/migrateProps.jsx

Crash-- (approved):
I approve, but if you can use migrateProps to deprecate old props it can be nice 👍

react/SquareAppIcon/index.jsx
doubleface: Actually, it's my mistake, these props are not deprecated anymore. I remove the comment.

PR #1984, opened 2021-12-01
Expose ImageLoader comp from Viewer

JF-Cozy:
Pour un besoin MesPapiers, ce composant est nécessaire en tant que tel

BC : j'ai quand même déclencher un BC pour indiquer que ImageLoader n'est plus accessible depuis le dossier `/Viewer`


Merkur39:
C'est tout bon pour moi 👍

JF-Cozy:
il faut une main innocente pour valider  @Ldoppea :)

react/ImageLoader/index.spec.jsx
```diff
@@ -116,4 +123,31 @@ describe('ImageLoader', () => {
...
+    const { root } = setup({ file })
+    const { getByTestId } = root
+    getMock.mockResolvedValue({
+      data: {
+        links: {
+          icon: 'http://urlnotvalidneither'
+        }
+      }
+    })
+    checkImageSource.mockRejectedValueOnce('KO')
```
Ldoppea, line 147: What is the goal of this 2nd mock?

Ldoppea (commented):
If the goal of `ImageLoader` is to load an image from a `file`, then maybe we should name it `FileImageLoader`? (or similar)

This name would be more descriptive on what it does. `ImageLoader` is too generic in my opinion.

As we are making a BreakingChange I suppose it is a good time to improve the name?

Everything else looks good to me.

react/ImageLoader/index.spec.jsx
Merkur39: No goal, just a mistake, it's fixed 👍

Ldoppea (approved):

PR #1988, opened 2021-12-07
fix: Replace jest function in documentation

Merkur39:
Fix of the documentation following this commit 621fe47


JF-Cozy (approved):

Crash--:
@Merkur39 we shouldn't have been tagged this commit as `fix`. We didn't need a new npm version for that, it's only used by documentation. `docs` would have been good :)

PR #1989, opened 2021-12-07
fix: Correct colors about Alerter buttons

flohhhh:
We have an order problem during the build concerning some css classes.
Classes are imported before others, which causes style errors.

In our case, some properties had to be overridden to keep the expected behavior in the app.

That being said, the Alerter component seems old and not in accordance with the specifications.

### DIFF
--- a/stylus/components/button.styl
+++ b/stylus/components/button.styl
@@ -188,47 +188,56 @@ $button-alert
 
 $button-alert--error
     @extend $button-alert
-    color:   palette['pomegranate']  // @stylint ignore
+    color: palette['pomegranate'] !important // @stylint ignore
+    background-color: var(--white) !important // @stylint ignore
+    border-color: var(--white) !important // @stylint ignore
 
     &:visited
-        color: palette['pomegranate']  // @stylint ignore
+        color: palette['pomegranate'] !important // @stylint ignore
 
     &:active
     &:hover
     &:focus
-        color  var(--monza)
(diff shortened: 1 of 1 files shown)


JF-Cozy (approved):
Peux-tu expliquer les pistes que tu as investiguées concernant le build stp ?

Ldoppea (commented):
Would it be possible to get the same behavior by forcing a higher CSS precedence instead of using `!important`?

Maybe adding an `alerter` class to the Alerter and then say : `.alerter $button-alert--error` would be enough?

(sorry if you already tried this strategy)

flohhhh:
I followed these leads regarding the build order.  [#1](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/200), [#2](https://github.com/webpack/webpack/issues/215) and [#3](https://github.com/webpack/webpack/issues/7094). There is a lot  of information in it.

I tried to gradually upgrade `mini-css-extract-plugin` but didn't get the expected result.

flohhhh:
Already tried @Ldoppea but it doesn't work unfortunately.

Crash--:
> We have an order problem during the build concerning some css classes.
Classes are imported before others, which causes style errors.

Do we have an issue to get this problem in mind? Having added `!important` in the stylus means that now we can't use anymore our utilities method with this component. 

I really don't think this fix should be landed permanently in UI and we should have clear actions on what to do next to fix this issue.

PR #2001, opened 2021-12-23
fix: BottomSheet no longer opens fully

Merkur39:
Moved the contact async fetch in the Qualification Panel to the FooterContent parent, this prevents `mui-bottom-sheet` from miscalculating one of its style variables.

Demo: https://merkur39.github.io/cozy-ui/react/#!/Viewer/1

### DIFF
--- a/react/Viewer/Footer/BottomSheetContent.jsx
+++ b/react/Viewer/Footer/BottomSheetContent.jsx
@@ -20,34 +20,41 @@ const useStyles = makeStyles(theme => ({
   }
 }))
 
-const BottomSheetContent = forwardRef(({ file, disableSharing }, ref) => {
-  const panelBlocks = getPanelBlocks({ panelBlocksSpecs, file })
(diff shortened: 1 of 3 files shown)


Merkur39:
[🐛 Régression Viewer](https://trello.com/c/YMhlRg3Z/820-%F0%9F%90%9B-r%C3%A9gression-viewer)

Ldoppea:
Seems like you are introducing a specific context (`contacts`) in a generic component (`viewer`). I don't know all the context, but this seems not to be ideal. Can you describe why this contact list is needed?

Ldoppea:
Also your demo link is targeting the prod demo, shouldn't it be targeting your fork?

Merkur39:
> Seems like you are introducing a specific context (`contacts`) in a generic component (`BottomSheetContent`). I don't know all the context, but this seems not to be ideal. Can you describe why this contact list is needed?

The Qualification block can contain the ID of the associated contact(s)
Fetch these contacts allows to display their name

JF-Cozy (approved):

PR #2014, opened 2022-01-14
feat: Replace data-test-id by data-testid

JF-Cozy:
BREAKING CHANGE: If you were using `data-test-id` prop to test `NestedSelect` or `UploadQueue` components, you should now use `data-testid`. Besides if something fail on `CozyDialog` components, take a look at our helpers in `react/CozyDialogs` to know how to test these components.

### DIFF
--- a/react/CozyDialogs/ConfirmDialog.jsx
+++ b/react/CozyDialogs/ConfirmDialog.jsx
@@ -25,18 +25,18 @@ const ConfirmDialog = props => {
       {!fullScreen && onClose && (
         <DialogCloseButton
           onClick={onClose}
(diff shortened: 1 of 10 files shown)


Crash-- (approved):
> If you were using data-test-id to test the CozyDialogs

Not really true, basically we provide helpers from `CozyDialogs/testing.js` in order to help third party to test CozyDialog. Since we've updated these helpers, it should not break something. 

We can add in the breaking change, that if something fail on CozyDialog because of that commit, third party should take a look at our helpers to know how to test CozyDialog. 


For others components 👍

JF-Cozy:
Exact le data-test-id n'était pas sur MUIDialog

PR #2015, opened 2022-01-14
feat: Open apps using their native mobile version when available

Ldoppea:
AppLinker can now send app data to the native host in order to open the
native app when available instead of its webview version

---

The new `app` param is given by `cozy-home` and corresponds to the app's `manifest.webapp` content. This manifest can contain a `mobile` field (this will be documented on https://docs.cozy.io/en/cozy-stack/apps/#the-manifest soon)

---

Related PRs: 
- https://github.com/cozy/cozy-react-native/pull/65
- https://github.com/cozy/cozy-libs/pull/1422

---

Todo:

- [x] Upgrade cozy-intent after https://github.com/cozy/cozy-libs/pull/1422


Ldoppea:
I'm not sure if I should edit the `README.md` example to insert à dummy `app` param. This would make the doc more complete, but the described usage would not reflect a real use case. What do you think?

JF-Cozy:
> I'm not sure if I should edit the `README.md` example to insert à dummy `app` param. This would make the doc more complete, but the described usage would not reflect a real use case. What do you think?

Effectivement. Par contre on devrait quand même documenter d'une manière ou d'une autre l'ajout, histoire d'éviter le côté magique ? Peut-être dans le descriptif général du composant ?

Crash--:
> I'm not sure if I should edit the README.md example to insert à dummy app param. This would make the doc more complete, but the described usage would not reflect a real use case.

Why this will not reflect a real use case?

Ldoppea:
> Why this will not reflect a real use case?

Because the real use case gives the `app` param from the app's manifest. In my example I would have to create an object like this:
```js
const app = {
  slug: 'passwords',
  mobile: {
    schema: 'cozypass://',
    id_playstore: 'io.cozy.pass',
    id_appstore: 'cozy-pass/id1502262449'
  }
}
```

Also the dev would not be able to test the result as the doc is not accessible from the react native app.

Ldoppea:
> Effectivement. Par contre on devrait quand même documenter d'une manière ou d'une autre l'ajout, histoire d'éviter le côté magique ? Peut-être dans le descriptif général du composant ?

Quelque chose comme ça? https://github.com/cozy/cozy-ui/pull/2015/commits/1184e69816f8f5871a6b14d5e810e75740f342ce

(le lien dans la doc n'est pas encore valide, il sera une fois cette PR mergée: https://github.com/cozy/cozy-stack/pull/3263 )

Crash--:
> Because the real use case gives the app param from the app's manifest. In my example I would have to create an object like this

To me the "slug prop" should be deprecated. It's always a bad idea to not pass down directly the full object. Since we can retrieve the slug from the app object, we should do it. 

Also, it'll be easier to document since you will have to create a fake app object ;)

Ldoppea:
> To me the "slug prop" should be deprecated. It's always a bad idea to not pass down directly the full object. Since we can retrieve the slug from the app object, we should do it.
>
> Also, it'll be easier to document since you will have to create a fake app object ;)

Ok I'll do that.

I just tried to use `migrateProps` with for this but it breaks all test as we cannot do `jest.spyOn(AppLinker, 'openNativeFromNative')` anymore.

Which seems to be logic as we don't export a class component anymore but a function component. I'll investigate this tomorrow.

react/AppLinker/index.jsx
```diff
@@ -68,7 +68,7 @@ export class AppLinker extends React.Component {
 
       if (context) {
         return {
-          onClick: () => context.call('openApp', href),
+          onClick: () => context.call('openApp', href, app),
```
trollepierre, line 71: nit: is it possible to add a test on this call, in order to make the code more robust?

Ldoppea: I tried, but with @acezard we don't know how to mock the `context` object (which is handled by the messaging service)

Ldoppea:
> To me the "slug prop" should be deprecated. It's always a bad idea to not pass down directly the full object. Since we can retrieve the slug from the app object, we should do it.
> 
> Also, it'll be easier to document since you will have to create a fake app object ;)

@Crash-- I deprecated the slug prop in https://github.com/cozy/cozy-ui/pull/2015/commits/a4bf62e7664559bc1c1b52e4b23d05e191650d5b

I did not use `migrateProp` as this would break all the tests that expect to work with a class component. To make it work we should rewrite the component as a functional one and the rewrite all tests so the spies/mocks can work. Instead I manually do a `console.warn` from the old implementation.

I added `index.deprecated.spec.jsx` that contains the same tests as in `index.spec.jsx` but with old `slug` prop, so we can verify no regression has been done until we completely delete the deprecated prop.

I added a documentation example with mobile data but we cannot test its behavior as we are not in a react native env.

❓ should I add a "deprecated" warning in the commit message? Is there a flag like `BREAKING CHANGE:` that we can use?

react/AppLinker/index.jsx
trollepierre: ok; I tried without success as well.

trollepierre (approved):

Ldoppea:
[E12.0 - Accéder à Pass depuis la Home de l'App Amirale (5 jh)](https://trello.com/c/duNksKWL/152-e120-acc%C3%A9der-%C3%A0-pass-depuis-la-home-de-lapp-amirale-5-jh)

PR #2027, opened 2022-01-25
feat: Add FolderMoveto icon

Merkur39:
Currently it is going will be used in Mes papiers, but it will be used probably in the connectors to change the destination folder.

Demo: https://merkur39.github.io/cozy-ui/react/#!/Icon/1


JF-Cozy (approved):
tu peux valider argos 👍

PR #2034, opened 2022-01-28
feat: Inject webviewService into BarContext

acezard:
Uses the new cozy-intent API in order to allow CozyBar to use a WebviewContext passed through props by the parent application


react/BarContextProvider/index.spec.jsx
```diff
@@ -58,35 +80,65 @@ describe('BarContextProvider', () => {
...
       </Provider>
     )
 
-    // Currently only the WebviewProvider should call this function in BarContext
-    // This is an easy way to test that the provider is working, albeit brittle
-    // A full test would need to mock cozy-intent, post-me and react-native
-    expect(isFlagshipApp).toHaveBeenCalled()
+    await waitFor(() => screen.getByRole('main'))
+
+    expect(screen.getByText(mockVoidWebviewService)).toBeInTheDocument()
```
trollepierre, line 120: Quand on teste des composants avec Testing Library, il me semble que 
:a:
```
      const { queryByTestId } = render(<MyComponent />)
      expect(queryByTestId('id')).toBeInTheDocument()
```
est plus précis que :
:b:
```
      render(<MyComponent />)
      expect(screen.queryByTestId('id')).toBeInTheDocument()
```

d'après :
>  In addition, if you just want to query document.body then you can use the screen export as demonstrated below (using screen is recommended).
d'après : https://testing-library.com/docs/queries/about#using-queries

J'en déduis que la version :a: va query uniquement dans MyComponent. Ce dont on abesoin dans la majorité des tests.


(from mattermost)

trollepierre (changes requested):

react/BarContextProvider/index.spec.jsx
```diff
@@ -1,21 +1,21 @@
...
-import { isFlagshipApp } from 'cozy-device-helper'
 
 import BarContextProvider from '.'
 import I18n, { useI18n, translate } from '../I18n'
 
-jest.mock('cozy-device-helper', () => ({
-  isFlagshipApp: jest.fn()
-}))
-
+window.cozy.isFlagshipApp = true
```
Crash--, line 16: it means that new we only test this component in this use case. We should test it when embedded but also when it's not.

react/BarContextProvider/index.jsx
```diff
@@ -6,16 +6,29 @@ import { CozyProvider } from 'cozy-client'
...
-        <I18nContext.Provider
-          value={{ f: props.f, t: props.t, lang: props.lang }}
-        >
+    <Provider store={store}>
+      <CozyProvider client={client}>
+        <I18nContext.Provider value={{ f, t, lang }}>
           <BreakpointsProvider>
-            <WebviewIntentProvider>{props.children}</WebviewIntentProvider>
+            {webviewService ? (
+              <WebviewIntentProvider webviewService={webviewService}>
```
Crash--, line 26: nit: I don't like having special condition for only one provider. Can't we just manage that inside the provider itself? OK we have a gain of not having a new wrapper, but does it come with a cost?

acezard: Here I specifically do not want to create a WebviewIntentProvider because it itself creates its own context value if no props are given. We don't want that to ever happen because it would mean a cozy-bar could have a webview context while the parent app doesn't (or didn't pass the context as props)

react/BarContextProvider/index.spec.jsx
acezard: updated

react/BarContextProvider/index.spec.jsx
acezard: updated

react/BarContextProvider/index.jsx
```diff
@@ -6,16 +6,29 @@ import { CozyProvider } from 'cozy-client'
...
+const BarContextProvider = ({
+  children,
+  store,
+  client,
+  f,
+  t,
+  lang,
+  webviewService
+}) => {
+  if (!children) return null
```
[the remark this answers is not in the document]
acezard: updated

trollepierre (approved):

PR #2039, opened 2022-02-02
Upgrade mui to 4.12.3

JF-Cozy:
afin de supprimer les warnings du fait qu'une app peut avoir la version 4.12.3 et cozy-ui la version 4.11

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "@babel/cli": "7.11.5",
     "@babel/core": "7.11.5",
     "@cozy/codemods": "^1.9.0",
-    "@material-ui/core": "4.11.4",
+    "@material-ui/core": "4.12.3",
     "@semantic-release/changelog": "5.0.1",
     "@semantic-release/git": "7.0.18",
     "@semantic-release/npm": "5.3.4",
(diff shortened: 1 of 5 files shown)


Crash-- (changes requested):
Be careful, it's a breaking change. If an application has cozy-ui 4.11.X has dep and not 4.12.X, then `createTheme` doesn't exist. You force an application to upgrade by making this move.  

Also, since this `createTheme` has been added in 4.12.X, peerDep should reflect that https://github.com/cozy/cozy-ui/blob/7210288bbccc290f528d21f8999cb772fb666114/package.json#L170

JF-Cozy:
- 🤔 on utilise la `4` de mui dans les app, et non une version très spécifique comme `4.Y.Z`... mais tu veux dire que dans l'absolu on ne peut pas prétendre connaître la version de mui l'app utilise, et qu'il y a un BC dans le cas que tu listes ?

- on a déjà `  "peerDependencies": { "@material-ui/core": "4",`

Crash--:
`createTheme` has been added in M-ui v4.12. 

If an application use an older version, we'll have an error. I just take a look at two of our apps and... https://github.com/cozy/cozy-drive/blob/master/yarn.lock#L1143 & https://github.com/cozy/cozy-banks/blob/master/yarn.lock#L1454 

So yes, it's a breaking change if an app need to change something when updating a lib. 

Also, having peerDep saying 4 is not enough. With this change, peerDep should be :`>= 4.12`

PR #2075, opened 2022-03-23
feat: Implement openApp animation from cozy-ui

acezard:
When opening an application from the home screen, we want to send with cozy-intent the position and dimensions of the icon that was clicked on

### DIFF
--- a/react/AppLinker/index.jsx
+++ b/react/AppLinker/index.jsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+
 import {
   checkApp,
   startApp,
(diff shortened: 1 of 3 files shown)


trollepierre (approved):

react/AppLinker/index.jsx
```diff
@@ -79,15 +88,21 @@ export class AppLinker extends React.Component {
...
+          imgRef &&
+          JSON.stringify({
+            ...imgRef.getBoundingClientRect().toJSON()
+          })
+
         return {
           onClick: event => {
             event.preventDefault()
-            context.call('openApp', href, app)
+            context.call('openApp', href, app, imgPayload)
```
trollepierre, line 105: nit: let's test this PR after the rush

acezard: for sure

PR #2087, opened 2022-04-04
fix: Implement flagship top/bottom spacing

acezard:
- [x] MuiDialog
- [ ] Viewer
- [ ] BottomSheet


react/BottomSheet/BottomSheet.jsx
```diff
@@ -96,10 +96,19 @@ const BottomSheet = ({ toolbarProps, settings, children }) => {
...
+        .getPropertyValue('--flagship-bottom-height')
+        .replace('px', '')
+    )
+
     const minHeight =
       headerRef.current.offsetHeight +
       actionButtonsHeight +
-      actionButtonsBottomMargin
+      actionButtonsBottomMargin +
+      flagshipBottomOffset
```
trollepierre, line 111: don't we test that?

don't we add inside the doc BottomSheet/README.md ?

acezard: I think it's a bit too inconsequential to be tested. It will always default to 0 if the var doesn't exist. This type of thing should be handled by TypeScript in my opinion, not unit test

trollepierre: Testing is also a way of preventing regression, and documenting why this code is added. I think it is worthy. 

But, you choose what you want

trollepierre (approved):

PR #2093, opened 2022-04-13
fix(deps):  Define puppeteer as optional peer dependencies + piwik-react-router as dependencies

trollepierre:
Following our guideline, inside any Cozy library,
we expect peerDependency to be either react, react-dom
or other Cozy-library.

As piwik-react-router is not following this case, it should
come as a dependency of Cozy-UI

__

Few projects use regression testing of Cozy-UI: 
https://github.com/cozy/cozy-ui/blob/master/docs/README.md#ui-regression-testing= 

That is why we can be using [peerDependenciesMeta]
(https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta) 
that will consider this dependency as optional.

### DIFF
--- a/package.json
+++ b/package.json
@@ -181,6 +181,11 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
   },
+  "peerDependenciesMeta ": {
+    "puppeteer": {
+      "optional": true
+    }
+  },
   "eslintConfig": {
     "extends": [
       "eslint-config-cozy-app"
--- a/package.json
+++ b/package.json
@@ -121,7 +121,6 @@
     "mini-css-extract-plugin": "0.6.0",
     "nodemon": "1.19.4",
     "npm-run-all": "4.1.5",
-    "piwik-react-router": "0.12.1",
     "postcss-cli": "6.1.3",
     "postcss-loader": "2.1.6",
     "pretty": "2.0.0",
@@ -162,6 +161,7 @@
     "mui-bottom-sheet": "https://github.com/cozy/mui-bottom-sheet.git#v1.0.6",
     "node-polyglot": "^2.2.2",
     "normalize.css": "^8.0.0",
+    "piwik-react-router": "0.12.1",
     "react-markdown": "^4.0.8",
     "react-pdf": "^4.0.5",
     "react-popper": "^2.2.3",
@@ -177,7 +177,6 @@
     "cozy-harvest-lib": "^6.7.3",
     "cozy-intent": ">=1.3.0",
     "cozy-sharing": "^3.10.0",
-    "piwik-react-router": "0.12.1",
     "puppeteer": "^1.20.0",
     "react": "^16.8.6",
     "react-dom": "^16.8.6"


JF-Cozy:
je me sens pas legit pour répondre, faudrait que je creuse plus le sujet ou faudrait une description de PR plus détaillée. Pinguer quelqu'un d'autre en attendant si besoin :)

package.json
```diff
@@ -182,6 +182,14 @@
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
   },
+  "peerDependenciesMeta ": {
+    "piwik-react-router": {
+      "optional": true
+    },
```
Crash--, line 188: you can't make piwik optional. The package is used inside UI itself: https://github.com/cozy/cozy-ui/blob/3826219cc24ea4182a8ea57f7d8cf14ff9ffb691/react/helpers/tracker.jsx#L45

trollepierre: piwik will stay a peer Dependency.

Quoting the doc: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta
> Marking a peer dependency as optional ensures npm will not emit a warning if the soy-milk package is not installed on the host. This allows you to integrate and interact with a variety of host packages without requiring all of them to be installed.

trollepierre: As it's not mandatory to be using tracker.jsx inside app, I think we should not be warning during install of apps that don't have piwik-react-router in dependencies.

Crash--: I think this lib is mandatory, since all our apps are using getTracker(). I don't know why it was put in peerDep at first, maybe because of version's conflict but this doesn't reflect our internal guidelines, right?

trollepierre: https://github.com/cozy/cozy-ui/commit/e3bd229a9b1b8a68aa760c078adff80ab61dab3d

@JF-Cozy maybe?

trollepierre: First adding was strange : https://github.com/cozy/cozy-ui/commit/77274a6631e8c97d15f785145dbdb0420545cad7

PeerDependencies should not impact yarn.lock, in my understanding.

trollepierre: Let's follow our guideline:

piwik-react-router should be part of our "dependencies"

Crash--: > PeerDependencies should not impact yarn.lock, in my understanding.

Since the lockfile is there to fix versions, it's legit to have the peerDep listed in it.

trollepierre: > Since the lockfile is there to fix versions, it's legit to have the peerDep listed in it.

In order to add version in the lockfile, the package should be set in the dependencies or devDependencies array.

That is why @JF-Cozy fix the warning (because time and time, the presence inside the yarnlock was removed, I guess), by adding the piwik-react-router as devDependency.

Crash--: >by adding the piwik-react-router as devDependency.

Since you need the dep to dev / test the feature, it's legit to have it in devDep. 

ok for the lock file

trollepierre: new version: 
- Define puppeteer as optional peer dependencies 
- piwik-react-router as dependencies

trollepierre:
@Crash-- ?

Crash-- (approved):

PR #2097, opened 2022-04-14
feat: Add makeStyles helper from Mui

JF-Cozy:
to be able to use this helper from the Mui version of cozy-ui


trollepierre (approved):
looks ok for me. 

I just wonder if Cozy-UI should be using MUI or Cozy-UI.

I think Cozy-UI's makeStyles usage would be better

PR #2100, opened 2022-04-19
test: Use app object instead of slug

Crash--:
Let's remove this warning. AppLinker now uses `app` props instead of `slug`


JF-Cozy (approved):
😘

react/ContactsListModal/AddContactButton.jsx
```diff
@@ -40,7 +40,9 @@ const DumbAddContactButton = props => {
```
trollepierre, line 40: should this commit be a "chore"?

PR #2109, opened 2022-04-22
chore: Remove prepush commands

JF-Cozy:
to give better DX


trollepierre (approved):
Apparently, there is still husky running, but it's not slow. We can keep it.

Thanks LGTM! 💪

PR #2116, opened 2022-04-27
fix: Upgrade cozy-device-helper in peerDeps

acezard:
BREAKING CHANGE: cozy-ui needs this updated version
The Flagship App API changed in cozy-device-helper.
Some components will break if the parent app does not have
cozy-device-helper at ^1.18.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,7 +176,7 @@
   "peerDependencies": {
     "@material-ui/core": ">=4.12",
     "cozy-client": ">=28.1.0",
-    "cozy-device-helper": "^1.16.0",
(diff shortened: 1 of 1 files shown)


trollepierre (approved):
Great!

I would recommend to add inside the commit message, which component could be impacted.
And the expected error.


Example:

> Expected error: "error message blabla..."
> Remediation: `yarn upgrade cozy-device-helper@^1.8.0`

acezard:
> yarn upgrade cozy-device-helper@^1.8.0

updated

PR #2134, opened 2022-05-09
Fix/sort issue

Crash--:


react/AppSections/components/AppsSection.jsx
```diff
@@ -18,23 +25,38 @@ export const AppsSection = ({
...
-              namePrefix={getTranslatedManifestProperty(app, 'name_prefix', t)}
-              name={getTranslatedManifestProperty(app, 'name', t)}
-              onClick={() => onAppClick(app.slug)}
-              key={app.slug}
-              showDeveloper={!isMobile}
-              displaySpecificMaintenanceStyle={displaySpecificMaintenanceStyle}
-              IconComponent={IconComponent}
-            />
-          ))}
+          {sortBy(apps, makeNameGetter(t)).map(app => {
```
trollepierre, line 39: why not moving the sortBy inside useMemo? that operation (sorting) costs and chould benefit of memo, here.

trollepierre (approved):

react/AppSections/components/AppsSection.jsx
Crash--: thanks. Done

PR #2136, opened 2022-05-10
Set default size of IconButton to `large`

JF-Cozy:
BREAKING CHANGE: `IconButton`'s default props has been changed from `medium` to `large`. To keep the old behavior, you can use a codemods `transform-iconButton.js` to automatically handle this breaking change by setting `IconButton` size to `medium`. [See the codemods documentation](https://github.com/cozy/cozy-libs/tree/master/packages/cozy-codemods). Using linter js auto-correction can be a good idea after that. Here a common example (don't forget to change `src` value): `jscodeshift -t $(yarn global dir)/node_modules/@cozy/codemods/src/transforms/transform-iconButton.js src --parser babel --extensions js,jsx && yarn lint:js --fix`

fix https://github.com/cozy/cozy-ui/issues/2107

pas de démo, il n'y a aucun modif à montrer, c'est le principe du codemods


Ldoppea (approved):
Looks good to me.

Maybe the BC is missing a bit of details on what has broken. The fact that previous behavior was defaulting to `medium` was not an evidence to me (but it is after reading the PR's code). Also keep in mind that the PR will be put in the CHANGELOG without the PR's title.

Maybe adding something like this in the commit message? 

> BREAKING CHANGE: iconButton's default props has been changed from `medium` to `large`. To keep the old behavior, you can use a codemods `transform-iconButton.js` and automatically handle this breaking change by setting `IconButton` size to `medium`.

What do you think about this?

JF-Cozy:
@Ldoppea le titre de la PR non, mais tous les titres des commits oui. Et il y en a un en `feat` qui porte le même nom. Mais 👍 pour rajouter de l'explication

trollepierre (approved):

PR #2139, opened 2022-05-12
fix(npmignore): Reduce npm size of the package by ignoring /react files

trollepierre:
BREAKING CHANGE: any import of @cozy-ui/react will fail
replace @cozy-ui/react by @cozy-ui/transpiled/react

While analysing @cozy-ui folder in node_modules I realized, a lot of useless file were present.
The .npmignore file will prevent those files to be present in the npm package.
More info: https://docs.npmjs.com/cli/v8/using-npm/developers#keeping-files-out-of-your-package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,12 +1,24 @@
 test
 build
-
+react
+scripts
+.bundlemonrc
+.nvmrc
 .editorconfig
 .eslintignore
 .eslintrc.json
+.nvmrc
 .stylintrc
 .travis.yml
+babel.config.js
+CODEOWNERS
+index.js
+preprocess.js
+release.config.js
 renovate.json
+rsgscreenshots.json
+svgo.config.js
+tsconfig.json
 yarn.lock
 
 .changelog


JF-Cozy:
il faut créer un codemods pour gérer le BC. L'idée serait de remplacer les `import XXX from 'cozy-ui/react'` par `import XXX from 'cozy-ui/transpiled/react'` c'est ça ?

codemods est ici https://github.com/cozy/cozy-libs/tree/master/packages/cozy-codemods, les sources ici https://github.com/cozy/cozy-libs/tree/master/packages/cozy-codemods/src/transforms

Crash--:
It will only change the package size on npm, not in our applications. By doing only an npmignore we don't have a gain while building or else. We've only a gain while publishing. 

I'm not: 
1/ Very confident about those changes (BC, package.json not changed even though you remove the index.js from the build , the main entry point is still index.js) 
2/ It changes nothing about our lib size. So the description is not really true

trollepierre:
Running `npm pack` on the branch:
Capture d’écran 2022-05-13 à 10 46 55
Running `npm pack` on master:
Capture d’écran 2022-05-13 à 10 47 44

trollepierre:
> il faut créer un codemods pour gérer le BC. L'idée serait de remplacer les `import XXX from 'cozy-ui/react'` par `import XXX from 'cozy-ui/transpiled/react'` c'est ça ?
> 
> codemods est ici https://github.com/cozy/cozy-libs/tree/master/packages/cozy-codemods, les sources ici https://github.com/cozy/cozy-libs/tree/master/packages/cozy-codemods/src/transforms

I was skeptical about that, but apparently a lot of imports are wrong:
https://cs.github.com/?scopeName=All+repos&scope=&q=org%3Acozy+%22from+%27cozy-ui%2Freact%22

trollepierre:
> It will only change the package size on npm, not in our applications. By doing only an npmignore we don't have a gain while building or else. We've only a gain while publishing.
> 
> I'm not: 1/ Very confident about those changes (BC, package.json not changed even though you remove the index.js from the build , the main entry point is still index.js) 2/ It changes nothing about our lib size. So the description is not really true

1/ `index.js` contains only `export default {}`. Is it worthy to keep it? Why?

1/ about BC, yes you are right. I saw a lot of wrong import inside our apps.

2/ description changed.

By the way, I added the difference after npm pack. Huge difference of size.

Crash--:
We should have in all our apps still using cozy-ui/react a webpack alias to resolve cozy-ui/react to cozy-ui/transpiled/react. 

> 1/ index.js contains only export default {}. Is it worthy to keep it? Why?

I'm not saying that we should keep it, I'm saying that right now, in your branch, the main entry point doesn't exist. 

> By the way, I added the difference after npm pack. Huge difference of size.

This is what I say "It will only change the package size on npm, not in our applications." Having an "huge difference of size" of npm, what is the gain for our DX? Does this BC worth the benefits? 🤔

PR #2145, opened 2022-05-22
fix: Improve App Amirale UI events

acezard:
# Description

Due to some cozy-apps structure, the Amirale App UI update with setFlagshipUI wasn't working as expected, and the native application UI looked broken. To tackle this, it was necessary to compute desired values in a better way. While fixing the issues, some small features and refactoring were added for a better developer experience as this feature was starting to get unreadable.

Addresses this [Trello ticket](https://trello.com/c/nFvqeznK/402-corriger-les-status-bar-nav-bar-sur-lensemble-des-%C3%A9crans)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Refactoring (non-breaking change which adds readability)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# How Has This Been Tested?

This was tested on an Android emulator only, not on iOS, with _rlinked_ `cozy-ui` running on local cozy-apps.

- [x] `Modal` component should mount with black on white top and bottom
- [x] `Modal` component should unmount with a fitting theme for `cozy-bar` and presence of bottom `Sidebar`
- [x] `Dialog` component should mount with no background colours if stacked on a `Dialog`
- [x] `Dialog` component should unmount with no background colours if stacked on a `Dialog`
- [x] `ActionMenu` component should work as before

**Test Configuration**:
* WSL2 (Ubuntu-20.04) on Windows 10 Pro
* Microsoft Edge (Official build)

# Checklist:

- [x] I have performed a self-review of my code
- [x] I have commented or added typings to my code, particularly in hard-to-understand areas
- [x] I have improved existing tests to take into account my feature
- [ ] I have added tests that prove my fix is effective or that my feature works


react/Dialog/DialogEffects.ts
```diff
@@ -0,0 +1,102 @@
...
+                theme.palette.background[sidebar ? 'default' : 'paper']
+            }),
+        bottomTheme: 'light',
+        bottomOverlay: 'rgba(0, 0, 0, 0.5)',
+        topOverlay: 'rgba(0, 0, 0, 0.5)',
+        ...(hasAnotherDialog || !cozyBar
+          ? {}
+          : {
+              topBackground:
+                cozyBar &&
```
Ldoppea, line 36: Do we really need to check `cozyBar` here? It should be protected by the `!cozyBar` condition.

react/Modal/ModalEffects.ts
```diff
@@ -0,0 +1,47 @@
...
+      topTheme: 'dark'
+    },
+    {
+      bottomBackground: theme.palette.background[sidebar ? 'default' : 'paper'],
+      bottomTheme: getFlagshipMetadata().immersive ? 'light' : 'dark',
+      bottomOverlay: 'transparent',
+      topOverlay: 'transparent',
+      topBackground: getTopBackground(theme, cozyBar),
+      topTheme: getTopTheme(cozyBar)
+    },
```
Ldoppea, line 36: This code has similarities with the one on `DialogEffects.ts`. Maybe we can homogenize them later?

Ldoppea (approved):

react/Dialog/DialogEffects.ts
acezard: You're right, this guard is redundant, removed it.

react/Modal/ModalEffects.ts
acezard: Yes I wanted to do that at first. But <Modal/> is a depreciated component, and there are too many small differences with <Dialog/> in my opinion. I'm not comfortable in mutualising this, I'm fearing side-effects. By the way, <Modal/> is only utilised in a single cozy-application (`cozy-store`). Once we remove <Modal/> from `cozy-store`, I will be comfortable with deleting this hook entirely.

JF-Cozy (approved):

PR #2180, opened 2022-06-30
fix(ListSubheader): Only 1rem of text indent on mobile

JF-Cozy:
demo : https://jf-cozy.github.io/cozy-ui/react/#!/List


Merkur39 (approved):
👌

PR #2190, opened 2022-07-18
fix: Shortcut with IconContent

Crash--:
We lose this feature during the creation on the
SquareAppIcon: A shortcut can have an attached
binary and in that case, we want to display
the IconContent instead of the Letter.

See argos for the visual diff

### DIFF
--- a/react/SquareAppIcon/index.jsx
+++ b/react/SquareAppIcon/index.jsx
@@ -109,12 +109,12 @@ export const SquareAppIcon = ({
           size="large"
           overlap="rectangular"
           style={
-            variant === 'shortcut'
+            variant === 'shortcut' && !IconContent
               ? { backgroundColor: nameToColor(name) }
               : null
           }
         >
-          {variant === 'shortcut' ? (
+          {variant === 'shortcut' && !IconContent ? (
             <Typography className={classes.letter} variant="h2" align="center">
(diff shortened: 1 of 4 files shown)


JF-Cozy (approved):

react/SquareAppIcon/index.jsx
```diff
@@ -109,12 +109,12 @@ export const SquareAppIcon = ({
...
           overlap="rectangular"
           style={
-            variant === 'shortcut'
+            variant === 'shortcut' && !IconContent
               ? { backgroundColor: nameToColor(name) }
               : null
           }
         >
-          {variant === 'shortcut' ? (
+          {variant === 'shortcut' && !IconContent ? (
```
trollepierre, line 117: nit: let's add a little test to prevent this regression to happen again.

I am preparing the commit

trollepierre: done

Crash--: To me, this is the argos' job to check we don't have any regression. 

Here, even with this test, we would have the regression, why? Because we created a brand new component with a new behavior. But we forgot to add one case. 

So neither UT or Argos will save us in that case. The only thing that can save us is to do TDD on that kind of development

PR #2199, opened 2022-07-26
fix(BottomSheet): Backdrop click behavior

JF-Cozy:
Now we have to click on the backdrop to close it, instead of clicking outside the BottomSheet. This is all the difference. Now we can have a modal above the BottomSheet, and clicking in it doesn't close the BottomSheet anymore.


Merkur39 (approved):
👍

PR #2206, opened 2022-08-05
fix: Prevent `AppLinker` to call openApp when target link has same slug

Ldoppea:
When in Flagship mobile app, `AppLinker` is responsible to call
`openApp` for the specified URL and the specified slug, so the mobile
app can handle navigation to the target app

But if the target slug is the same as the current displayed app, it is
unnecessary to call `openApp` as the mobile app would have to handle
navigation to the same route

Instead we want to use the default `web` behaviour in this scenario.
Which means that the web app would redirect to the target URL by using
classic `window.location` redirection

We only check for the URL's slug. The domain here should not be a
problem because if it is the same, then we stay on the app, and if it
is different, then the mobile app would be responsible to detect it and
to open a dedicated view for the external link

Note: this feature is enabled ONLY IF the `AppLinker` component is
inside a `CozyProvider`. If it is not the case, then `AppLinker` would
always call `openApp` like before

___

Related PRs:

- https://github.com/cozy/cozy-notes/pull/318
- https://github.com/cozy/cozy-client/pull/1217

### DIFF
--- a/react/AppLinker/__snapshots__/index.deprecated.spec.jsx.snap
+++ b/react/AppLinker/__snapshots__/index.deprecated.spec.jsx.snap
@@ -2,23 +2,24 @@
 
 exports[`app icon should not crash if no href 1`] = `
 <div>
-  <a
-    onClick={[Function]}
-  >
-    Open 
-    Cozy Drive
-  </a>
+  <div>
+    <a>
+      Open 
+      Cozy Drive
+    </a>
+  </div>
 </div>
 `;
 
 exports[`app icon should render correctly 1`] = `
 <div>
-  <a
-    href="https://fake.link"
-    onClick={null}
-  >
-    Open 
-    Cozy Drive
-  </a>
+  <div>
+    <a
+      href="https://fake.link"
+    >
+      Open 
+      Cozy Drive
+    </a>
+  </div>
 </div>
 `;
--- a/react/AppLinker/index.deprecated.spec.jsx
+++ b/react/AppLinker/index.deprecated.spec.jsx
@@ -7,25 +7,26 @@
  */
 
 import React from 'react'
-import { shallow } from 'enzyme'
+import { render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+
 import {
   isMobileApp,
   isMobile,
   openDeeplinkOrRedirect,
   startApp,
-  isAndroid
+  isAndroid,
+  checkApp
 } from 'cozy-device-helper'
 
 import AppLinker from './index'
 import { generateUniversalLink } from './native'
 jest.useFakeTimers()
 
-const tMock = x => x
-
-class DeprecatedAppItem extends React.Component {
-  render() {
-    const { app, onAppSwitch } = this.props
-    return (
+const setup = ({ app, onAppSwitch }) => {
+  return {
+    user: userEvent.setup({ delay: null }),
+    ...render(
       <AppLinker
         onAppSwitch={onAppSwitch}
         slug={app.slug}
@@ -42,6 +43,7 @@ class DeprecatedAppItem extends React.Component {
     )
   }
 }
+
 jest.mock('./native', () => ({
   ...jest.requireActual('./native'),
   generateUniversalLink: jest.fn()
@@ -53,7 +55,8 @@ jest.mock('cozy-device-helper', () => ({
   isMobile: jest.fn(),
   openDeeplinkOrRedirect: jest.fn(),
   startApp: jest.fn().mockResolvedValue(),
-  isAndroid: jest.fn()
+  isAndroid: jest.fn(),
+  checkApp: jest.fn()
 }))
 
 const app = {
@@ -62,7 +65,7 @@ const app = {
 }
 
 describe('app icon', () => {
-  let spyConsoleError, spyConsoleWarn, openNativeFromNativeSpy, appSwitchMock
+  let spyConsoleError, spyConsoleWarn, appSwitchMock
 
   beforeEach(() => {
     isMobileApp.mockReturnValue(false)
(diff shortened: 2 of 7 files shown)


Crash-- (approved):

Ldoppea:
I just updated unit tests by migrating to react-testing-library.

Old enzyme implementation did not work anymore. Especially all `root.find('a')` related code. Instead of fixing it, migrating to react-testing-library allowed to homogenise test with team's latest practices and `container.querySelector('a')` is straightforward.

I also added `@testing-library/user-event` to devDependencies to replace `fireEvent.click()` by `userEvent.click()` which is better at handling async code and also it may be useful later to simulate user interactions.

Updated snapshots lack of `onClick` props in `<a>` elements. This is because react-testing-library seems to display only valid DOM attributes.

react/AppLinker/index.deprecated.spec.jsx
```diff
@@ -84,64 +86,68 @@ describe('app icon', () => {
...
+    const link = container.querySelector('a')
+    await user.click(link)
+
     expect(startApp).toHaveBeenCalledWith({
       appId: 'io.cozy.drive.mobile',
       name: 'Cozy Drive',
       uri: 'cozydrive://'
     })
     expect(appSwitchMock).toHaveBeenCalled()
+    console.log('FINITO')
```
Crash--, line 115: to remove

react/AppLinker/index.spec.jsx
```diff
@@ -73,20 +75,25 @@ describe('app icon', () => {
...
-  it('should work for native -> native', () => {
-    const root = shallow(
-      <AppItem t={tMock} app={app} onAppSwitch={appSwitchMock} />
-    ).dive()
-    root.find('a').simulate('click')
+  it('should work for web -> web', async () => {
+    isMobileApp.mockReturnValue(false)
+    const { container, user } = setup({ app, onAppSwitch: appSwitchMock })
+    const link = container.querySelector('a')
+    await user.click(link)
```
Crash--, line 86: Why do we need to await the click? What are the async stuff we're doing in that case?

react/AppLinker/index.spec.jsx
```diff
@@ -95,53 +102,53 @@ describe('app icon', () => {
...
       <AppLinker onAppSwitch={appSwitchMock} app={app}>
-        {({ onClick, href, name }) => (
-          <div>
-            <a href={href} onClick={onClick}>
-              Open {name}
-            </a>
-          </div>
-        )}
+        {({ onClick, href, name }) => {
+          console.log({ onClick, href, name })
```
Crash--, line 141: to remove

react/AppLinker/__snapshots__/index.deprecated.spec.jsx.snap
```diff
@@ -2,23 +2,24 @@
...
-    Cozy Drive
-  </a>
+  <div>
+    <a
+      href="https://fake.link"
+    >
+      Open 
+      Cozy Drive
+    </a>
+  </div>
```
Crash--, line 23: is is attented to have a new `<div>` node? 

Also, are you sure that `onClick` is not a valid DOM Event for an `a`?

react/AppLinker/index.spec.jsx
Ldoppea: In this specific example (`should work for web -> web`) there was no need to await for click result.

The original need was for `should work for native -> native` test. With this configuration, AppLinker calls `AppLinker.openNativeFromNative()` which then calls `startApp()` which is a promise.

I'm not sure why, but in that case, when calling `fireEvent.click(link)` then the check `expect(startApp).toHaveBeenCalledWith` failed.

My intuition was that it wouldn't fail because `startApp` should be called before the method returns (event if it does not wait its result). But in practice: it appears that it is not the case. Maybe this is a race condition?

However, `user.click()` bring the `act()` mechanism by default which seems to fix our problem. The implication is that `user.click()` is a promise, and so we have to await it.

For all other tests, I used `user.click` for homogenisation purpose.

react/AppLinker/index.deprecated.spec.jsx
Ldoppea: Fixed in https://github.com/cozy/cozy-ui/compare/afdf6b775cfac127955c238b95653c6fd14ca6ae..367be680b4fb6f8f7e96a3fe5507001e5e2d672e

Thanks 👍

react/AppLinker/__snapshots__/index.deprecated.spec.jsx.snap
Ldoppea: > is is attented to have a new `<div>` node?

This is because we now snapshot the `container` from testing-library's `render()`.

`containers` add the `<div>` around the component.

> Also, are you sure that onClick is not a valid DOM Event for an a?

I'm not sure about what defines if a prop is kept or not.

For exemple the following test : 
```js
    const { container } = render(
      <div>
        <a href='http://someHREF' onClick='onClick' someOtherProp='somePROP'>
          Open
        </a>
      </div>
    )
    expect(container).toMatchSnapshot()
```

Produces the following snapshot:
```
exports[`SOME TEST`] = `
<div>
  <div>
    <a
      href="http://someHREF"
      someotherprop="somePROP"
    >
      Open
    </a>
  </div>
</div>
`;
```

The result is the same with `onclick=` (all lowercase like in w3c spec)

Crash--: Capture d’écran 2022-08-08 à 16 30 00
Capture d’écran 2022-08-08 à 16 30 41
Capture d’écran 2022-08-08 à 16 30 44


So the DOM representation is not having the onclick attached. if it's specified later?

Crash-- (approved):

PR #2211, opened 2022-08-18
chore: specify node version that doesn't break examples.spec.jsx

vdnj:


package.json
```diff
@@ -16,6 +16,9 @@
     "url": "https://github.com/cozy/cozy-ui/issues"
   },
   "homepage": "https://github.com/cozy/cozy-ui",
+  "engines": {
+    "node": ">=16.16"
+  },
```
Crash--, line 21: I'm not a big fan of requiring node 16.16 for cozy-ui. Can you tell us the issue you had with older node version please?

Crash--: By settings engines node >= 16.16 you tell that ALL stuff that consume that cozy-ui should at least be on this version. So you can't install cozy-ui in your project if you still use node 12/14/16.14 even if there is no need of 16.16 for cozy-ui for this application. It a bit brutal. 

We only need >=16.16 for our dev process, right?

JF-Cozy:
@vdnj petit up pour cette PR ?

package.json
vdnj: @trollepierre do you remember why older version breaks examples.spec.js ? Is it a better way to solve it ?

JF-Cozy: si c'est pour du dev, seul la modif dans nvmrc suffit non ? Faudrait peut-être rajouter aussi dans travis 🤔

trollepierre: @JF-Cozy : https://github.com/cozy/cozy-ui/blob/master/.nvmrc the .nvmrc is updated.

trollepierre: > do you remember why older version breaks examples.spec.js ? Is there a better way to solve it ?

No. It was an async error, that we usually faced when Node version is not correct. 
I asked you to launch: `nvm use 16.16 && yarn test` and that solved the problem. 
But I think you said later that you still encounter the problem again, don't you?

vdnj: Yes

trollepierre: On this branch, or without this fix?

trollepierre: then, let's close this PR.

JF-Cozy:
je close donc

PR #2216, opened 2022-08-23
feat: Upgrade cozy-client to 33.0.0

Ldoppea:
`cozy-client` as been upgraded to to `33.0.0` to retrieve improved
useCapabilities from cozy/cozy-client#1227

BREAKING CHANGE: You need to update `cozy-client` to `>=33.0.0`.


package.json
```diff
@@ -87,7 +87,7 @@
     "babel-preset-cozy-app": "2.0.2",
     "browserslist-config-cozy": "0.4.0",
     "copyfiles": "2.4.1",
-    "cozy-client": "28.2.0",
+    "cozy-client": "33.0.0",
```
trollepierre, line 90: let's update cozy-stack-client as well 🙏

trollepierre: can we use ^ prefix for the cozy-* library?

trollepierre (approved):

package.json
Ldoppea: Fixed, I also had to add `@cozy/minilog`, `cozy-flags` and `cozy-logger` in devDependencies du to this breaking change https://github.com/cozy/cozy-client/blob/master/packages/cozy-stack-client/CHANGELOG.md#2800-2022-04-08

Not sur if I should add them as peerDependencies

PR #2218, opened 2022-08-24
feat: Let the user able to select content

Crash--:
An user reported us that he was not able
to select a text from a PDF (https://github.com/cozy/cozy-drive/issues/2673).

After a quick look, it appears that HammerJS
adds a `user-select:none` inline style in
order to be able to add its gestures things.

We can disable this option when creating the
Hammer object. This is what I've done here
for the desktop version.

I also removed the swipe gesture for desktop
because we can have "conflict" between text
selection and swipe. This is why I didn't
remove this behavior on mobile, since I don't
have the time to test it correctly on several
devices.

It also appears that we don't need the tap/
doubletap gesture anymore in order to display
the controls because the control is now
only displayed on hover. Else wehre, the opacity
is set to false.

https://crash--.github.io/cozy-ui/react/#!/Viewer

### DIFF
--- a/react/Viewer/ViewerControls.jsx
+++ b/react/Viewer/ViewerControls.jsx
@@ -47,11 +47,6 @@ class ViewerControls extends Component {
     }
   }
 
-  onTap = () => {
-    if (this.state.hidden) this.showControls()
-    else this.hideAfterDelay()
-  }
-
   hideAfterDelay = () => {
     clearTimeout(this.hideTimeout)
     this.hideTimeout = setTimeout(() => {
@@ -65,13 +60,18 @@ class ViewerControls extends Component {
   }
 
   initGestures = () => {
-    const gestures = new Hammer(this.wrapped)
-    gestures.on('swipe', this.onSwipe)
-    gestures.on('tap', this.onTap)
-    const tap = gestures.get('tap')
-    const doubleTap = gestures.get('doubletap')
-    doubleTap.recognizeWith(tap)
-    tap.requireFailure(doubleTap)
+    const gestures = new Hammer(
+      this.wrapped,
+      this.props.breakpoints.isDesktop
+        ? {
+            cssProps: {
+              userSelect: 'auto'
+            }
+          }
+        : {}
+    )
+    if (!this.props.breakpoints.isDesktop) gestures.on('swipe', this.onSwipe)
+
     return gestures
   }
 
@@ -116,7 +116,6 @@ class ViewerControls extends Component {
     } = this.props
     const { showToolbar, showClose, toolbarRef } = toolbarProps
     const { hidden } = this.state
-
     return (
       <div
         className={cx(styles['viewer-controls'], {


Crash--:
We can't fix it right now on mobile since we don't display PDF on mobile (only screenshot of the PDF). But we still have the issue on mobile with the markdown. All in good time

JF-Cozy:
@Crash-- ça plante sur le fichier après le fichier encrypté, à savoir l'image

Autre chose, j'ai un avertissement de sécu quand je vais sur ton lien... ?

Crash--:
Oh, bien vu pour l'erreur sur l'ImageViewer. J'étais pas allé jusque là. Je pense voir pourquoi, je regarderai 👍

Crash--:
@JF-Cozy fixed.

react/Viewer/ViewerControls.jsx
```diff
@@ -65,13 +60,18 @@ class ViewerControls extends Component {
...
 
   initGestures = () => {
-    const gestures = new Hammer(this.wrapped)
-    gestures.on('swipe', this.onSwipe)
-    gestures.on('tap', this.onTap)
-    const tap = gestures.get('tap')
-    const doubleTap = gestures.get('doubletap')
-    doubleTap.recognizeWith(tap)
-    tap.requireFailure(doubleTap)
+    const gestures = new Hammer(
```
JF-Cozy, line 63:
```
const { breakpoints: { isDesktop }} = this.props
const gestures = new Hammer(...
```
pour un peu plus de lisibilité...

JF-Cozy (approved):
on a supprimé onTap, qui utilisait les méthodes `this.showControls()` et `this.hideAfterDelay()`. Pas besoin de les supprimer également, elles sont utilisées ailleurs ?

Crash--:
> on a supprimé onTap, qui utilisait les méthodes this.showControls() et this.hideAfterDelay(). Pas besoin de les supprimer également, elles sont utilisées ailleurs ?

Oui showControls & hideControls sont utilisé dans le composant `<Toolbar>`. Et les deux appellent hideAfterDelay.

Ldoppea:
> It also appears that we don't need the tap/
doubletap gesture anymore in order to display
the controls because the control is now
only displayed on hover. 

There is no "hover" on mobile.

Is that included in your "We can't fix it right now on mobile since we don't display PDF on mobile" indication?

Crash--:
> There is no "hover" on mobile.
Is that included in your "We can't fix it right now on mobile since we don't display PDF on mobile" indication?

Nein. I'll reformulate the commit to clarify that we don't display the control on mobile, so yes there is no hover on mobile but it doesn't change anything. 

By saying "We can't fix it right now on mobile since we don't display PDF on mobile" I mean: When we display a "pdf document" on mobile, we don't really display the PDF. We display a thumbnail of the first page of the PDF. So there is no text selection available at all. Even without gestures. 

The only case remaining on mobile, is the case were we display a markdown. In that fact, we can't select the text.

react/Viewer/ViewerControls.jsx
Crash--: Will do it for the next PR 🙏

PR #2242, opened 2022-09-08
fix: isLoadingContacts on useReferencedContactName hook

JF-Cozy:
The Query being conditioned by `contactIds.length > 0`,
`isLoadingContacts` must also be.
Otherwise `isQueryLoading(contactsQueryResult)` is always `true`.


Merkur39 (approved):
👌

PR #2249, opened 2022-09-19
fix(Field): Use ref instead of inputRef when  using Input comp

JF-Cozy:


react/Field/index.jsx
```diff
@@ -21,59 +21,52 @@ const inputSpecificPropTypes = {
...
-class InputPassword extends React.Component {
-  state = {
-    visible: false
-  }
+const InputPassword = ({
+  hideLabel,
+  showLabel,
+  fullwidth,
+  secondaryComponent,
+  inputRef,
```
Merkur39, line 29: nit : Un petit forwardRef ? :P

Merkur39 (approved):
Ok, vu en call 👍

PR #2281, opened 2022-11-04
Add gap utility classes

PolariTOON:
We often need to put some regular spacing between elements. I learned there are the `u-stack-*` classes that basically allow that, but :

- in only one direction
- by overriding margins

At the end, it seems we don't really use these classes but rather margin utility classes on each element.

As we often are in flex contexts, i propose adding new gap utility classes that do not have these limiations and allow to add gutters between child *nodes* (not elements) of a flex element by declaring them only at a single location.
Gaps can also be used in grid contexts and multi-column layouts, and eventually will likely be compatible with subgrids, masonry layouts, table layouts, multi-row layouts...

Demo link: https://polaritoon.github.io/cozy-ui/styleguide/section-utilities.html#kssref-utilities-gaps

### DIFF
--- a/stylus/cozy-ui/build.styl
+++ b/stylus/cozy-ui/build.styl
@@ -1010,6 +1010,49 @@ Display an chip that represents complex identity
 */
 
 /*
+    Gaps
+
+    Classes to add gaps between children nodes, vertically, horizontally or in both directions based on default font-size (16px)<br>
+    [Can be used with breakpoints modifier suffixes (`-t`,`-s`,`-m`)](section-settings.html#kssref-settings-breakpoints)
+
+    .u-g-normal - Normal gaps in both directions
+    .u-g-0 - No gaps in both directions
+    .u-g-half - Gaps in both directions with half a line (8px)
+    .u-g-1 - Gaps in both directions with 1 line (16px)
+    .u-g-1-half - Gaps in both directions with 1 line an a half (24px)
+    .u-g-2 - Gaps in both directions with 2 lines (32px)
+    .u-g-2-half - Gaps in both directions with 2 lines an a half (40px)
+    .u-g-3 - Gaps in both directions with 3 lines (48px)
+    .u-gv-normal - Normal vertically gaps
+    .u-gv-0 - No vertically gaps
+    .u-gv-half - Vertically gaps with half a line (8px)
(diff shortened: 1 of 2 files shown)


JF-Cozy:
Partons sur les composants Stack ou Grid pour gérer ces cas

JF-Cozy:
a few moment later... ⏳ around 4 years 🤣

stylus/utilities/spaces.styl
```diff
@@ -21,6 +21,27 @@ directions = {
     h: horizontal
 }
 
+gapTypes = {
+    g: gap
+}
+
+gapSizes = {
+    'auto': normal,
```
JF-Cozy: 'normal': normal ?

stylus/cozy-ui/build.styl
```diff
@@ -1009,6 +1009,49 @@ Display an chip that represents complex identity
     Styleguide utilities.paddings
 */
 
+/*
+    Gaps
+
+    Classes to add gaps between children nodes, vertically, horizontally or in both directions based on default font-size (16px)<br>
+    [Can be used with breakpoints modifier suffixes (`-t`,`-s`,`-m`)](section-settings.html#kssref-settings-breakpoints)
+
+    .u-g-auto - Auto gaps in both directions
```
JF-Cozy: .u-g-normal

stylus/cozy-ui/build.styl
```diff
@@ -1009,6 +1009,49 @@ Display an chip that represents complex identity
...
+    .u-g-0 - No gaps in both directions
+    .u-g-half - Gaps in both directions with half a line (8px)
+    .u-g-1 - Gaps in both directions with 1 line (16px)
```
JF-Cozy, line 1021: don't know if u-g- is clearer than u-gap- . I assume with u-g we copy margin/padding approach 🤔

stylus/cozy-ui/build.styl
JF-Cozy, line 1: commit should be squashed, you know it ;)

JF-Cozy (approved):

stylus/cozy-ui/build.styl
zatteo: Yes it is the purpose. Actually everything is copy/pasted from margin/padding except we have only vertical/horizontal and auto is replaced by normal.

PR #2293, opened 2022-11-28
fix: Remove duplicate paths in icons

vdnj:
Some svg icons have duplicate paths. It causes bad opacity when we use them.
Removed duplicates from "folder", "gear", "share" and "trash" icons to fix it.


JF-Cozy (approved):
good job 👍

PR #2310, opened 2022-12-09
Update `date-fns` to 2.29.3

PolariTOON:
This is a major update of `date-fns` which brings breaking changes that will affect libs and apps consuming Cozy UI.

### DIFF
--- a/react/BarContextProvider/index.spec.jsx
+++ b/react/BarContextProvider/index.spec.jsx
@@ -22,7 +22,7 @@ const DumbHelloWorld = translate()(({ t, f, lang }) => (
   <div>
     {t('helloworld')}
     <br />
-    {f('2020-01-06', 'DDD MMM')}
+    {f(new Date('2020-01-06'), 'd MMM')}
     <br />
     {lang}
   </div>
--- a/react/I18n/format.jsx
+++ b/react/I18n/format.jsx
@@ -1,7 +1,7 @@
 import format from 'date-fns/format'
 import { DEFAULT_LANG } from '.'
-import formatDistanceToNow from 'date-fns/distance_in_words_to_now'
-import formatDistanceStrict from 'date-fns/distance_in_words_strict'
+import formatDistanceToNow from 'date-fns/formatDistanceToNow'
+import formatDistanceToNowStrict from 'date-fns/formatDistanceToNowStrict'
 
 const locales = {}
 let lang = DEFAULT_LANG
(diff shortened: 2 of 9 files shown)


react/BarContextProvider/index.spec.jsx
```diff
@@ -22,7 +22,7 @@ const DumbHelloWorld = translate()(({ t, f, lang }) => (
   <div>
     {t('helloworld')}
     <br />
-    {f('2020-01-06', 'DDD MMM')}
+    {f(new Date('2020-01-06'), 'd MMM')}
```
Crash--, line 25: Can't, `f` deal with this BC by itself? It it's a string, then create a new Date() by itself? In order to help the migration.

react/I18n/format.jsx
```diff
@@ -3,26 +3,31 @@ import { DEFAULT_LANG } from '.'
...
-let lang = DEFAULT_LANG
+let lang = DEFAULT_LANG === 'en' ? 'en-US' : DEFAULT_LANG
 
 const getWarningMessage = lang =>
   `The "${lang}" locale isn't supported by date-fns. or has not been included in the build. Check if you have configured a ContextReplacementPlugin that is too restrictive.`
 
 export const provideDateFnsLocale = (userLang, defaultLang = DEFAULT_LANG) => {
-  lang = userLang
+  const resolvedDefaultLang = defaultLang === 'en' ? 'en-US' : defaultLang
+  const resolvedUserLang = userLang === 'en' ? 'en-US' : userLang
```
Crash--, line 14: This should also be done for `fr-FR`, right? And also `es-ES` since these are the 3 supported langages

react/I18n/format.jsx
```diff
@@ -3,26 +3,31 @@ import { DEFAULT_LANG } from '.'
...
+        resolvedUserLang
+      ] = require(`date-fns/locale/${resolvedUserLang}/index.js`)
     } catch (e) {
-      console.warn(getWarningMessage(lang))
+      console.warn(getWarningMessage(resolvedUserLang))
     }
   }
+  lang = resolvedUserLang
   return locales[lang]
 }
```
Crash--, line 34: It also means that the consuming app should change the call to `<I18n` to pass the right ISO code. Today this call is based on what the stack / flagship are injecting. For instance https://github.com/cozy/mespapiers/blob/48b5ec048b00d29c418f06606cc9ec15c2a203c7/src/targets/browser/setupApp.jsx#L15

react/BarContextProvider/index.spec.jsx
Crash--: Also, is there any codemods that are helping to convert old date notation (ex `DDD MMM`) to the new one?

react/I18n/format.jsx
PolariTOON: French and Spanish localizations have not been moved. They are still in `fr` and `es` directory and `fr-FR` and `es-ES` do not exist, so we should not change anything for them.

PolariTOON:
Superseded by #2730.

PR #2313, opened 2022-12-14
fix: Add padding bottom in immersive app for BottomSheet

zatteo:
I also add to move the immersiveFlagship span which was ignored when at the top of the component.

Capture d’écran 2022-12-14 à 16 21 52

### DIFF
--- a/react/BottomSheet/BottomSheet.jsx
+++ b/react/BottomSheet/BottomSheet.jsx
@@ -48,7 +48,7 @@ const ContainerWrapper = ({ showBackdrop, children }) => {
   return children
 }
 
-const createStyles = ({ squared, hasToolbarProps }) => ({
+const createStyles = ({ squared, hasToolbarProps, immersive }) => ({
   root: {
     borderTopLeftRadius: '1rem',
     borderTopRightRadius: '1rem',
@@ -72,7 +72,8 @@ const createStyles = ({ squared, hasToolbarProps }) => ({
     backgroundColor: 'var(--secondaryTextColor)'
   },
   stack: {
-    backgroundColor: 'var(--defaultBackgroundColor)'
+    backgroundColor: 'var(--defaultBackgroundColor)',
+    paddingBottom: immersive ? 'var(--flagship-bottom-height)' : 0
   },
   bounceSafer: {
     height: 50,
@@ -131,7 +132,8 @@ const BottomSheet = memo(
 
     const styles = createStyles({
       squared,
-      hasToolbarProps
+      hasToolbarProps,
+      immersive: getFlagshipMetadata().immersive
     })
     const overriddenChildren = makeOverridenChildren(children, headerContentRef)
 
@@ -224,9 +226,6 @@ const BottomSheet = memo(
 
     return (
       <ContainerWrapper showBackdrop={showBackdrop}>
-        {getFlagshipMetadata().immersive && (
-          <span style={styles.flagshipImmersive} />
-        )}
         <BackdropOrFragment
           showBackdrop={showBackdrop}
           onClick={() =>
@@ -279,6 +278,9 @@ const BottomSheet = memo(
           </div>
           <div style={{ height: backdrop ? 0 : bottomSpacerHeight }} />
         </MuiBottomSheet>
+        {getFlagshipMetadata().immersive && (
+          <span style={styles.flagshipImmersive} />
+        )}
         {!isBottomPosition && (
           <Fade in timeout={ANIMATION_DURATION}>
             <div style={styles.bounceSafer} />


react/BottomSheet/BottomSheet.jsx
```diff
@@ -72,7 +72,8 @@ const createStyles = ({ squared, hasToolbarProps }) => ({
     backgroundColor: 'var(--secondaryTextColor)'
   },
   stack: {
-    backgroundColor: 'var(--defaultBackgroundColor)'
+    backgroundColor: 'var(--defaultBackgroundColor)',
+    paddingBottom: immersive ? 'var(--flagship-bottom-height)' : 0
```
Crash--, line 76: I don't know why do we need to test this `immersive` stuff. Don't the `--flagship-bottom-height` is updated by the flagship app by itself? If flagship detect that this is not in immersive mode then the value is set to 0? 

@acezard @Ldoppea

react/BottomSheet/BottomSheet.jsx
```diff
@@ -279,6 +278,9 @@ const BottomSheet = memo(
           </div>
           <div style={{ height: backdrop ? 0 : bottomSpacerHeight }} />
         </MuiBottomSheet>
+        {getFlagshipMetadata().immersive && (
+          <span style={styles.flagshipImmersive} />
+        )}
```
Crash--, line 283: OK so it works now for the bottom, but let's populate the bottomsheet with a lot of content in order to have the top of the bottom sheet above your screen. What happens? Don't we have to also put this span at the top of the bottomsheet?

Crash--: In fact, I don't understand the need of this `<span` since now we have a padding ?

zatteo: I did not succeed in adding a margin bottom to the BottomSheet (maybe not possible ?). So I think that is the purpose of the `span` : taking the space of the Android buttons.

But for me it did not work as is, so I add to move it down (why ?).

But then it hides the last item of the BottomSheet. That's why I added the paddingBottom.

JF-Cozy:
est-ce que vous avez essayé avec et sans backdrop ? Le comportement du BS n'est pas le même dans les 2 cas. Attention il y a déjà un div `<div style={{ height: backdrop ? 0 : bottomSpacerHeight }} />` qui sert à faker une hauteur, à avoir si ça ne rentre pas en conflit. Enfin, est-ce possible d'avoir une doc de démo svp ? Le BS c'est très, très sensible comme sujet...

react/BottomSheet/BottomSheet.jsx
acezard: This looks safe to me, on one condition: the bottomSheet in `immersive` mode is never child of something giving it the Native bottom padding (which should not be the case as it is a fixed element)

acezard:
> est-ce que vous avez essayé avec et sans backdrop ? Le comportement du BS n'est pas le même dans les 2 cas. Attention il y a déjà un div `<div style={{ height: backdrop ? 0 : bottomSpacerHeight }} />` qui sert à faker une hauteur, à avoir si ça ne rentre pas en conflit. Enfin, est-ce possible d'avoir une doc de démo svp ? Le BS c'est très, très sensible comme sujet...

On peut ajouter une démo avec un fake mode immersif je vais tester

react/BottomSheet/BottomSheet.jsx
acezard: @zatteo padding-bottom is not sufficient by itself?

acezard (changes requested):
There is also the problem of handling setFlagshipUi() here. It is not implemented. It's an issue because it seems really touchy to add a global setFlagshipUi in BottomSheet. I would rather add a customer modification to the Home component calling BottomSheet

Crash--:
> I would rather add a customer modification to the Home component calling BottomSheet

I really would love not to. I'm OK with doing it right now for a quick fix for the release, BUT the app should not be aware of that stuff. App developer should not know about this. UI's component should adapt themself.

react/BottomSheet/BottomSheet.jsx
Crash--: > No, the --flagship-bottom-height is instantiated at boot and never updated. 

Are you sure about that? For what I can read, we inject this CSS variable when serving it from the http server: https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/httpserver/httpServerProvider.js#L107. 

This method is called from the `CozyProxyWebView` and each time the slug changes, this method is called. So the behavior is dynamic each time the user moves from an app to an other, no? If yes, then we can rely on it. 

cc @Ldoppea

Crash--:
> There is also the problem of handling setFlagshipUi() here. It is not implemented. It's an issue because it seems really touchy to add a global setFlagshipUi in BottomSheet. 

Why are you talking about `setFlagshipUi()` since we don't use it? I think I'm misleading something, but I don't know what

react/BottomSheet/BottomSheet.jsx
acezard: Yes from that point of view you're right, but I was referring to the raw pixel value: https://github.com/cozy/cozy-flagship-app/blob/master/src/libs/dimensions.js#L16

It is created once and does not support change during runtime

Crash--: > It is created once and does not support change during runtime

Yep, but we don't need it. Since it won't change, right?

acezard:
> > There is also the problem of handling setFlagshipUi() here. It is not implemented. It's an issue because it seems really touchy to add a global setFlagshipUi in BottomSheet.
> 
> Why are you talking about `setFlagshipUi()` since we don't use it? I think I'm misleading something, but I don't know what

When the bottomSheet is opened from the Home, the Android buttons remain light, so we can't see them on the white background. I think there should be a call from somewhere asking for black icons on navbar and white icons (above backdrop) in statusBar

Crash--:
OK I got it now, thanks! 

You're saying that this is not implemented yet, but can't we call
```
webviewIntent.call('setFlagshipUI',
..
```

for instance? Is their something wrong with that?

acezard:
> OK I got it now, thanks!
> 
> You're saying that this is not implemented yet, but can't we call
> 
> ```
> webviewIntent.call('setFlagshipUI',
> ..
> ```
> 
> for instance? Is their something wrong with that?

yes, the issue is call it from cozy-ui or cozy-home to be safer?

Crash--:
> yes, the issue is call it from cozy-ui or cozy-home to be safer?

As I said, the app's developer should not know about this flagship stuff. Either the flagship app should handle it by itself, either our UI's lib component should do it. 

We've chose the last one, so let's keep that way, no?

react/BottomSheet/BottomSheet.jsx
acezard: > > It is created once and does not support change during runtime
> 
> Yep, but we don't need it. Since it won't change, right?

Yes, so the value will always be the same number in immersive or not immersive webview

Crash--: > Yes, so the value will always be the same number in immersive or not immersive webview.

I don't get it. 

In immersive, the value will be the result of `getNavbarHeight()`. So it can be 0 or an other value depending on the device. 

In not immersive webview, the variable will not be defined, so we'll have a 0 value. 

Is that right?

acezard: Yes, that is correct because we inject the variable only on `home` slug. So technically we could ignore the immersive or not and only fetch the CSS value. It would currently work but I think it's more clear if we check for immersive presence. In the other case, if we just fetch the CSS value (that exists or does not exists), we give the impression we always want that value, not knowing it does not exist in certain contexts.
But all this is up to opinion

Crash--: > not knowing it does not exist in certain contexts.

Maybe this is what we should fix. Always having this var defined, but its value can change for each app. Like that, we always use it, we don't need to ask ourselves if we're immersive or not. The component should handle this value as it should handle the `safe-area` stuff.

react/BottomSheet/BottomSheet.jsx
zatteo: Nope just adding a padding-bottom is not enough :/

react/BottomSheet/BottomSheet.jsx
acezard: > > not knowing it does not exist in certain contexts.
> 
> Maybe this is what we should fix. Always having this var defined, but its value can change for each app. Like that, we always use it, we don't need to ask ourselves if we're immersive or not. The component should handle this value as it should handle the `safe-area` stuff.

ok, it would be a simple fix in the native app to add a 0 value to this CSS var if we don't need it.
@zatteo in that case you can remove mention to immersive and just use the CSS var

Ldoppea: I agree that this immersive variable is a bit redundant. It give more context, but maybe too much? Do we really want to know from apps that the margin should be applied for "immersive" purpose? Or do we just need to know that we can manipulate a value that ensure "safe area"?

acezard:
I am closing this pull request, I will handle the issue in another one.

PR #2327, opened 2023-01-12
refactor: Use makeStyles from cozy-ui instead of mui

JF-Cozy:
this is the same comp at the end, but this way we have only one way to import it


Crash-- (approved):
👍

PR #2341, opened 2023-01-31
Port "Make sure the message fills all the available width in `Alert` and allow overriding its classes" to `next`

PolariTOON:
This forward ports #2333 (identical changes).

Demo link : https://polaritoon.github.io/cozy-ui/react/#!/Alert


PolariTOON:
@JF-Cozy Can we merge as is? It seems the same test is already failing on the `next` branch.

JF-Cozy:
> @JF-Cozy Can we merge as is? It seems the same test is already failing on the `next` branch.

Non, les tests doivent être verts. Il faut les restart si c'est côté CI que ça déconne, sinon les corriger

PR #2380, opened 2023-03-20
feat: Add success/fail animations to SquareAppIcon

acezard:
From variant 'loading' to either variant 'default' or 'error'

https://acezard.github.io/cozy-ui/react/#!/SquareAppIcon

image

### DIFF
--- a/react/SquareAppIcon/styles.styl
+++ b/react/SquareAppIcon/styles.styl
@@ -55,7 +55,77 @@ $color = constants['color'] // hard-coded color, because the component is curren
         transform scale(0.8334)
 
 .SquareAppIcon-spinner
+    margin 0 !important
+
     svg
         position absolute
         height 100%
         width 100%
+
+animation-duration = 1.50s
+icon-duration = 0.2s
+icon-start = 0.2s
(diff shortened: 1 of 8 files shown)


react/SquareAppIcon/styles.styl
```diff
@@ -55,7 +55,77 @@ $color = constants['color'] // hard-coded color, because the component is curren
...
 
 .SquareAppIcon-spinner
+    margin 0 !important
+
     svg
         position absolute
         height 100%
         width 100%
+
+animation-duration = 1.50s
```
Ldoppea, line 65: I'm not used to stylus files. Is that an equivalent to css `var()` ? If yes, is this scoped to this file?

Ldoppea (approved):

react/SquareAppIcon/styles.styl
acezard: they will be local to that file, they are not equivalent to CSS var but more like sass variables. I didn't feel the need to have them as CSS vars because I don't have to change them at runtime (animationColor though, needs to be changed at runtime and thus is a CSS var)

PR #2389, opened 2023-03-23
feat: Export Breadcrumbs from MUI

zatteo:
In cozy-drive we have issues because we import the Breadcrumbs component directly from material-ui. Here I just reexport it.

Question : what name should I use ? Because we already have a Breadcrumbs legacy component.


Crash--:
Can you name it MuiBreadcrumbs? I know this is not ideal but at least we have  a distinction between the two components

Crash-- (approved):

cballevre (approved):

JF-Cozy (approved):

PR #2410, opened 2023-04-06
fix(SelectionBar): Limit the number of actions displayed

cballevre:
In cozy-drive, up to 8 or more actions can be displayed at the same time. On both large and tiny screens, the actions overflow behind the close button. To solve this problem, we decided to limit the number of actions displayed for certain screen sizes. Actions exceeding this limit are displayed in an additional menu. 

problem

**Live preview:** https://cballevre.github.io/cozy-ui/react/#!/SelectionBar/

**Some examples on Drive:**

Large screen size:
isLarge

Medium screen size:
isMedium

Tiny screen size: 

Large screen size on banks : 
banks-large

### DIFF
--- a/react/SelectionBar/helpers.js
+++ b/react/SelectionBar/helpers.js
@@ -1,0 +1,22 @@
+/**
+ * Compute the maximum number of actions to display according to the breakpoint
+ * @param {number|object} maxAction an number or a set of maximum for each breakpoint
+ * @param {object} breakpoints a set of breakpoints with a boolean to know if it is active or not
+ * @returns the maximum number or undefined if there is no matching result
+ */
+export function computeMaxAction(maxAction, breakpoints) {
+  if (typeof maxAction === 'object' && maxAction !== null && breakpoints) {
+    const maxActionKeys = Object.keys(maxAction)
+    const currentMaxAction = Object.keys(breakpoints)
+      .filter(key => breakpoints[key])
+      .filter(key => maxActionKeys.includes(key))
+
+    if (currentMaxAction.length > 0) {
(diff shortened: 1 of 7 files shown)


react/SelectionBar/index.jsx
zatteo, line 1: Nice refacto 🎉

react/SelectionBar/helpers.js
```diff
@@ -0,0 +1,16 @@
+export function computeMaxAction(maxAction, breakpoints) {
+  if (typeof maxAction === 'object' && maxAction !== null) {
```
zatteo, line 2: nit: if maxAction === 'object', it is necessarily !== null no ?

react/SelectionBar/helpers.js
```diff
@@ -0,0 +1,16 @@
+export function computeMaxAction(maxAction, breakpoints) {
```
zatteo, line 1: nit: Tests would be nice here and will help to understand what the function do

zatteo:
Is it possible to see the SelectionBarMore in cozy-ui demo ?

react/SelectionBar/helpers.js
cballevre: This prevent an error on `Object.keys` because in javascript `typeof null` returns `object` ([reason](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null))

zatteo: Wow. I just learnt something...

zatteo (approved):

PR #2411, opened 2023-04-11
fix: Delete `max-width` on Input component

Merkur39:
Removal of the maximum width of the Input component, because in the case where the Input is very wide, not all of it can be clicked


Crash--:
This is a big BC here. Maybe it will break a few apps. I don't think a need a codemod, but a detailed description should be enough

PR #2415, opened 2023-04-14
feat: Add isExtraExtraLarge breakpoint

cballevre:
This allows more granularity in the breakpoint for extra large devices between 1200px and 1400px. This change follows the device ranges used by bootstrap (cf: [documentation](https://getbootstrap.com/docs/5.0/layout/breakpoints/))

BREAKING CHANGE : Before, we used `isExtraLarge` to get the same result now we have to use `isExtraLarge || isExtraExtraLarge` because `isExtraLarge` has a limit now and not `isExtraExtraLarge`

From my research, this break change has relatively low impact (cf: [research](https://github.com/search?q=org%3Acozy+isExtraLarge&type=code). It is used once in cozy-drive and 3 times in cozy-banks.

### DIFF
--- a/react/helpers/breakpoints.js
+++ b/react/helpers/breakpoints.js
@@ -1,12 +1,14 @@
 import mapValues from 'lodash/mapValues'
 
+const extraLarge = 1400
 const large = 1200
 const medium = 1023
 const small = 768
 const tiny = 543
 
 const breakpoints = {
-  isExtraLarge: [large + 1],
+  isExtraExtraLarge: [extraLarge + 1],
+  isExtraLarge: [large + 1, extraLarge],
   isLarge: [medium + 1, large],
   isMedium: [small + 1, medium],
   isSmall: [tiny + 1, small],


zatteo (approved):

react/helpers/breakpoints.js
```diff
@@ -1,12 +1,14 @@
 import mapValues from 'lodash/mapValues'
 
+const extraLarge = 1400
```
JF-Cozy, line 3: Peut-être quelque chose de plus "parlant" :

```
const huge = 1400

const breakpoints = {
  isWide: [huge + 1],
  isHuge: [large + 1, huge],
```

De plus, il y avait un sujet concernant ces breakpoints, de potentiellement les réduire... ici on en rajoute encore, je prends le temps de me remettre dans le sujet et on fait un point. Je te transfère le paper en question...

cballevre:
In the end we decided not to add one more breakpoint. isTiny, isSmall, isMedium, isLarge, isExtraLarge will be depreciated in favour of isMobile, isTablet, isDesktop. Which are already used in most cases.

PR #2417, opened 2023-04-19
feat: Add useScroll

cballevre:
This PR adds a new useScroll hook that gets the vertical scroll position. You can see a preview of the documentation here : https://cballevre.github.io/cozy-ui/react/#!/UseScroll

This hook will be used to make extended FABs only at the top of the screen to improve discoverability like in My Papers. I have added an example of this behaviour in the FAB documentation. You can see a preview here: https://cballevre.github.io/cozy-ui/react/#!/Fab

This work is largely based on @Merkur39's work on My Papers (see: [source](https://github.com/cozy/cozy-libs/blob/master/packages/cozy-mespapiers-lib/src/components/Hooks/useScroll.jsx))


Merkur39 (commented):
Bravo, et merci pour le portage ! 🤗

react/hooks/useEventListener.spec.js
```diff
@@ -30,6 +47,15 @@ describe('useEventListener', () => {
     expect(result.error).not.toBeDefined()
   })
 
+  it('should not subscribe for an undefined ref element', async () => {
+    const cb = jest.fn()
+    const ref = {
+      current: undefined
+    }
+    const { result } = renderHook(() => useEventListener(ref, 'click', cb))
+    expect(result.error).not.toBeDefined()
```
zatteo, line 56: Why should result.error be not defined ?

zatteo:
Just for information if you don't know this, React 18 has a new way to do this kind of thing with useSyncExternalStore

https://react.dev/reference/react/useSyncExternalStore#subscribing-to-a-browser-api
https://julesblom.com/writing/usesyncexternalstore#why-not-useeffect-with-usestate

zatteo (approved):

react/hooks/useScroll.jsx
```diff
@@ -0,0 +1,44 @@
+import { useState } from 'react'
+import debounce from 'lodash/debounce'
+
+import useEventListener from './useEventListener'
+import { unRef } from '../helpers/ref'
+
+/**
+ * Get scrollTop & direction from scroll event
```
Merkur39, line 8:
```suggestion
 * Get scrollTop & scrollLeft from scroll event
```

Merkur39 (approved):
LGTM 👏

react/hooks/useEventListener.spec.js
cballevre: I done as [line 47](https://github.com/cozy/cozy-ui/pull/2417/files/800fda4f3b1081c7012afddcb7bdb2268b35916d#diff-05864bb9fd827c4bc4b3a26e900bb6e37d706e8213eee55251fbfee49bd53e9eR47) it would have an error if we tried to add an event listened on nothing

PR #2466, opened 2023-06-28
docs: Add a fake first component to reduce first time loading

JF-Cozy:
the `core`, `extra`, `legacy` groups in the documentation contains now a lot of components, so the first load of these pages is very long. This is a quick fix but we need to investigate further (and maybe migrates to storybook...)

démo : https://jf-cozy.github.io/cozy-ui/react/


Merkur39 (approved):
storybook 🤗

PR #2488, opened 2023-08-09
feat(Viewer): Display file path into the toolbar

cballevre:
[Live demo](https://cballevre.github.io/cozy-ui/react/#!/Extra/Viewer)

Capture d’écran 2023-08-09 à 12 00 04

**Remaining works:**
- ~Make a functional example into documentation (mock query to get parent path)~

**Related PRs:**
- https://github.com/cozy/cozy-libs/pull/2272

### DIFF
--- a/react/Viewer/Readme.md
+++ b/react/Viewer/Readme.md
@@ -34,6 +34,8 @@ The `Viewer` can display an **information panel** to show additional information
     * **toolbarRef** : `<object>` – React reference of the toolbar node
     * **showToolbar** : `<boolean>` – Whether to show the toolbar or not. Note that the built-in close button is in the toolbar
     * **showClose** : `<boolean>` – Whether to show close button in toolbar
+    * **showFilePath** : `<boolean>` – Whether to show file path below his name
+
 
 ### Demo
 
@@ -68,13 +70,15 @@ const files = [
     _id: 'audio',
     class: 'audio',
     name: 'Sample.mp3',
-    mime: 'audio/mp3'
+    mime: 'audio/mp3',
+    dir_id: 'parent_folder'
   },
   {
     _id: 'slide',
     class: 'slide',
     name: 'Slide.pptx',
-    mime: 'application/vnd.openxmlformats-officedocument.presentationml.presentation'
+    mime: 'application/vnd.openxmlformats-officedocument.presentationml.presentation',
+    dir_id: 'parent_folder'
   },
   {
     _id: 'pdf',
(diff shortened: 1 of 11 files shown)


react/Viewer/Readme.md
Crash--, line 1: Did you create a codemod for that? If yes, you should add it to the commit message

react/Viewer/helpers.js
```diff
@@ -168,3 +169,31 @@ export const isEditableAttribute = (name, file) => {
...
+export const normalizeAndSpreadAttributes = rawFile => {
+  const normalizedFile = normalize(rawFile)
+
+  return {
+    ...normalizedFile,
+    ...normalizedFile?.attributes
+  }
+}
+
+export const removeFilename = path => /(.+)(\/.+)$/.exec(path)?.[1] || path
```
Crash--, line 182: You have the real filename where you call this method. You should pass it to this method along side with the path. And then you just have to remove the all string instead of doing regex?

react/Viewer/Readme.md
cballevre: Yes, I have created one. The instructions are in the second commit message.

Crash--: Let's add it to this one to. Or maybe we should squash the both?

react/Viewer/components/ToolbarFilePath.jsx
```diff
@@ -0,0 +1,42 @@
...
+  const client = useClient()
+  const { isDesktop } = useBreakpoints()
+
+  const fileWithPath = useFileWithPath(file)
+
+  if (fileWithPath) {
+    const app = 'drive'
+    const nativePath = `/folder/${fileWithPath.dir_id}`
+    const path = removeFilename(fileWithPath.path)
+    const link = makeParentWebLink({ client, nativePath, app })
```
zatteo, line 23: Is this function correctly named? It generates parent web link just because you pass parent nativePath no?

zatteo:
Happy to see `onlyOfficeProps` removed!

react/Viewer/Readme.md
cballevre: The idea with @JF-Cozy was that the commit messages would be merged into the changelog and the codemod instructions would be described for the version

react/Viewer/helpers.js
cballevre: Yes, this was a copy and paste from cozy-notes without thinking about it. Now, I use `.replace` instead of an regex

react/Viewer/components/ToolbarFilePath.jsx
cballevre: I've renamed the function to `makeWebLink` and made the parameters more generic. Does this seem more clear to you?

Crash-- (approved):

PR #2495, opened 2023-08-30
fix(Viewer): Remove filename on path ending with slash

cballevre:
When the file was at the root and the path ended with a slash, the path displayed to the user was `//`. I've improved the function to avoid this

### DIFF
--- a/react/Viewer/helpers.js
+++ b/react/Viewer/helpers.js
@@ -203,3 +203,13 @@ export const makeWebLink = ({ client, slug, path }) => {
     return null
   }
 }
+
+/**
+ * Remove the file name at the end of a path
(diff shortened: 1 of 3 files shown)


zatteo (approved):

react/Viewer/helpers.js
```diff
@@ -203,3 +203,13 @@ export const makeWebLink = ({ client, slug, path }) => {
     return null
   }
 }
+
+/**
+ * Remove the file name at the end of a path
+ * @param {string} path
+ * @returns new path
```
Crash--, line 210:
```suggestion
 * @returns {string} new path
```

nit(you can add //@ts-check at the top of the file ;))

PR #2504, opened 2023-09-15
feat(NestedSelect)/Manage ellipsis

Merkur39:
In order to keep the default behavior of ListItem(ListItemtext), we only use the `ellipsis` prop when it is `false`.

### DIFF
--- a/react/NestedSelect/ItemRow.jsx
+++ b/react/NestedSelect/ItemRow.jsx
@@ -12,13 +12,21 @@ import Typography from '../Typography'
 
 const infoStyle = { color: 'var(--secondaryTextColor)' }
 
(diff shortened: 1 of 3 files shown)


Crash--:
is it a breaking change?

Merkur39:
> is it a breaking change?

This is not a BC, the `ellipsis` prop is `true` by default on `ListItemText` (used by `ListItem`).
Here we just add the possibility to disable it within `NestedSelectModal`
This will not change any existing uses of `NestedSelectModal`

> Shouldn't we have a few changes on argos?

Indeed, I updated the doc

Crash-- (approved):

PR #2511, opened 2023-09-29
feat(Viewer): On mobile, the title benefits well from the MidEllipsis

Merkur39:
Demo: https://merkur39.github.io/cozy-ui/react/#!/Extra/Viewer

Before:
Capture d’écran 2023-09-29 à 11 24 20

After:
Capture d’écran 2023-09-29 à 11 23 59

### DIFF
--- a/react/Viewer/components/Toolbar.jsx
+++ b/react/Viewer/components/Toolbar.jsx
@@ -20,6 +20,7 @@ import { useEncrypted } from '../providers/EncryptedProvider'
 import { ToolbarFilePath } from './ToolbarFilePath'
 
 import styles from './styles.styl'
+import MidEllipsis from '../../MidEllipsis'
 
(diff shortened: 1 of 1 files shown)


Crash--:
can you add an exemple with the extension displayed?

Merkur39:
> can you add an exemple with the extension displayed?

Here with MidEllipsis component:
Capture d’écran 2023-09-29 à 12 07 22

Crash--:
> Here with MidEllipsis component:

And without? Because we don't want the caller to add this component, right?

Merkur39:
> > Here with MidEllipsis component:
> 
> And without? Because we don't want the caller to add this component, right? Or we don't need to add it because it's already present somewhere?

The MidEllipsis component is well used within the Viewer, this does not impact the way of using the Viewer

Crash-- (approved):

PR #2552, opened 2023-12-14
chore: Don't run CI for branches if not master

JF-Cozy:
Travis is configured to run on branches and PRs. The problem is that the CI is triggered at force push for all branches even if there is no associated PR, and for those with PRs the CI is therefore triggered twice.


cballevre (approved):

cballevre:
Nice it will free up bandwidth for other builds 👌

PR #2593, opened 2024-03-15
chore: Describe default type attribute in Buttons documentation

Ldoppea:
MUI documentation does not describe default button's `type`, nor explain that is it possible to set classic HTML attributes, so it is easy to forget applying the correct button type when inside of a form

This is what happened on a previous PR when migrating from old `react/deprecated/Button` to new `react/Buttons` as those components have opposite default behaviors

Related PR: cozy/cozy-settings#674

### DIFF
--- a/react/Buttons/Readme.md
+++ b/react/Buttons/Readme.md
@@ -1,3 +1,11 @@
+Buttons are a wrapper arround [MuiButton](https://mui.com/material-ui/react-button/) integrated with Cozy's design system
+
+The `Button` component will produce an HTML `<button>`. The React component accepts classic HTML attributes (i.e. `type`) through React props.
+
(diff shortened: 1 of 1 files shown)


react/Buttons/Readme.md
```diff
@@ -1,4 +1,10 @@
-The `ref` is forwarded to the root element
```
cballevre, line 1: Is it deliberate to remove this line?

cballevre (approved):
Thank you for documenting this point 👍

react/Buttons/Readme.md
Ldoppea: Yes, I explained why on [the first commit](https://github.com/cozy/cozy-ui/pull/2593/commits/b38b6a02259aa87143677da0f4318db1bc9a6ff0) 😋

JF-Cozy (approved):
pour les commit, ce ne sont pas des "chore" mais des "docs" :)

PR #2601, opened 2024-03-25
fix(Viewer): Remove WebviewIntentProvider

Merkur39:
In commit 97f28ff, WebviewIntentProvider has been added to ViewerContainer, in order to call intent's methods in the container children.

This is not needed because we expect the webapp to implemented on the root level.

Also, this would produce a bug because WebviewIntentProvider is not mean to be declared twice in a same webapp.

When doing so, the child WebviewIntentProvider would overwrite the parent one on the flagship app side. So when opening the Viewer, the parent provider would stop working.

### DIFF
--- a/react/Viewer/Readme.md
+++ b/react/Viewer/Readme.md
@@ -4,12 +4,11 @@ Once rendered, the `Viewer` will take up all the available space in it's contain
 
 The `Viewer` can display an **information panel** to show additional information about the current file (e.g. whether a file is certified).
 
-### ⚠️ Important
+### ⚠️ Requirement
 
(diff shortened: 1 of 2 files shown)


JF-Cozy (approved):
C'est maintenant un pré-requis pour que le Viewer fonctionne, au même titre que useI18n, je pense qu'on devrait au moins le documenter, et peut-être faire un BC (mais moins sûr pour ce coup là)

zatteo:
> When doing so, the child WebviewIntentProvider would overwrite the parent one on the flagship app side. So when opening the Viewer, the parent provider would stop working.

Interesting, I did not know about this.

PR #2609, opened 2024-04-05
fix: Add vertical scroll to sidebar when needed (VO-628)

acezard:
This will handle cases in desktop where the
sidebar was hiding its own content because it was too tall
and not letting the user scroll inside it.
This should have no effect on the mobile sidebar
since it has a fixed height


acezard:
argos false negative

JF-Cozy (approved):

PR #2641, opened 2024-05-29
docs(ListItem): Use date string instead function

Merkur39:
Calling the date function generates unnecessary changes in Argos.


JF-Cozy (approved):
👍 😘

PR #2642, opened 2024-05-30
[VO-758] feat(AppsSection): Change only ISP wording

cballevre:
The telecom category is still used because konnector installed needs to run to updated there manifest.

This reverts a part of commit d5d45988929961fd076039cdf4bc60b12817bac5.


JF-Cozy (approved):
attention à bien taguer le commit comme étant un fix alors stp

PR #2646, opened 2024-06-04
feat: Improve contacts list performances

paultranvan:
When dealing with thousands of contacts, the default 100 pagination will make dozens of requests, thus making it quite slow.

### DIFF
--- a/react/ContactsListModal/queries.js
+++ b/react/ContactsListModal/queries.js
@@ -4,7 +4,7 @@ import { CONTACTS_DOCTYPE } from 'cozy-client/dist/models/contact'
 const defaultFetchPolicy = fetchPolicies.olderThan(30 * 1000)
 
 export const buildContactsQuery = () => ({
-  definition: () => Q(CONTACTS_DOCTYPE),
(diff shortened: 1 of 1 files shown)


zatteo (approved):
Nice quick win 🎉

paultranvan:
I see travis is failing with `Build failed: Build rejected: subscribe to a Pro plan to use Team features` and after `"if [[ "${PR_TITLE}" != *"[skip argos]"* ]]; then...`. Does it mean:
- Argos is currently broken as long as we do not have the subscription?
- I need to add `[skip argos]` in the PR?
- Should we deactivate argos?

JF-Cozy:
> I see travis is failing with `Build failed: Build rejected: subscribe to a Pro plan to use Team features` and after `"if [[ "${PR_TITLE}" != *"[skip argos]"* ]]; then...`. Does it mean:
> 
>     * Argos is currently broken as long as we do not have the subscription?
> 
>     * I need to add `[skip argos]` in the PR?
> 
>     * Should we deactivate argos?

C'est maintenant corrigé, il faut relancer le build et tenir compte de Argos

PR #2658, opened 2024-06-17
feat(CozyTheme): Use `io.cozy.settings` to determine default type

JF-Cozy:
BREAKING CHANGE: To use `<CozyTheme>` properly, you must have permission in your app to get `io.cozy.settings` documents. The experience would be even better with realtime on this doctypes.

### DIFF
--- a/react/providers/CozyTheme/index.jsx
+++ b/react/providers/CozyTheme/index.jsx
@@ -1,12 +1,14 @@
 import React, { createContext, useContext, useLayoutEffect } from 'react'
 import PropTypes from 'prop-types'
 import cx from 'classnames'
-import flag from 'cozy-flags'
 
+import flag from 'cozy-flags'
 import log from 'cozy-logger'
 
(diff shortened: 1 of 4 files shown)


react/providers/CozyTheme/index.jsx
```diff
@@ -25,7 +28,30 @@ export const useCozyTheme = () => {
...
+const CozyThemeQuery = props => {
+  const instanceQuery = buildSettingsInstanceQuery()
+  const { data: instance, ...instanceQueryLeft } = useQuery(
+    instanceQuery.definition,
+    instanceQuery.options
+  )
+
+  if (
+    isQueryLoading(instanceQueryLeft) &&
+    !hasQueryBeenLoaded(instanceQueryLeft)
```
zatteo, line 40: It is useful to check if query is loading AND if not query been loaded ?

JF-Cozy: c'est pour éviter de repasser à null après expiration de la fetchpolicy avec un fetch en arrière plan

PR #2663, opened 2024-06-28
[SCR-683] feat: Add new categories to appSections locales

KillianCourvoisier:
This PR adds recently added categories to appSections to display correctly in the store


cballevre (approved):
Looks good to me 👍

PR #2675, opened 2024-07-18
[SCR-785] typo: Replace commercial '&' by their equivalents

KillianCourvoisier:
Replace "&" char by their equivalent in each language to uniformize display in store


cballevre (approved):
LGTM 👍

PR #2693, opened 2024-09-04
[VO-710] feat: Add Markdown component

cballevre:
This component is responsible to render Markdown with Cozy design. At Cozy, we use component to style whereas React-markdown transform markdown with html tag only so the style is not correct. In some apps ([cozy-settings](https://github.com/cozy/cozy-store/blob/ba6323e6e0176ba415c2b7454884dca82a93fe75/src/ducks/components/ReactMarkdownWrapper.jsx), [cozy-store](https://github.com/cozy/cozy-store/blob/ba6323e6e0176ba415c2b7454884dca82a93fe75/src/ducks/components/ReactMarkdownWrapper.jsx), [harvest](https://github.com/cozy/cozy-libs/blob/abcbff2c15896acda8853b875b5123303936e09f/packages/cozy-harvest-lib/src/components/Markdown.jsx), [cozy-ui](https://github.com/cozy/cozy-ui/blob/97882fe42fee48fc63e994379a6190e35d2504d3/react/Paywall/Paywall.jsx)) we already use this library but each time make a new version of the wrapper. This PR allows you to share the component and its styles between apps.

You can found a demo on my github : https://cballevre.github.io/cozy-ui/react/#!/Core/Markdown

Capture d’écran 2024-09-04 à 15 39 19


react/Markdown/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+          <Typography variant="body1" className="u-mb-1">
+            {children}
+          </Typography>
+        )
+      }}
+    />
+  )
+}
+
+export { Markdown }
```
JF-Cozy, line 32: je sais qu'il y a débat, mais par souci de cohérence, peut-on faire une export par défaut stp ? 90% des composants de cozy-ui sont en default.

JF-Cozy (approved):

react/Markdown/index.jsx
cballevre: Yep je te fais ça

PR #2715, opened 2024-11-07
fix: Import resolution problem inside Storybook of cozy-harvest

cballevre:
The Storybook for Harvest could not resolve cozy-ui/react/ as it isn't present in the transpiled version. To resolve this, I updated the import to use relative paths consistently with the rest of the codebase. Additionally, as ShortcutTile is a TypeScript file, I needed typed versions of all components it relies on, so I added JSDoc to Typography to ensure compatibility.

### DIFF
--- a/react/Typography/index.jsx
+++ b/react/Typography/index.jsx
@@ -1,6 +1,16 @@
 import MuiTypography from '@material-ui/core/Typography'
 import React, { forwardRef } from 'react'
 
+/**
+ * @typedef TypographyPropTypes
+ * @property {string} [color] - The color of the text.
+ * @property {string} [variant] - The variant of the text.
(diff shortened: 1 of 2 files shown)


react/Typography/index.jsx
```diff
@@ -1,6 +1,16 @@
 import MuiTypography from '@material-ui/core/Typography'
 import React, { forwardRef } from 'react'
 
+/**
+ * @typedef TypographyPropTypes
+ * @property {string} [color] - The color of the text.
+ * @property {string} [variant] - The variant of the text.
+ * @property {React.ReactNode} children - The content of the component.
+ */
```
JF-Cozy, line 9: on aurait peut-être pu mettre toutes les props... 🤔 https://v4.mui.com/api/typography/#typography-api

PR #2717, opened 2024-11-15
[VO-1194] feat(SquareAppIcon): Hide shortcut badges for a more app-like feel

cballevre:
By default, the badges are displayed for the shortcut variant, but in some cases you want to be able to hide them so that they look more like applications

**Demo :** https://cballevre.github.io/cozy-ui/react/#!/Extra/SquareAppIcon


zatteo (approved):
👍 👍

PR #2733, opened 2024-12-24
feat(DatePicker): Add DatePicker component

Merkur39:
This component is already in use on the MesPapiers app, and we're going to need it on cozy-sharing, so we're going to add it to cozy-ui while making it more generic.

The style has been kept from the original, as it was a generic need and not specific to the app.

As the official documentation is no longer online, I've copied the API into the component documentation.

The `@date-io/date-fns` package is blocked in v1 for this reason:
https://github.com/dmtrKovalenko/date-io/releases/tag/v2.0.0

Demo: https://merkur39.github.io/cozy-ui/react/#!/Extra/DatePicker


react/DatePicker/locales/withOwnLocales.jsx
```diff
@@ -0,0 +1,10 @@
+import en from './en.json'
+import fr from './fr.json'
+import withOnlyLocales from '../../providers/I18n/withOnlyLocales'
+
+export const locales = {
+  en,
+  fr
+}
+
+export default withOnlyLocales(locales)
```
zatteo, line 10: `withOnlyLocales` ?

package.json
```diff
@@ -156,8 +156,10 @@
   },
   "dependencies": {
     "@babel/runtime": "^7.3.4",
+    "@date-io/date-fns": "1",
```
zatteo, line 159: What is this version ? Can't we have a more recent version ?

zatteo:
Commit message is a work in progress message

react/DatePicker/index.jsx
```diff
@@ -0,0 +1,302 @@
...
+
+const localesFNS = {
+  fr: LocaleFR,
+  en: LocaleEN
+}
+
+const useStyles = makeStyles(() => ({
+  overrides: {
+    width: '100%',
+    height: isDesktop => (isDesktop ? '5rem' : 'inehrit'),
```
zatteo, line 34: `inehrit`-> `inherit`

react/DatePicker/index.jsx
```diff
@@ -0,0 +1,302 @@
...
+        setError(null)
+        isValid?.(true)
+        onChange(val)
+      } else if (val === '') {
+        setError(null)
+        isValid?.(true)
+        onChange(null)
+      } else {
+        setError(errorLabel ?? t('invalidDate'))
+        isValid?.(false)
```
zatteo, line 123: What is the purpose of `isValid` ? Is it a callback ?

react/DatePicker/index.jsx
```diff
@@ -0,0 +1,302 @@
...
+        break
+      case 'dateTime':
+        DatePickerComponent = enableKeyboard
+          ? MuiKeyboardDateTimePicker
+          : MuiDateTimePicker
+        break
+      default:
+        DatePickerComponent = enableKeyboard
+          ? MuiKeyboardDatePicker
+          : MuiDatePicker
```
zatteo, line 159: Default mode is already handled in props destructuration `mode = 'date'`

package.json
Merkur39: That's why: [Do not update if you are using @material-ui/pickers v3](https://github.com/dmtrKovalenko/date-io/releases/tag/v2.0.0)

react/DatePicker/index.jsx
Merkur39: Yes, for example, in our apps we very often use modals, and in this case we also want to make the modal buttons dynamic.
This callback is used to obtain the validity or invalidity of the field in order to pass it on to other components.

zatteo: Ok 👍 nit: Is it confusing to have a callback that does not start with `on`.

zatteo (approved):

package.json
zatteo: Maybe to add in the commit message

react/DatePicker/index.jsx
Merkur39: Indeed, I don't like this approach either.
We opted for a second parameter in the `onChange` callback.

zatteo:
A clean PR for a new year 🎉

PR #2762, opened 2025-04-01
Add u-bg-transparent and u-elevation-0

zatteo:


stylus/utilities/bgcolor.styl
```diff
@@ -8,3 +8,6 @@ for color in utility_colors
         background-color: json_colors[color] !important // @stylint ignore
 
     global('.u-bg-' + color, '$bg-color-' + color)
+
+.u-bg-transparent
```
JF-Cozy, line 12: no need for `global()` ?

stylus/utilities/elevation.styl
```diff
@@ -1,4 +1,7 @@
 @require '../settings/shadows'
 
+.u-elevation-0
+    box-shadow none !important // @stylint ignore
```
JF-Cozy, line 4: we should rely on same approach, so `extend $elevation-0` even it's legacy

JF-Cozy (approved):

PR #2763, opened 2025-04-01
Update search bar style

zatteo:


JF-Cozy (approved):
typo in the BC message `string` instead of `bool`

JF-Cozy:
We have to deal with the BC inside cozy-ui component and doc, cf Argos

PR #2782, opened 2025-04-18
feat: Update --primaryFont to Inter

zatteo:
BREAKING CHANGE: --primaryFont has been updated to Inter

### DIFF
--- a/stylus/settings/fontstack.styl
+++ b/stylus/settings/fontstack.styl
@@ -5,7 +5,7 @@
 /*
     Font Stack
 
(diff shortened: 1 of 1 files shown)


JF-Cozy:
This is done in the override for now, so to avoid forgetting anything, we need to replace everything in the override here or there where necessary

zatteo:
Done here https://github.com/cozy/cozy-ui/pull/2791/commits/d70f8f46f9aa49a0b4663ebea965161b70840e5a

PR #2788, opened 2025-05-07
fix: Wait for flags initialization before making MUI theme

Ldoppea:
Since #2748 the `makeTheme()` methods relies on cozy's flags

Because flags are initialized asynchronously, the `makeTheme()` methods were called too early and so on the first cozy-app render, the app would render with the incorrect theme. Then after a refresh, because the flags is now on the local storage, then the app would render with the correct theme

To reproduce this bug, either:
- clear the app storage, re-login and access the app
- or clear the `flag__ui.theme-twake.enabled` entry in local storage

By awaiting for flags initialization the first cozy-app render would display the correct theme

### DIFF
--- a/react/MuiCozyTheme/index.jsx
+++ b/react/MuiCozyTheme/index.jsx
@@ -1,11 +1,29 @@
 import PropTypes from 'prop-types'
-import React from 'react'
+import React, { useEffect, useState } from 'react'
+
+import { useClient } from 'cozy-client'
 
 import { getTheme } from './theme'
 import { ThemeProvider } from '../styles'
 
 const MuiCozyTheme = ({ type, variant, children }) => {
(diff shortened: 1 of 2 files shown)


react/MuiCozyTheme/theme.jsx
JF-Cozy, line 1: is this still relevant?

react/MuiCozyTheme/index.jsx
JF-Cozy, line 1: can we add big comment to remove all of this when purging cozy theme please?

JF-Cozy (approved):

JF-Cozy:
maybe we can close it because of https://github.com/cozy/cozy-ui/pull/2790 ?

PR #2798, opened 2025-05-19
fix: Don't register Flagship UI on ActionsMenu until opened

Ldoppea:
With previous implementation, useActionMenuEffects() would call SetFlagshipUI even if the menu is not open

This would result with incorrect top bar color in Flagship app

By moving the hook call in a custom component under the `open` condition, we ensure the hook will register/unregister correctly when the ActionsMenu opens or closes

### DIFF
--- a/react/ActionsMenu/ActionsMenuWrapper.jsx
+++ b/react/ActionsMenu/ActionsMenuWrapper.jsx
@@ -8,6 +8,12 @@ import { useActionMenuEffects } from '../deprecated/ActionMenu/ActionMenuEffects
 import isTesting from '../helpers/isTesting'
 import useBreakpoints from '../providers/Breakpoints'
 
+const ActionsMenuFlagshipUI = ({ children }) => {
+  useActionMenuEffects()
+
+  return children
+}
+
 const ActionsMenuWrapper = ({
   children,
   autoClose,
@@ -17,8 +23,6 @@ const ActionsMenuWrapper = ({
 }) => {
   const { isMobile } = useBreakpoints()
 
-  useActionMenuEffects()
-
   const overrideClick = props => ev => {
     props.onClick?.(ev) // this is ActionsItems onClick prop
     autoClose && onClose?.()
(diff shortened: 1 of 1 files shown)


react/ActionsMenu/ActionsMenuWrapper.jsx
```diff
@@ -8,6 +8,12 @@ import { useActionMenuEffects } from '../deprecated/ActionMenu/ActionMenuEffects
 import isTesting from '../helpers/isTesting'
 import useBreakpoints from '../providers/Breakpoints'
 
+const ActionsMenuFlagshipUI = ({ children }) => {
+  useActionMenuEffects()
```
zatteo, line 12: useActionMenuEffects() could not take `open` as a parameter to avoid this wrapper?

zatteo (approved):

react/ActionsMenu/ActionsMenuWrapper.jsx
JF-Cozy: or use an `isEnabled`... but the approach use by yannick is simpler to deal with a hook I think 🤔 

But other question, here we use this hook to change the color on flagship side, I think we also use this hook in the BottomSheet itself, does this pose any compatibility problems? And also, I see (with my own eye) the hook comes from a deprecated component, is it still relevant?

Ldoppea: I'm not sure how to do this in a simple way as this hooks only call a more generic hook by settings expected colors. Using a conditional component seems to be, by far, the simplest way to do it. At lease for now. In the future we may want to change the internal implementation of the FlagshipUI mechanism to ease support of conditional components.

 if you want to explore another path, we can discuss about it.

JF-Cozy:
image
🤔 Is there a problem with Argos?

react/ActionsMenu/ActionsMenuWrapper.jsx
Ldoppea: @JF-Cozy sorry i missed your comment (we commented at the same time).

> I think we also use this hook in the BottomSheet itself, does this pose any compatibility problems?

Friday I didn't succeeded to make the FlagshipUI mechanism to bug again, so I suppose it should work.

> And also, I see (with my own eye) the hook comes from a deprecated component, is it still relevant?

ActionsMenu is used in the cozy-home sections view, the mode that split connectors from apps and that allows to regroup connectors by category (this mode is not enabled in twake environment however).

PR #2805, opened 2025-06-05
Refactoring on Layout and Sidebar

JF-Cozy:
For a better understanding and readability of the css concerning Layout and related elements.

We've also experienced a number of problems with the use of a complete application in an iframe, via intents. This refactoring corrects some of these issues and enables better rendering in this case.


zatteo:
Looks good 👍

PR #2806, opened 2025-06-11
Add parentBasedIframe prop on breakpoints provider

JF-Cozy:
By default, using `useBreakpoints` inside an iframe will refere to the iframe inner width. If you use `parentBasedIframe=true` on the provider wrapping the iframe, the breakpoints returned inside the iframe will refere to the parent window inner width instead.

### DIFF
--- a/react/helpers/breakpoints.js
+++ b/react/helpers/breakpoints.js
@@ -16,12 +16,20 @@ const breakpoints = {
   isMobile: [0, small]
 }
 
-export const getBreakpointsStatus = breakpoints => {
-  const width = window.innerWidth
+export const getBreakpointsStatus = (breakpoints, innerWidth) => {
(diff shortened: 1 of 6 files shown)


zatteo:
I think we should talk about iframe messaging in our next front meeting.

Allowing BreakpointsProvider to be based on parent data is a good solution for your problem. But we are starting to write iframe messaging code in different places and I am afraid to lose track of it and start a big mess with iframe code everywhere.

If we can centralize it when we can in cozy-interapp it could be nice so we can: 
- centralize every iframe messaging at the same place
- avoid to manage multiple times parent/iframe selection (no need for `window.frames[1]` which is hardcoded and could be another iframe one day)
- avoid to manage multiple times data extraction from message (no need for `data.split(':')`)

zatteo (approved):

PR #2814, opened 2025-06-23
fix(Button): Use primary light background for ghost variant

lenhanphung:
Ghost buttons now use theme.palette.primary.light (#D2E9FF) instead of theme.palette.background.default (grey) to match Figma design specifications.

This change affects the default ghost button variant:
- Background: theme.palette.primary.light
- Hover: darken(theme.palette.primary.light, 0.05)
- Mobile: maintains primary.light background

The customColor primary ghost buttons remain unchanged to preserve existing behavior and avoid breaking changes.


lenhanphung:
> According to Argos, normal light theme doesn't seem to be changed 🤔 and others (normal inverted, dark...) doesn't look very good. Could you please check with Swann ? You can build the doc locally and from there, check with him.

I was updated as we discuss in meeting with Swann

PR #2820, opened 2025-07-02
feat(Layout): Simplify the approach and make it compliant with v-tables

JF-Cozy:


react/Layout/Layout.jsx
```diff
@@ -59,20 +53,16 @@ export const Layout = forwardRef(
...
-        className,
-        styles['o-layout-main'],
-        styles[`o-layout-main-${variant}`],
-        {
-          [styles[`o-layout-main-${variant}--topbar`]]: withTopBar
-        }
-      )}
+      className={cx(className, styles['o-layout-main'], {
+        [styles['o-layout-main-2panes']]: !monoColumn,
+        [styles[`o-layout-main-topbar`]]: withTopBar && !isDesktop
```
zatteo, line 64: Sure about the `!isDesktop` ? If this class is responsible of moving the content down of the size of the topbar we need it in desktop too?

zatteo (approved):

PR #2839, opened 2025-09-01
Remove Cordova related code

zatteo:
To ease the merging of https://github.com/cozy/cozy-libs/pull/2664


JF-Cozy (approved):
I suppose we can remove the BREAKING CHANGE keyword if there is nothing to do in the app. Unless we have to create a specific function in the app and pass it as prop in one way or another...?

Besides, if we don't use `cozy-device-helper` anywhere else, we can remove it.

zatteo:
I was not sure about the breaking change. I will remove it 👍

PR #2845, opened 2025-09-04
fix: Postinstall hook not working using Yarn 4

zatteo:
When installing cozy-ui with Yarn 4, we get the following errors :
command not found: if
command not found: then
command not found: fi

Fixing it following this advice https://stackoverflow.com/questions/70805712/command-not-found-if-when-running-package-json-script-with-yarn-v3-berry


JF-Cozy:
> > maybe we can force using yarn 4 locally and in the CI too to be consistent?
> 
> I got the issue in cozy-libs repository which has been migrated to Yarn 4.

My point is : maybe you should use yarn 4 in cozy-ui too

PR #2850, opened 2025-09-08
feat: Add Storage component

JF-Cozy:
to show storage progress bar and upgrade button


JF-Cozy:
> Just a reflexion : this is exactly the case where we use cozy-client to fetch data where as we could pass the data directly to the component if we wanted to uncouple from cozy-client. Something like below
> 
> ```
> <Storage
>   diskUsed={10.58}
>   diskQuota={100}
>   link="https://..."
> />
> ```

Yes I thought about it. I don't know when and how we will start this job, so I decided to be consistent for now I don't introduce new approach aka fetching data in app and don't have all-in-one component in cozy-ui. There is a lot of component relies deeply on cozy-client, I think one more is not a big deal.

PR #2853, opened 2025-09-12
feat(virtualizedTable): update defaultOrder to be an object containing the properties used to sort the table

Coussecousse:
Hello ! 
We needed to sort the trash by last_update and by default the virtualizedTable componant set the direction to 'asc'. 
So we're adding a new property to dynamically change this value.
Ticket : https://www.notion.so/linagora/Display-files-in-the-trash-by-deletion-date-1c062718bad1803d8aa6c3f51d515d4c?source=copy_link

### DIFF
--- a/react/Table/Readme.md
+++ b/react/Table/Readme.md
@@ -121,7 +121,7 @@ const ExampleTable = ({ variant, ...props }) => {
       <Typography className="u-mt-1" variant="h4">Sorted table</Typography>
       <div className="u-mt-half" style={{ border: "1px solid var(--borderMainColor)", height: 400, width: "100%" }}>
         <SelectionProvider>
-          <ExampleTable variant={variant} defaultOrder={columns[0].id} />
+          <ExampleTable variant={variant} defaultOrder={{by: columns[0].id, direction: 'asc'}} />
         </SelectionProvider>
       </div>
     </>
--- a/react/Table/Virtualized/index.jsx
+++ b/react/Table/Virtualized/index.jsx
@@ -28,8 +28,10 @@ const VirtualizedTable = forwardRef(
     },
     ref
   ) => {
-    const [orderDirection, setOrderDirection] = useState('asc')
-    const [orderBy, setOrderBy] = useState(defaultOrder)
+    const [orderDirection, setOrderDirection] = useState(
+      defaultOrder?.direction ?? 'asc'
+    )
+    const [orderBy, setOrderBy] = useState(defaultOrder?.by ?? undefined)
 
     const sortedData = orderBy
       ? stableSort(rows, getComparator(orderDirection, orderBy))
@@ -117,10 +119,27 @@ VirtualizedTable.defaultProps = {
 }
 
 VirtualizedTable.propTypes = {
-  /** Column ID to be used for initiating the sort */
-  defaultOrder: PropTypes.string,
+  /** Rows to display in the table */
+  rows: PropTypes.array,
+  /** Column configuration */
+  columns: PropTypes.array,
   /** Returned object is: PropTypes.shape({ groupLabels: PropTypes.array, groupCounts: PropTypes.array }) */
   groups: PropTypes.func,
+  /** Default sorting configuration */
+  defaultOrder: PropTypes.shape({
+    direction: PropTypes.oneOf(['asc', 'desc']),
+    by: PropTypes.string
+  }),
+  /** Sort files by type to put directory and trash before files */
+  secondarySort: PropTypes.func,
+  /** Array of selected items */
+  selectedItems: PropTypes.array,
+  /** Callback function when a row is selected */
+  onSelect: PropTypes.func,
+  /** Callback function when all rows are selected/deselected */
+  onSelectAll: PropTypes.func,
+  /** Function to determine if a row is selected */
+  isSelectedItem: PropTypes.func,
   /** Callback called after the sort */
   onSortChange: PropTypes.func
 }


react/Table/Readme.md
JF-Cozy, line 1: Please for the BC description in the commit message, can you add an example of what we had before, and how we have to modify it to deal with the BC. That's way, no need to thing to apply it. You can see examples on the cozy-ui release page.

react/Table/Readme.md
```diff
@@ -118,7 +124,7 @@ const onSortChange = ({ order, orderBy }) => {
         rows={rows}
         columns={columns}
         groups={variant.grouped ? makeGroups : undefined}
-        defaultOrder={columns[0].id}
+        defaultOrder={{ attribute: 'calories', order: 'desc' }}
```
JF-Cozy, line 127: `desc` is a keyword, `calories` is not so I think `columns[0].id` is more relevant, specially if the column construction changes

react/Table/Virtualized/index.jsx
```diff
@@ -43,10 +47,9 @@ const VirtualizedTable = forwardRef(
     }
 
     const handleSort = property => {
-      const isAsc = orderBy === property && order === 'asc'
+      const isAsc = order.attribute === property && order.order === 'asc'
```
JF-Cozy, line 50: there is no conditional chaining `?.` here, is it working if `attribute` or `order` are undefined ?

react/Table/Virtualized/index.jsx
```diff
@@ -28,10 +28,14 @@ const VirtualizedTable = forwardRef(
     },
     ref
   ) => {
-    const [order, setOrder] = useState('asc')
-    const [orderBy, setOrderBy] = useState(defaultOrder)
+    const [order, setOrder] = useState(
+      defaultOrder ?? { attribute: columns[0].id, order: 'asc' }
```
JF-Cozy, line 32: using object in state is tricky but why not, here.

To avoid confusion I think we can have `orderBy` and `orderDirection` everywhere in all components... so here let's have `{ by: columns[0].id, direction: 'asc' }` here and let's reflect that everywhere

Other thing, we didn't have default value previously I think for old `orderBy` state (`defaultOrder` prop could have been `undefined`). Maybe we don't need `?? { attribute: columns[0].id, order: 'asc' }`

JF-Cozy (commented):
sorry I understand it's a bit painful to merge this PR with lot of reviews, but your are in a lib here (we have to be smart and take time to do thing) and it's about virtualized table (complicated component !) so it's normal ;)

react/Table/Virtualized/index.jsx
```diff
@@ -113,14 +115,32 @@ VirtualizedTable.defaultProps = {
...
+  columns: PropTypes.array,
   /** Returned object is: PropTypes.shape({ groupLabels: PropTypes.array, groupCounts: PropTypes.array }) */
   groups: PropTypes.func,
+  /** Default sorting configuration */
+  defaultOrder: PropTypes.shape({
+    direction: PropTypes.oneOf(['asc', 'desc']),
+    by: PropTypes.string
+  }),
+  /** Array of secondary sort configurations */
+  secondarySort: PropTypes.array,
```
JF-Cozy, line 135: nit: I though it was a func instead of an array 🤔

react/Table/Virtualized/index.jsx
```diff
@@ -113,14 +115,32 @@ VirtualizedTable.defaultProps = {
...
+  /** Array of selected items */
+  selectedItems: PropTypes.array,
```
JF-Cozy, line 137: is it really full items or only their ids?

react/Table/Virtualized/index.jsx
```diff
@@ -113,14 +115,32 @@ VirtualizedTable.defaultProps = {
...
+  /** Callback function when a row is selected */
+  onSelect: PropTypes.func,
+  /** Callback function when all rows are selected/deselected */
+  onSelectAll: PropTypes.func,
```
JF-Cozy, line 141: it's a func triggered only when click "all select" checked. The fact that it's a toggle to select/deselect is handle by the app, not here I think (and not really a callback actually I think, like `onSelect`)

react/Table/Virtualized/index.jsx
```diff
@@ -113,14 +115,32 @@ VirtualizedTable.defaultProps = {
   selectedItems: [],
   isSelectedItem: () => {},
   onSelect: () => {},
-  onSelectAll: () => {}
+  onSelectAll: () => {},
+  onSortChange: () => {}
```
JF-Cozy, line 119: don't need to add `onSortChange` here, we did `onSortChange?.()` in the code (or we have to remove this part to be consistent, as you wish). But I think passing empty default func was not really a good idea anyway, it forces a behavior not necessarily real regarding to the app 🤔

react/Table/Virtualized/index.jsx
```diff
@@ -28,8 +28,10 @@ const VirtualizedTable = forwardRef(
     },
     ref
   ) => {
-    const [orderDirection, setOrderDirection] = useState('asc')
-    const [orderBy, setOrderBy] = useState(defaultOrder)
+    const [orderDirection, setOrderDirection] = useState(
+      defaultOrder?.direction ?? 'asc'
+    )
+    const [orderBy, setOrderBy] = useState(defaultOrder?.by ?? undefined)
```
JF-Cozy, line 34: Why not using only one state to deal with sort and use object as default prop, like you did before? Sorry we misunderstood each other probably.

```
defaultProps = { defaultOrder: {} }

const [order, setOrder] = useState(defaultOrder)
```

Then replace `orderDirection` by `order.direction`, `orderBy` by `order.by` and `setOrderDirection | setOrderBy` in `handleSort` by `setOrder({ by: property, direction: newOrder })`

PR #2856, opened 2025-09-17
Some fixes and improvement on VirtualizeTable

JF-Cozy:


rezk2ll:
This is conflicting with https://github.com/cozy/cozy-ui/pull/2853 i propose merging the two PRs

JF-Cozy:
> This is conflicting with #2853 i propose merging the two PRs

No BC here for the Table things, I think she will be good to go by simply rebasing. Don't know when her PR will be ready, I think it's not a big deal if we do thing seperatly

react/Table/Readme.md
```diff
@@ -70,59 +73,30 @@ const columns = [
...
-const onSortChange = ({ order, orderBy }) => {
-  setState({ order, orderBy })
-}
+const ExampleTable = ({ variant, ...props }) => {
+  const { selectedItemsId, isSelectedItem, toggleSelectedItem, toggleSelectAllItems } = useSelection()
 
-<Variants initialVariants={initialVariants} screenshotAllVariants>
-  {variant => (
+  const onSortChange = ({ order, orderBy }) => {
+    console.info('order:', order, 'oderBy:', orderBy)
```
rezk2ll, line 86: we can keep the same setState example in the readme

rezk2ll (approved):

react/Table/Readme.md
JF-Cozy: If think it was a bug to use it here. `setState` refers to `initialState` and there is no `order` neither `orderBy` in it. There is no purpose to use this callback in the doc as is, so I think just a console do the tricks

PR #2860, opened 2025-09-24
feat: Add ru and vi locales

zatteo:


rezk2ll (approved):
i don't understand any of those languages but погнали

PR #2861, opened 2025-09-24
feat(TableRow, VirtualizedTable): add isNewItem prop highlight new items

Coussecousse:
Bob ask me to work on a functionnality that 'highlight' in blue the new items. 
It used the same blue as selected items. 
The isNewItem prop can be null without causing an issue.

Example : 
image

Code in tdrive is coming :)


react/Table/Virtualized/Dnd/TableRow.jsx
```diff
@@ -83,7 +84,7 @@ const TableRow = ({ item, context, ...props }) => {
     <TableRowClassic
       {...props}
       ref={node => dragRef(dropRef(node))}
-      selected={isSelected || dropCollect.isOver}
+      selected={isSelected || dropCollect.isOver || isNew}
```
JF-Cozy, line 87: As we finally use `selected` prop from `Mui TableRow`, I don't think we have to change anything in cozy-ui, isn't it? Could we just have to deal with `isSelectedItem` main prop (maybe it conflicts with DnD... 🤔) in the app?

JF-Cozy: I agree, that's why we decided to not use `selected` prop in the first place and to add a specific css or something for it, if I remember correctly.

rezk2ll (approved):

PR #2863, opened 2025-10-01
feat(virtualized table): Disable table row on componentsProps.tableRow.disabled

lethemanh:
### Change:

Disable the Tablerow when `componentsProps.tableRow.disabled` is set

### Related to:

[cozy/cozy-drive#3516](https://github.com/cozy/cozy-drive/pull/3516)

### DIFF
--- a/react/Table/Virtualized/Dnd/TableRow.jsx
+++ b/react/Table/Virtualized/Dnd/TableRow.jsx
@@ -4,7 +4,7 @@ import { getEmptyImage } from 'react-dnd-html5-backend'
 
 import TableRowClassic from '../../../TableRow'
 
-const TableRow = ({ item, context, ...props }) => {
+const TableRow = ({ item, context, componentsProps, ...props }) => {
   const {
     selectedItems,
     itemsInDropProcess,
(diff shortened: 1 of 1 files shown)


react/Table/Virtualized/Dnd/TableRow.jsx
```diff
@@ -17,8 +17,9 @@ const TableRow = ({ item, context, ...props }) => {
     canDrag: canDragProps,
     dragId
   } = dragProps
+  const { isCut } = props
```
rezk2ll, line 20:
```suggestion
  const { isDimmed } = props
```

I would use other names not to mix Drive and Cozy-UI terms:
- isInactive
- isDimmed
- hasReducedOpacity

react/Table/Virtualized/Dnd/TableRow.jsx
```diff
@@ -17,8 +17,9 @@ const TableRow = ({ item, context, ...props }) => {
...
   const isSelected = context?.isSelectedItem(item)
-  const isDisabled = itemsInDropProcess.includes(item._id)
+  const isDisabled = itemsInDropProcess.includes(item._id) || isCut
```
rezk2ll, line 22:
```suggestion
  const isDisabled = itemsInDropProcess.includes(item._id)
```

Let's not mix drag and drop with our newly added prop.
It should be better to just use it like this for example

```tsx
<TableRowClassic
  {...props}
...
  disabled={isDisabled || isDimmed}
  hover
/>
```

JF-Cozy:
If think "all" is in the title : "Disable table row on cut". So here, we want the ability to disable row. That's for the "what". The how is "on cut" but cozy-ui doesn't care about the "how" here, it's the app responsibility.

So the question is : how to disable a row. We have already `componentsProps` to help us pass some props to deep components, we can rely on it. It should be `componentsProps.tableRow.disabled`, and get it in TableRow comp like you did `const isDisabled = disabled || itemsInDropProcess.includes(item._id)` and I think we are done

PR #2916, opened 2026-04-07
feat: Change sidebar style

doubleface:
image

https://doubleface.github.io/cozy-ui/react/

https://www.notion.so/linagora/Sidebar-design-changes-33062718bad1809697d4e0758874cb28

### DIFF
--- a/stylus/components/nav.styl
+++ b/stylus/components/nav.styl
@@ -21,7 +21,7 @@ $nav
 $nav-item
     position        relative
     z-index         var(--zIndex-app)
-    height          rem(44)
+    height          rem(36)
     box-sizing      border-box
     cursor          pointer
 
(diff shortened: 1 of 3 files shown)


stylus/components/nav.styl
```diff
@@ -104,12 +105,14 @@ $nav-link
         color  var(--actionColorActive)
 
     &.is-active
-        box-shadow   inset rem(4) 0 0 0 var(--primaryColor)
-        font-weight  bold
-        color  var(--primaryTextColor)
+        background-color  var(--primaryColorLightest)
```
JF-Cozy: I don't think we should use --primaryColorLightest it's an exception 🤔 not really in mui. Could we just use --primaryColorLight ?

stylus/components/nav.styl
JF-Cozy, line 1: Argos shows that elements in mobile are not centered and only two items are shown, is it ok?

doubleface: Fixed

JF-Cozy (approved):

PR #2917, opened 2026-04-08
feat: Add new menu icon

doubleface:
image


JF-Cozy (approved):
the icon seems pretty thin to me 🤔 specially visible in dark mode. 

Image

If you are sure we can merge go for it, but maybe a double check is not wasted time. Maybe the selection wasn't edges to edges in a 16x16 square in the first place... ?

paultranvan:
> If you are sure we can merge go for it

Yes, validated by Swann

PR #2919, opened 2026-04-10
feat: Add skill reference generator for cozy-ui components

doubleface:
Add a script that generates AI skill reference files from source code (PropTypes, Readme examples, CSS classes). Uses remark for robust markdown parsing and picks the best JSX example across all blocks.

css classes reference is directly ai generated because it is easier like this and it does not change a lot at the moment

Just a draft. Not totally satisfied with the result but it could be a start


zatteo:
Thanks for starting this work! It can raise a lot of question but right now I have these three observation:
1. I have no idea how to check if what is generated is relevant or not
2. Following 1., maybe we need to quickly "try it" to check if it is good
3. Could our current documentation engine manage this? It is already doing the job to generate doc for humans, maybe it can do it for AI too? (I know that the one we use is outdated so maybe it is part of a bigger project to update our doc in general)

Also, how did you choose _picks the best JSX example across all blocks_

JF-Cozy:
I think we should add some doc to explain what is it, and what is it for.

But for the AI to have cozy-ui knowledge in our project, is it not enough to give it the link of the documentation? It worked for Cursor 🤔

doubleface:
> But for the AI to have cozy-ui knowledge in our project, is it not enough to give it the link of the documentation? It worked for Cursor 🤔

It can be enough but it will use a lot more tokens

Crash--:
So happy to see this kind of PR. Do not hesitate to merge and iterate on it. Yes it’s not perfect but let’s fix it during the way. 

I’m convinced that this skill should belongs to twake-guidelines but if it complexifies the pipeline right now let’s keep it there.

zatteo (approved):

zatteo:
My two cent plan following current discussion @doubleface :
- merge it
- share to front end team that it exists with how to install for OpenCode
- check in two weeks if it works + why (=> add a event in calendar to avoid forgetting it)

PR #2924, opened 2026-04-22
refactor(AlertProvider): clarify default duration handling

rezk2ll:
## Context

`AlertProvider`'s `defaultState` declared `duration: null`, which reads as "alerts are persistent by default". The actual default behavior is a 2s auto-dismiss — `null` was MUI plumbing (`Snackbar` treats it as "disabled"), never an API contract.

## Change

Drop the misleading default. The 2s auto-dismiss keeps applying as before; explicit `duration: null` remains the documented opt-out for persistent alerts. Tests pin the three duration paths.

No user-visible behavior change.

### DIFF
--- a/react/providers/Alert/index.jsx
+++ b/react/providers/Alert/index.jsx
@@ -30,7 +30,6 @@ export const useAlert = () => {
 const defaultState = {
   title: '',
   message: '',
-  duration: null,
   open: false
 }
 
@@ -53,7 +52,7 @@ const AlertProvider = ({ children }) => {
        * @param {ShowAlertArgs} args
        */
       showAlert: args => {
-        setState({ open: true, ...args })
+        setState({ ...defaultState, ...args, open: true })
       }
     }),
     []
--- a/react/providers/Alert/Readme.md
+++ b/react/providers/Alert/Readme.md
@@ -17,6 +17,7 @@ const initialVariants = [{
   outlined: false,
   noClickAway: false,
   noTimeOut: false,
+  persistent: false,
   close: true
 }]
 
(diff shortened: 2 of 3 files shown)


react/providers/Alert/index.jsx
```diff
@@ -53,7 +53,7 @@ const AlertProvider = ({ children }) => {
        * @param {ShowAlertArgs} args
        */
       showAlert: args => {
-        setState({ open: true, ...args })
+        setState({ ...defaultState, ...args, open: true })
```
Crash--, line 55: Can we have tests for this?

rezk2ll: yes, i added them

zatteo:
Thank you for the analysis. We should mark it as a breaking change to not forget to update all calls that need to keep current behavior.

JF-Cozy:
By adding `duration null` https://github.com/cozy/cozy-ui/pull/2671 I don't think we wanted to ignore the auto hide by default. I think we just wanted to be able to customize the duration. So if I understood correctly (`null` disable the auto hide, and `undefined` fallback to default duration) we need to fix the default behavior from `null` to `undefined` as well, isn't it?

rezk2ll:
> By adding `duration null` #2671 I don't think we wanted to ignore the auto hide by default. I think we just wanted to be able to customize the duration. So if I understood correctly (`null` disable the auto hide, and `undefined` fallback to default duration) we need to fix the default behavior from `null` to `undefined` as well, isn't it?

i like your proposal more, this way we won't have to introduce a breaking change

JF-Cozy (approved):

PR #2925, opened 2026-04-27
feat: Change overline style

doubleface:
This is part of the migration to the new sidebar.
It is hard to put this change in a new component or behind a flag but since the overline variant is only used in a small unnoticable part of drive application, I think it is quite safe to do it now.

### DIFF
--- a/react/MuiCozyTheme/makeTypography.js
+++ b/react/MuiCozyTheme/makeTypography.js
@@ -69,10 +69,10 @@ export const makeTypography = () => ({
     letterSpacing: '0.5px'
   },
   overline: {
-    fontSize: 11,
-    fontWeight: 500,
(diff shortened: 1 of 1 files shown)


rezk2ll (approved):

JF-Cozy:
@doubleface @rezk2ll If I understand correctly, we are working here on the new sidebar, and we use overline inside it. So we modify it to fit the new needs inside the new sidebar. And because it only used in a small part of our apps, we decided to merge.

I don't think it's the right way, and this is a good reason.

We have to think globally for component and other stuff, we can't just modify overline without thinking about what is it for in our designs.

If you need to work on the new sidebar, go for it in a isolated PR and replace the old one after validation with designers.

Besides please don't merge without checking Argos !

PR #2927, opened 2026-05-06
feat: Change NavDesktopDropdown label to caption variant

doubleface:
image

Validated with product

### DIFF
--- a/react/Nav/index.jsx
+++ b/react/Nav/index.jsx
@@ -140,7 +140,7 @@ export const NavDesktopDropdown = ({
     <>
       <ListItem size="small" className={isActivated ? 'u-c-pointer' : ''}>
         <DropdownText
-          variant="overline"
+          variant="caption"
           color="textSecondary"
           spaceBetween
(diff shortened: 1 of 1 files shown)


react/Nav/index.jsx
```diff
@@ -140,7 +140,7 @@ export const NavDesktopDropdown = ({
     <>
       <ListItem size="small" className={isActivated ? 'u-c-pointer' : ''}>
         <DropdownText
-          variant="overline"
+          variant="caption"
```
JF-Cozy, line 143: we previously changed the style of overline for that, I suppose... is it still relevant?

doubleface: The overline style change is not relevant anymore, no. Should I change it back ?

JF-Cozy: Maybe. We should ask designers for that. Is it still relevant for the future, or was it only for the nav

JF-Cozy (approved):

PR #2929, opened 2026-05-13
feat: Allow to customize background color of ProgressionBanner

doubleface:


react/ProgressionBanner/index.jsx
```diff
@@ -63,11 +63,14 @@ ProgressionBanner.propTypes = {
   /** Button to use in the banner */
   button: PropTypes.node,
   /** Progression bar is hidden if set to false (defaults to true) */
-  progressBar: PropTypes.bool
+  progressBar: PropTypes.bool,
+  /** Background color of the banner */
```
rezk2ll: we can also document the expected format

react/ProgressionBanner/index.jsx
```diff
@@ -19,7 +19,7 @@ const styles = () => ({
 })
 
 const ProgressionBanner = withStyles(styles)(
-  ({ classes, value, text, icon, button, progressBar }) => {
+  ({ classes, value, text, icon, button, progressBar, backgroundColor }) => {
```
rezk2ll:
```suggestion
  ({ classes, value, text, icon, button, progressBar, color }) => {
```

i think we should stick with the same naming as the Alert provider ( or change the latter to use backgroundColor instead of color ).

But we shouldn't have two names for the same thing

rezk2ll (commented):
Please don't forget to update the component docs in `react/ProgressionBanner/Readme.md`

rezk2ll (approved):

PR #2933, opened 2026-05-28
Table: add double click

JF-Cozy:
The browser fires click events before dblclick — there's no way to intercept or prevent the first click. The only alternative is delaying every click with a timer, which adds unwanted lag.
We accept this native behavior: double-click triggers click first, then double-click. Google does the same in Drive and Sheets — single-click selects, double-click opens.

### DIFF
--- a/react/Table/Readme.md
+++ b/react/Table/Readme.md
@@ -15,7 +15,7 @@ const createData = (id, name, calories, fat, carbs, protein) => {
 
 const rows = [
   createData(0, 'Cupcake', 305, 3.7, 67, 4.3),
-  createData(1, 'Frozen yoghurt', 159, 6.0, 24, 4.0),
+  createData(1, 'Frozen yoghurt (unselectable)', 159, 6.0, 24, 4.0),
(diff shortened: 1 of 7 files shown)


react/Table/Readme.md
```diff
@@ -95,7 +95,7 @@ const ExampleTable = ({ variant, ...props }) => {
         groups={variant.grouped ? makeGroups : undefined}
         selectedItems={selectedItemsId}
         isSelectedItem={row => isSelectedItem(row.id)}
-        onSelect={(row, event, index) => toggleSelectedItem(row.id)}
+        onSelect={(row, event, index) => row.id !== 1 ? toggleSelectedItem(row.id) : undefined}
```
zatteo, line 98: Why do we want to to disable selection on item.id === 1?

zatteo (approved):

react/Table/Readme.md
JF-Cozy: Just for demo, to show how to disable a row

PR #2935, opened 2026-06-01
feat(useSubmitWithLoader): Avoid double submit

zatteo:


react/hooks/useSubmitWithLoader/index.spec.jsx
```diff
@@ -0,0 +1,74 @@
...
+import { useSubmitWithLoader } from './index'
+import AlertProvider from '../../providers/Alert'
+
+jest.mock('../../Snackbar', () => ({
+  __esModule: true,
+  default: ({ children }) => children
+}))
+
+describe('useSubmitWithLoader', () => {
+  const renderUseSubmitWithLoader = () =>
```
JF-Cozy: we can extract this outside the describe in a setup func

JF-Cozy (approved):
it's a fix rather than a feat I think

PR #2942, opened 2026-07-02
fix(Autocomplete): Small size is now the same as TextField component

JF-Cozy:
Previously the small size of autocomplete was smaller than Textfield small.

demo: https://jf-cozy.github.io/cozy-ui/react/#/Autocomplete


zatteo (approved):
Oh great!

PR #2943, opened 2026-07-08
Add CustomColorPicker

zatteo:
## Screenshot

image

## Demo

https://zatteo.github.io/cozy-ui/react/#!/Core/ColorPickerCustom


zatteo:
Let's see how we can integrate this component in the circle color picker!

docs/skill/css-classes.md
JF-Cozy, line 1: is it with AI or with package.json script ?

package.json
JF-Cozy, line 1: why not in peerDep, like we did with react-dnd for example, what the AI says about it? Is the lib in the final bundle if we let it in dep, but don't use the color picker component in the app?

react/ColorPickerCustom/styles.styl
```diff
@@ -0,0 +1,19 @@
+.wrapper
```
JF-Cozy: .wrapper is probably useless, unless we use 2 color pickers at the same time in the page I suppose 🤔 we can probably remove it to make it simple, what do you think?

JF-Cozy (approved):
works great 👍

docs/skill/css-classes.md
zatteo: With _yarn lint_

react/ColorPickerCustom/styles.styl
zatteo: I tried to remove it without success, so I added a comment explaining why

package.json
zatteo: Normally, it is not an issue here:
- color picker lib is scoped to color picker component
- rsbuild supports tree shaking for barrel file

So it should not be in final bundle of app using it. I propose to merge like this and test after if there is an build size update in an app. It will be a good check if our tree shaking work.

PR #2950, opened 2026-07-29
fix(breakpoints)!: target the requesting iframe

### LINKED ISSUE #2949: BreakpointsProvider crashes when a parent-based iframe is the only iframe
## Context

This issue was discovered while integrating `BreakpointsProvider` with `parentBasedIframe` into a File Picker rendered through a Cozy intent.

The corresponding Drive integration is still work in progress and is not available on Drive's main branch. However, the failure can be reproduced independently in `cozy-ui`.

The affected setup contains:

- a parent application using `BreakpointsProvider`;
- a child application rendered inside an iframe;
- a child `BreakpointsProvider` configured with `parentBasedIframe`;
- no other iframe in the parent document.

## Environment

- Reproduced with `cozy-ui` 140.5.2
- Also present on the current `cozy-ui` master branch
- Browser: Chromium
- The child is the first and only iframe in the parent document

## Minimal setup

The parent application renders its regular provider and an iframe:

```jsx
<BreakpointsProvider>
  <iframe src="/child.html" />
</BreakpointsProvider>
```

The application loaded inside the iframe enables parent-based breakpoints:

```jsx
<BreakpointsProvider parentBasedIframe>
  <App />
</BreakpointsProvider>
```

No additional iframe should be present in the parent document.

## Minimal test reproduction

The parent-side failure can also be reproduced with this focused test:

```jsx
import { act, render } from '@testing-library/react'
import React from 'react'

import { BreakpointsProvider } from './index'

it('sends parent breakpoints to the requesting iframe', () => {
  const iframeWindow = { postMessage: jest.fn() }
  const message = new MessageEvent('message', {
    data: 'UI-breakpoints-needParentBreakpoints'
  })

  Object.defineProperty(message, 'source', {
    value: iframeWindow
  })

  render(
    <BreakpointsProvider>
      <div />
    </BreakpointsProvider>
  )

  act(() => {
    window.dispatchEvent(message)
  })
})
```

## Actual behavior

When the child requests the parent breakpoints, the parent throws:

```text
TypeError: Cannot read properties of undefined (reading 'postMessage')
```

The exception prevents the iframe application from becoming ready and may leave the caller application with a blank page.

## Expected behavior

The child should receive the parent window breakpoints and render normally when it is:

- the only iframe in the document;
- the first iframe in the document;
- one of several iframes in the document.

The behavior should not depend on the iframe position in the caller document.

## Additional observations

The problem may remain unnoticed when an unrelated iframe is already present in the parent page. It reproduces consistently in a fresh Chromium or Playwright context where the child is the only iframe.

doubleface:
## Summary

- Send parent breakpoint widths to the iframe that requested them instead of relying on a fixed frame index.
- **Breaking change:** replace the old hook usage with:

  ```js
  const { iframeWindow } = useIframeConnection({ parentBasedIframe })
  useIframeToSendWidth({ iframeWindow })
  ```

- Closes #2949.

### DIFF
--- a/react/providers/Breakpoints/useIframeConnection.jsx
+++ b/react/providers/Breakpoints/useIframeConnection.jsx
@@ -3,7 +3,7 @@ import { useEffect, useState } from 'react'
 import { isInsideIframe } from '../../helpers/breakpoints'
 
 export const useIframeConnection = ({ parentBasedIframe }) => {
-  const [hasIframe, setHasIframe] = useState(false)
+  const [iframeWindow, setIframeWindow] = useState(null)
 
   const isIframe = parentBasedIframe && isInsideIframe()
 
@@ -11,7 +11,7 @@ export const useIframeConnection = ({ parentBasedIframe }) => {
   useEffect(() => {
     const handleMessage = ev => {
       if (ev.data === 'UI-breakpoints-needParentBreakpoints') {
-        setHasIframe(true)
+        setIframeWindow(ev.source)
       }
     }
 
@@ -27,5 +27,5 @@ export const useIframeConnection = ({ parentBasedIframe }) => {
     }
   }, [isIframe])
 
-  return { hasIframe }
+  return { iframeWindow }
 }
--- a/react/providers/Breakpoints/useIframeToSendWidth.jsx
+++ b/react/providers/Breakpoints/useIframeToSendWidth.jsx
@@ -1,27 +1,19 @@
 import throttle from 'lodash/throttle'
 import { useEffect } from 'react'
 
-/**
- * To send window innerWidth to first iframe
- * @returns
- */
-const sendWidthToIframe = () =>
-  window.frames[1].postMessage(`UI-breakpoints-value:${window.innerWidth}`, '*')
+const sendWidthToIframe = iframeWindow =>
+  iframeWindow?.postMessage(`UI-breakpoints-value:${window.innerWidth}`, '*')
 
-export const useIframeToSendWidth = ({ hasIframe }) => {
+export const useIframeToSendWidth = ({ iframeWindow }) => {
   // parent window send its innerWidth
   useEffect(() => {
-    if (hasIframe) {
-      sendWidthToIframe()
-    }
-  }, [hasIframe])
+    sendWidthToIframe(iframeWindow)
+  }, [iframeWindow])
 
   // parent window send its innerWidth on resize
   useEffect(() => {
     const handleResize = throttle(() => {
-      if (hasIframe) {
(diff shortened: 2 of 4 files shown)


react/providers/Breakpoints/useIframeConnection.jsx
```diff
@@ -27,5 +27,5 @@ export const useIframeConnection = ({ parentBasedIframe }) => {
     }
   }, [isIframe])
 
-  return { hasIframe }
+  return { iframeWindow }
```
JF-Cozy, line 30: it's a BC

react/providers/Breakpoints/useIframeToSendWidth.jsx
```diff
@@ -1,33 +1,25 @@
...
- * To send window innerWidth to first iframe
- * @returns
- */
-const sendWidthToIframe = () =>
-  window.frames[1].postMessage(`UI-breakpoints-value:${window.innerWidth}`, '*')
+const sendWidthToIframe = iframeWindow =>
+  iframeWindow?.postMessage(`UI-breakpoints-value:${window.innerWidth}`, '*')
 
-export const useIframeToSendWidth = ({ hasIframe }) => {
+export const useIframeToSendWidth = ({ iframeWindow }) => {
```
JF-Cozy, line 7: also a BC

JF-Cozy:
this all thing doesn't seams to be a good idea anymore https://github.com/cozy/cozy-ui/pull/2806#issuecomment-2962770365 did you speak with @zatteo about iframe messaging?

zatteo:
> this all thing doesn't seams to be a good idea anymore [#2806 (comment)](https://github.com/cozy/cozy-ui/pull/2806#issuecomment-2962770365) did you speak with @zatteo about iframe messaging?

I add the topic of "resizing of iframe related to useBreakpoints" in the topic related to File Picker. We may need a more general approach but we need to think to what extend our other Flutter/Calendar apps are concerned. But if it solve an issue right now, let's merge this kind of fix (taking into account the BC).

zatteo (approved):

JF-Cozy:
@doubleface for the BC I prefere something really specific to avoid any thought. Could you please add this in the BC message?

So you must use `const { iframeWindow } = useIframeConnection({ parentBasedIframe })`and `useIframeToSendWidth({ iframeWindow })`

doubleface:
> @doubleface for the BC I prefere something really specific to avoid any thought. Could you please add this in the BC message?
> 
> So you must use `const { iframeWindow } = useIframeConnection({ parentBasedIframe })`and `useIframeToSendWidth({ iframeWindow })`

Done

PR #2951, opened 2026-08-25
fix(chips): align primary active colors

shepilov:
## Summary

- Use the primary ghost background for active primary chips.
- Keep semantic active chip colors filled.
- Cover primary and semantic active styles with regression tests.


JF-Cozy:
If my deductions are right, and after talking with Swann, your needs seems to be for the filter chips. But I think we override the style in Drive because we don't have the same thing as cozy-ui provides. So no need to touch cozy-ui here, you have to continue overriding the style in Drive for the dark mode...

shepilov:
Closing this library-level change. The styling will be implemented in Twake Drive alongside the existing application theme overrides.

